PR #4, opened 2021-06-10
Feature/rdf node managment

bigerl:
simple implementation of managed RDF Nodes https://github.com/rdf4cpp/rdf4cpp/projects/3#card-62613096 

This is not yet optimized for optimal performance. The focus was to provide an implementation which is very easy to use and can later be optimized for performance.


src/rdf4cpp/rdf/graph/node_manager/BNodeBackend.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <rdf4cpp/rdf/graph/node_manager/NodeID.h>
+
+#include <compare>
+#include <string>
+
+namespace rdf4cpp::rdf::graph::node_manager {
+
+class BNodeBackend {
+    std::string identifier_;
```
TortugaAttack, line 12: why identifier_ and not identifier ?

src/rdf4cpp/rdf/graph/node_manager/LiteralBackend.cpp
```diff
@@ -0,0 +1,19 @@
+#include "LiteralBackend.h"
+
+namespace rdf4cpp::rdf::graph::node_manager {
+
+LiteralBackend::LiteralBackend(std::string lexical, const NodeID &dataType, std::string langTag)
```
TortugaAttack, line 5: just as a note: as discussed we  will use an own DataType Class

src/rdf4cpp/rdf/graph/node_manager/LiteralBackend.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <rdf4cpp/rdf/graph/node_manager/NodeID.h>
+
+#include <compare>
+#include <string>
+
+namespace rdf4cpp::rdf::graph::node_manager {
+
+class LiteralBackend {
+    NodeID datatype_id_;
```
TortugaAttack, line 12: just a note: same here with XSDDataType/DataType class

src/rdf4cpp/rdf/node/IRIResource.cpp
```diff
@@ -1,5 +1,20 @@
...
-// Created by Lixi Alié Conrads on 5/30/21.
-//
-
 #include "IRIResource.h"
+
+namespace rdf4cpp::rdf::node {
+
+IRIResource::IRIResource(void *ptr, const RDFNode::NodeID &id) : RDFNode(ptr, id) {}
+IRIResource::IRIResource() : RDFNode() {}
+IRIResource::IRIResource(const std::string &iri, RDFNode::NodeManager &node_manager)
```
TortugaAttack, line 7: just as a note: As discussed in final we will use IRI object, rather than string to assure validity

TortugaAttack (approved):
ResourceNode would be still good to have for Graph as a RDFNode -> ResourceNode -> {BNode, IRIResource}

Where Subject can only be a ResourceNode

src/rdf4cpp/rdf/graph/node_manager/BNodeBackend.h
bigerl: to avoid a name clash with https://github.com/rdf4cpp/rdf4cpp/blob/1f8c0ca63f063a87d6a032c33274bf1a7c2913ae/src/rdf4cpp/rdf/graph/node_manager/BNodeBackend.h#L19

I find suffix `_` preferable over prefix, because IDE auto completion works better this way.

PR #5, opened 2021-06-23
first draft of RDF Dataset and RDF Graph with a transparent RDF NodeManager

bigerl:
This pull-request addresses https://github.com/rdf4cpp/rdf4cpp/projects/1#card-62606613

It adds the following features:
- adds RDF Dataset and RDF Graphs 
  - be loaded with RDF Quads and RDF Triples
  - can be queried with Quad Patterns / Triple patterns 
  - use the same backend and own each other
  - custom Backends can be used by implementing the Interface DatasetBackend
- adds support for transparent RDF Node management and storage
  - transperent means: NodeStorage takes care of everything in the background
  - you can have multiple NodeStorages; by default a default primary NodeStorage is used
  - NodeStorage's have a usage-counting backend and will self destruct when not used anymore; (except for the primary which lives forever)
  - total number of NodeStorage instances is limited to 1024
  - custom Backends can be used by implementing the Interface INodeStorageBackend
- adds writers for NTriples and NQuads
- doxygen configuration

missing things:
- counting RDFNodes: currently, the user is responsible that (a non-primary) NodeStorage outlives the RDFNodes it holds; with a future patch this should automatically be taken care of
- typed literal features like numbers etc. 
- type tags for common types in NodeID
- CI hook for docs
- see todos in the code

Checkout https://github.com/rdf4cpp/rdf4cpp/blob/feature/rdf_dataset_support/delete_me_later_please.cpp about how things work currently.


src/rdf4cpp/rdf/writer/NTriplesWriter.cpp
```diff
@@ -0,0 +1,24 @@
...
+NTriplesWriter::operator std::string() const {
+    std::stringstream stream;
+    stream << *this;
+    return stream.str();
+}
+std::ostream &operator<<(std::ostream &os, const NTriplesWriter &writer) {
+    using namespace query;
+    auto solutions = writer.graph_.backend()->match(QuadPattern(writer.graph_.name(), Variable("s"), Variable("p"), Variable("o")));
+    for (const auto &solution : solutions) {
+        os << solution[0] << " "
```
TortugaAttack: Shouldn't it be also wrapped with the NNodeWriter ?

src/rdf4cpp/rdf/Dataset.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    [[nodiscard]] size_t size() const;
+
+    [[nodiscard]] size_t size(const IRI &graph_name) const;
+
+    Graph graph(const IRI &graph_name);
+
+    Graph graph();
+
+    void add_ttl_file(const std::string &path);
```
TortugaAttack, line 50: good for now, 
Would change that in future to read_from_stream()

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -0,0 +1,56 @@
...
+    : Node(BackendNodeHandle{
+              node_storage.get_typed_literal(
+                                  lexical_form,
+                                  datatype.handle_.id())
+                      .second}) {}
+Literal::Literal(const std::string &lexical_form, const std::string &lang, Node::NodeStorage &node_storage)
+    : Node(BackendNodeHandle{node_storage.get_lang_literal(lexical_form, lang).second}) {}
+
+
+IRI Literal::datatype() const {
```
TortugaAttack, line 22: IRI is good for now, 
would change that in the future to an own class, which can validate the literal value.

src/rdf4cpp/rdf/Quad.cpp
```diff
@@ -0,0 +1,11 @@
+#include "Quad.hpp"
+
+namespace rdf4cpp::rdf {
+Quad::Quad(Node graph, Node subject, Node predicate, Node object) : QuadPattern(graph, subject, predicate, object) {}
```
TortugaAttack, line 5: maybe validate on creation as well? :thinking:

src/rdf4cpp/rdf/Statement.cpp
```diff
@@ -0,0 +1,9 @@
+#include "Statement.hpp"
+namespace rdf4cpp::rdf {
+Statement::Statement(Node subject, Node predicate, Node object) : query::TriplePattern(subject, predicate, object) {}
```
TortugaAttack, line 3: same here: maybe validate on creation as well. 

(maybe a factory would be better even though it is less concise)

src/rdf4cpp/rdf/query/QuadPattern.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+    return (graph().is_iri() and ((IRI) graph()).identifier() == "")  // Graph
+                   ? std::string{}
+                   : (((std::string) graph()) + " ") +
+                             (std::string) subject() + " " +  // Subject
+                             (std::string) predicate() +      // Predicate
+                             (std::string) object() + " . ";  // Object
+}
+QuadPattern::QuadPattern(Node graph, Node subject, Node predicate, Node object) : entries_({graph, subject, predicate, object}) {}
+Node &QuadPattern::graph() { return entries_[0]; }
```
TortugaAttack, line 13: ok, would have put the context as last, but thats just me :D
but we should rename that still to context. :thinking:

src/rdf4cpp/rdf/query/TriplePattern.cpp
```diff
@@ -0,0 +1,30 @@
...
+}
+TriplePattern::TriplePattern(Node subject, Node predicate, Node object) : entries_{{subject, predicate, object}} {}
+Node &TriplePattern::subject() { return entries_[0]; }
+const Node &TriplePattern::subject() const { return entries_[0]; }
+Node &TriplePattern::predicate() { return entries_[1]; }
+const Node &TriplePattern::predicate() const { return entries_[1]; }
+Node &TriplePattern::object() { return entries_[2]; }
+const Node &TriplePattern::object() const { return entries_[2]; }
+bool TriplePattern::valid() const {
+    return ((subject().is_iri() or subject().is_variable()) and
```
TortugaAttack: or `is_bnode()` right?

src/rdf4cpp/rdf/query/TriplePattern.cpp
```diff
@@ -0,0 +1,30 @@
...
+            (predicate().is_iri() or predicate().is_variable()) and
+            (object().is_iri() or object().is_literal() or object().is_variable()));
```
TortugaAttack: same here: or `is_bnode()` right?

src/rdf4cpp/rdf/storage/node/BNodeBackend.cpp
```diff
@@ -10,10 +10,10 @@ std::strong_ordering BNodeBackend::operator<=>(const BNodeBackend *other) const
     else
         return std::strong_ordering::greater;
 }
-std::string BNodeBackend::as_string([[maybe_unused]] bool quoting) const {
+std::string BNodeBackend::n_string() const {
```
TortugaAttack, line 13: why `n_string`?  (aka what does the n stands for?)

src/rdf4cpp/rdf/storage/node/LiteralBackend.cpp
```diff
@@ -0,0 +1,19 @@
...
+LiteralBackend::LiteralBackend(std::string lexical, const NodeID &dataType, std::string langTag)
+    : datatype_id_(dataType),
+      lexical(std::move(lexical)),
+      lang_tag(std::move(langTag)) {}
+std::strong_ordering LiteralBackend::operator<=>(const LiteralBackend &other) const noexcept {
+    return std::tie(this->datatype_id_.node_id(), this->lexical, this->lang_tag) <=> std::tie(other.datatype_id_.node_id(), other.lexical, other.lang_tag);
+}
+
+std::string LiteralBackend::quote_lexical() const {
+    return "\"" + lexical + "\"";
```
TortugaAttack, line 15: this is problematic if lexical contains an unescaped "

TortugaAttack (approved):
Some stuff i annotated. 
Documentation is missing for a lot of these classes. Will be a real hassle to add them later on. :D

src/rdf4cpp/rdf/writer/NTriplesWriter.cpp
TortugaAttack: or like use the n_string method?

bigerl: right, I'll fix that.

src/rdf4cpp/rdf/Dataset.hpp
bigerl: Yes, I agree. I'll add a todo to remove it later.

src/rdf4cpp/rdf/Literal.cpp
bigerl: Yes, we will also need a concept on how exactly we integrate datatype support.

src/rdf4cpp/rdf/Quad.cpp
bigerl: I would suggest to add an 5th parameter `validate = true`. That validate by default but would leave the user the freedom to first create a Quad and then fill it e.g. in a loop with valid values.

src/rdf4cpp/rdf/Statement.cpp
bigerl: What benefits do we get by using a factory?

TortugaAttack: Using a constructor we cannot validate, because we would need to throw an exception if it wouldn't be valid. (so we could, but it would be messy)
Using a factory we can validate on the spot and return null or better a Result object, containing the error why it's invalid if it's invalid.

src/rdf4cpp/rdf/query/QuadPattern.cpp
bigerl: I would rather stick with graph (or graph_name). I tried to stick as close as possible to https://www.w3.org/TR/rdf11-datasets/#each-named-graph-defines-its-own-context . 
Here they also mention a context but say that it is defined by a named graph.

src/rdf4cpp/rdf/Quad.cpp
TortugaAttack: fine with me, would there be a use case where it would be good to have it first and validate it later? :thinking:  
I mean normally I would use interfaces like `Quad::Quad(IRINode graph, SubjectNode subject, IRINode predicate, Node object)` or similar, but as this seems to be messy in C++ i think that will work as well. 

Same here with the benefit of using a factory.

src/rdf4cpp/rdf/query/QuadPattern.cpp
TortugaAttack: fair enough. 
I was thinking about that it is actually called context, but you're right.

src/rdf4cpp/rdf/storage/node/BNodeBackend.cpp
bigerl: N is taken from the n-formats N-Triples/N-Quads. I named it n_string to make clear which serialization the user can expect.

src/rdf4cpp/rdf/storage/node/LiteralBackend.cpp
bigerl: agreed. I would add an todo for now. I need to re-check specs on this. If we escape `"` here I would also escape everything else which is not allowed in N-Tripels/N-Quads (if there is anything else). I expect this function to be used by those serialisations, mainly.

bigerl:
> Some stuff i annotated.
> Documentation is missing for a lot of these classes. Will be a real hassle to add them later on. :D

Yep. You are right. I will add more documentation. I started only with Term so far, I think.  
Still a bit todo :)

src/rdf4cpp/rdf/Quad.cpp
bigerl: Ah, or better: we have already a default constructor which initialized Quad with null Terms. So we can validate this constructor always.

bigerl: or, like you commented below: we have a static factory method which does the validation and returns an optional. So we would need to explicitly check if it was valid. The constructor could stay non-validating then.

TortugaAttack: sounds good to me, so if one really knows what they are doing, they can use the constructor directly (e.g. if they know it is definitely valid) 
otherwise the factory method can be used.

bigerl: please re-review the updated class.

src/rdf4cpp/rdf/Statement.cpp
bigerl: (see above)

src/rdf4cpp/rdf/Quad.cpp
TortugaAttack: seems good

(we may want to use a rdf4cpp::sec::Result in the future to let the user know what was wrong, but that is debatable and future work)

TortugaAttack:
that was fast :D 
looks good to me

PR #21, opened 2022-01-25
Better interface for custom node storage

bigerl:
This introduces wide changes and restructuring to node storage backend, including:

- introduced new sub-namespaces `storage::node::identifier`, `storage::node::handle` and  `storage::node::default_node_storage` for clearer separation of functional units
- `storage::node::handle`s Views provide a backed-agnostic view on actual backend storage node (using `std::string_view`)
- use `std::string_view` wherever possible to be able to be able use in the backend different continuous memory string representations (e.g. `boost::string` which supports custom allocators)
- node and types that are used transitively as fields have now noexcept constructors and comparison functions (this increases performance with some collections)


nkaralis:
- Should the `DefaultNodeStorageBackend::lookup_or_insert` functions be split into two functions (i.e., lookup and insert)? Could there be a use case where we want to just lookup for a Node and not insert it in the Storage in case it is missing?
- The backend functions of `NodeBackendHandle` are marked as `noexcept`. However, a `Node *` holding an IRI is able to call `NodeBackendHandle::bnode_backend`, which will throw an exception. It might be a good idea to have dedicated exceptions for these functions.
- Is the static cast in line 81 of file NodeID.cpp necessary?
- When a new `DefaultNodeStorageBackend` is constructed, its next_values are initialized with the minimum_values of `NodeID`. Is it okay for a Node to have different identifiers in different backends?

bigerl:
Thank you very much for the review. I'll address the points below:

> * Should the `DefaultNodeStorageBackend::lookup_or_insert` functions be split into two functions (i.e., lookup and insert)? Could there be a use case where we want to just lookup for a Node and not insert it in the Storage in case it is missing?

It would definitely make sense to have those two versions. The lookup might for example be relevant if we want to decide if a SPARQL query string has a trivially empty answer because it contains triple with nodes which are not in storage. 

>     * The backend functions of `NodeBackendHandle` are marked as `noexcept`. However, a `Node *` holding an IRI is able to call `NodeBackendHandle::bnode_backend`, which will throw an exception. It might be a good idea to have dedicated exceptions for these functions.

Right, we definitely will need to handle that. Another option would be to return a null-pointer and test on it. 

>     * Is the static cast in line 81 of file NodeID.cpp necessary?

right. it's unnecessary.

>     * When a new `DefaultNodeStorageBackend` is constructed, its next_values are initialized with the minimum_values of `NodeID`. Is it okay for a Node to have different identifiers in different backends?

Yes. Unfortunately, that is kind of a necessity. Hash is not an option because it would be hard to handle hash collisions. Also, the IDs of two colliding Nodes could be different in two subsequent runs when they are inserted in different order. That would also make it more complicated to implement a presistent NodeStorage Backend. 


I'll address the first two points and would come back to you then.

bigerl:
superseded by #22

PR #22, opened 2022-02-01
Better interface for custom node storage (Draft 2)

bigerl:
this supersedes #21


src/rdf4cpp/rdf/storage/node/NodeStorage.hpp
```diff
@@ -17,88 +17,332 @@ namespace rdf4cpp::rdf::storage::node {
...
 
+/**
+ * NodeStorage provides an interface to the internal storage <div>Node</div>s.
+ * Each NodeStorage has a INodeStorageBackend, which is uniquely identified by a identifier::NodeStorageID.
+ * There can be at most 1024 different INodeStorageBackend instances at once.
+ * If no NodeStorage with the identifier::NodeStorageID of a certain INodeStorageBackend exists anymore, the corresponding INodeStorageBackend is destructed.
+ * This does not apply to the default_instance.
+ * The live cycle of NodeStorage's and their backends is managed by the static methods within this class.
+ */
 class NodeStorage {
```
bigerl, line 28: @nkaralis I tried to address the points from your review https://github.com/rdf4cpp/rdf4cpp/pull/21#issuecomment-1025506011 . Could you check if you think that this reworked interface is an improvement and reevaluate?

bigerl:
Things to address before merge:

- [x] make sure null() NodeID's are nowhere used (could be that they are currently used for default graph IRIs).
- [x] fix race condition in INodeStorageBackend
- [x] document INodeStorageBackend
- [x] document ReferenceNodeStorageBackend
- [x] document Views
- [x] document identifier::*
- [x] investigate performance regression

src/rdf4cpp/rdf/storage/node/NodeStorage.hpp
nkaralis: The split of the functions looks good to me.

nkaralis:
I have a few comments on the new additions. 

- First of all, I found the documentation very helpful. 
- Regarding NodeStorageID, I think its documentation has conflicts. The `identifier/README.md` states that NodeStorageID is used to identify the NodeStorage of a Node. However, the file `NodeStorageID.hpp` states that a NodeStorageID identifies a Literal of a given type within a given NodeStorage. Is the usage of NodeStorageID limited only to Literals?
- In `ReferenceBackendImplementation.hpp/.cpp` the functions `find_*_backend_view()` are marked `noexcept`, however they use `std::map::at method`, which throws an exception if the provided key is missing. Would it be better to first check with `std::map::contains` if the provided key exists? In case the provided key is missing, we could return a `nullptr`

bigerl:
> I have a few comments on the new additions.
> 
>     * First of all, I found the documentation very helpful.
> 
>     * Regarding NodeStorageID, I think its documentation has conflicts. The `identifier/README.md` states that NodeStorageID is used to identify the NodeStorage of a Node. However, the file `NodeStorageID.hpp` states that a NodeStorageID identifies a Literal of a given type within a given NodeStorage. Is the usage of NodeStorageID limited only to Literals?
> 
>     * In `ReferenceBackendImplementation.hpp/.cpp` the functions `find_*_backend_view()` are marked `noexcept`, however they use `std::map::at method`, which throws an exception if the provided key is missing. Would it be better to first check with `std::map::contains` if the provided key exists? In case the provided key is missing, we could return a `nullptr`

I've addressed all points from this review. `find_*_backend_view` are in the current code not noexcept but are expected to throw an exception if the id does not exist. This is documented in the interface definition at INodeStorageBackend.hpp. 

Thanks a lot for reading the documentation so carefully. I would not have found have overlooked the forst flaw you mentioned.

PR #27, opened 2022-05-23
RDF Core Types & Testcases

kaimal11:
implementation of string, boolean, decimal, and integer datatypes with test cases


tests/datatype/tests_Integer.cpp
```diff
@@ -0,0 +1,60 @@
...
+    CHECK(lit5.value<datatypes::xsd::Integer>() == value);
+
+    auto lit6 = Literal{"+1", type_iri};
+    CHECK(lit6.value<datatypes::xsd::Integer>() == 1);
+
+    CHECK(lit1 != lit2);
+    CHECK(lit2 != lit3);
+    CHECK(lit1 == lit4);
+    CHECK(lit3 == lit5);
+    //CHECK(lit4 == lit6);
```
nkaralis: In this case, the string representation should be different. You can use `value()` to check if "1" and "+1" are handled in the same way

tests/datatype/tests_Integer.cpp
```diff
@@ -0,0 +1,60 @@
...
+    auto lit1 = Literal::make<datatypes::xsd::Integer>(value);
+    CHECK(lit1.value<datatypes::xsd::Integer>() == value);
+    CHECK(lit1.lexical_form() == std::to_string(value));
+
+    value = -2147483648;
+    auto lit2 = Literal::make<datatypes::xsd::Integer>(value);
+    CHECK(lit2.value<datatypes::xsd::Integer>() == value);
+    CHECK(lit2.lexical_form() == std::to_string(value));
+
+    value = 2147483647;
```
nkaralis, line 32: This is the limit for `int32_t`. If you wanted to check the limits, you should use the `int64_t`.

tests/datatype/tests_String.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    value = "\n";
+    auto lit6 = Literal{value, type_iri};
+    CHECK(lit6.value<datatypes::xsd::String>() == value);
+
+    value = "\t";
+    auto lit7 = Literal{value, type_iri};
+    CHECK(lit7.value<datatypes::xsd::String>() == value);
+
+    value = "\r";
```
nkaralis, line 53: It would be nice if you could add tests for more escaped characters. Check Literal.cpp for all characters that are currently being escaped.

src/rdf4cpp/rdf/datatypes/xsd.hpp
```diff
@@ -1,8 +1,12 @@
 #ifndef RDF4CPP_XSD_HPP
 #define RDF4CPP_XSD_HPP
 
+#include <rdf4cpp/rdf/datatypes/xsd/Boolean.hpp>
+#include <rdf4cpp/rdf/datatypes/xsd/Decimal.hpp>
+#include <rdf4cpp/rdf/datatypes/xsd/Integer.hpp>
+#include <rdf4cpp/rdf/datatypes/xsd/String.hpp>
 #include <rdf4cpp/rdf/datatypes/xsd/Float.hpp>
-#include <rdf4cpp/rdf/datatypes/xsd/Int.hpp>
+/*#include <rdf4cpp/rdf/datatypes/xsd/Int.hpp>*/
```
nkaralis, line 9: Why is this commented out?

src/rdf4cpp/rdf/datatypes/xsd/Int.hpp
```diff
@@ -13,13 +13,14 @@ namespace rdf4cpp::rdf::datatypes::xsd {
...
 namespace rdf4cpp::rdf::datatypes {
+constexpr const char xsd_int[] = "http://www.w3.org/2001/XMLSchema#int";
+
 template<>
-inline std::string RegisteredDatatype<xsd::Int>::datatype_iri() noexcept { return "http://www.w3.org/2001/XMLSchema#int"; }
+inline std::string RegisteredDatatype<xsd::Int, xsd_int>::datatype_iri() noexcept { return "http://www.w3.org/2001/XMLSchema#int"; }
 
 template<>
-inline xsd::Int RegisteredDatatype<xsd::Int>::from_string(std::string_view s) {
+inline xsd::Int RegisteredDatatype<xsd::Int, xsd_int>::from_string(std::string_view s) {
```
nkaralis, line 23: Datatype for Int is not using the new registry

tests/datatype/tests_Decimal.cpp
```diff
@@ -0,0 +1,68 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+
+#include <doctest/doctest.h>
+#include <rdf4cpp/rdf.hpp>
+
+using namespace rdf4cpp::rdf;
+
+TEST_CASE("Datatype Decimal") {
```
nkaralis, line 8: You need to write test cases that check how we handle trailing zeros in the decimal part. The standard says that "1.0" and "1.00" are not distinct.

src/rdf4cpp/rdf/datatypes/xsd/Float.hpp
```diff
@@ -6,21 +6,57 @@
...
+};
 
-namespace rdf4cpp::rdf::datatypes {
+/**
+ * Specialisation of from_string template function.
+ */
 template<>
-inline std::string RegisteredDatatype<xsd::Float>::datatype_iri() noexcept { return "http://www.w3.org/2001/XMLSchema#float"; }
+inline LiteralDatatypeImpl<xsd_float>::cpp_type LiteralDatatypeImpl<xsd_float>::from_string(std::string_view s) {
+    return std::stof(std::string{s.data()});
```
nkaralis, line 35: Does this handle NaN and INF? You need to check this with your test cases. In addition, the regular expression is missing.

nkaralis (changes requested):

src/rdf4cpp/rdf/datatypes/xsd/Int.hpp
kaimal11: This was the earlier implementation, included with the base code. We have commented it in the datatype registry, and the current PR does not include its exact implementation. Basic Types only include: Boolean, String, Integer, and Decimal

src/rdf4cpp/rdf/datatypes/xsd/Float.hpp
kaimal11: The exact implementation of these checks will be included in the PR of IEEE Float Point Numbers. Changes were made to the original files just to make it run.

src/rdf4cpp/rdf/datatypes/xsd.hpp
kaimal11: This PR was just for Basic Datatypes. Int will be pushed with PR of Limited Range Numbers

tests/datatype/tests_Decimal.cpp
```diff
@@ -33,36 +33,49 @@ TEST_CASE("Datatype Decimal") {
...
+    value = std::numeric_limits<double>::max();
+    auto lit7 = Literal{std::to_string(value), type_iri};
+    CHECK(lit7.value<datatypes::xsd::Decimal>() == value);
+
     CHECK(lit1 != lit2);
     CHECK(lit2 != lit3);
     CHECK(lit1 == lit4);
     CHECK(lit2 == lit5);
+    CHECK(lit1 == lit6);
+    CHECK(lit4 == lit4);
```
nkaralis: Is the last CHECK intended to be this way?

nkaralis (approved):

PR #29, opened 2022-06-06
Improved cmake, introduced conan recipe and imporved CI

bigerl:
Complete polishing and overhaul of the the building, installing and packaging processes. 

CMake
- simplified
- removed the hack to build both, shared and static library at the same time
- moved install and export definition to a function in code to `cmake/install_library.cmake`
- added test for FetchContent

Conan
- introduced conanfile.py with basic package config
- include also test_package folder

CI:
- updated build matrix to compilers supported on ubuntu 22.04
- added conan package build test
- added cmake FetchContent test


bigerl:
@LukasKerk I would merge this now because we need to CI to fix another failing test. I would still ask you to have a look if the cmake and conan setup is all fine. Please check the develop branch as this one will be deleted.

PR #31, opened 2022-06-12
Fixes test cases and Boolean datatype

kaimal11:


src/rdf4cpp/rdf/datatypes/xsd/Decimal.hpp
```diff
@@ -49,10 +49,8 @@ template<>
 inline std::string LiteralDatatypeImpl<xsd_decimal>::to_string(const cpp_type &value) {
 
     std::ostringstream str_os;
-    // Set Fixed -Point Notation
-    str_os << std::fixed;
-    str_os << value;
-    // Get string from output string stream
+    // Set Fixed -Point Notation for Decimal
+    str_os << std::fixed << value;
```
bigerl: This does not produce a canonical representation. We need "2.0" not "2.000000". See: https://www.w3.org/TR/xmlschema-2/#decimal

You could add fmt 8.1.1 via [fetchcontent](https://cmake.org/cmake/help/latest/module/FetchContent.html) and link it as PRIVATE to rdf4cpp's target link libraries. (If it helps). If you use fmt, you should add an cpp file for Decimal.hpp and include the `<fmt/format.h>` only in that cpp file. The rational behind this is that we do not want to include all of fmt's headers with rdf4cpp. 

The canonization should produce the following results: 
- "0.000000031" -> "0.000000031"
- "1", "1.0" -> "1.0"
- "10000000" -> "10000000.0"

Those resources might also be relevant:
- https://stackoverflow.com/questions/1521607/check-double-variable-if-it-contains-an-integer-and-not-floating-point
- https://www.zverovich.net/2019/02/11/formatting-floating-point-numbers.html

tests/datatype/tests_Boolean.cpp
```diff
@@ -18,34 +18,49 @@ TEST_CASE("Datatype Boolean") {
...
+    CHECK(lit3.lexical_form() == true_val);
 
     value = 0;
     auto lit4 = Literal::make<datatypes::xsd::Boolean>(value);
     CHECK(lit4.value<datatypes::xsd::Boolean>() == value);
+    CHECK(lit4.lexical_form() == false_val);
 
     value = true;
-    auto lit5 = Literal{std::to_string(value), type_iri};
+    auto lit5 = Literal{"true", type_iri};
```
bigerl: use `true_value`/`false_value` everywhere

tests/nodes/tests_Literal.cpp
```diff
@@ -65,14 +65,16 @@ TEST_CASE("Literal - Check for lexical form with IRI") {
...
         auto lit1 = Literal{"2.0", iri};
 
+        std::string value = std::to_string(lit1.value<datatypes::xsd::Decimal>());
+
         CHECK(not lit1.is_blank_node());
         CHECK(lit1.is_literal());
         CHECK(not lit1.is_variable());
         CHECK(not lit1.is_iri());
-        CHECK(lit1.lexical_form() == "2.0");
+        CHECK(lit1.lexical_form() == value);
```
bigerl: it would be best to rename `value` to `canonical_lexical` and assign it a fixed value of `"2.0"`.

bigerl (changes requested):
Integer and Boolean look mostly fine. Decimal still needs some work. Please see the comments below.

src/rdf4cpp/rdf/datatypes/xsd/Decimal.hpp
```diff
@@ -49,18 +50,30 @@ inline LiteralDatatypeImpl<xsd_decimal>::cpp_type LiteralDatatypeImpl<xsd_decima
 template<>
 inline std::string LiteralDatatypeImpl<xsd_decimal>::to_string(const cpp_type &value) {
 
-    double int_part;
-    auto fract_part = modf(value, &int_part);
+    double int_part, fract_part;
+    fract_part  = modf(value, &int_part);
+    bool flag = false;
```
bigerl: rename `flag` to: `remove_trailing_zeros`

src/rdf4cpp/rdf/datatypes/xsd/Decimal.hpp
```diff
@@ -49,18 +50,30 @@ inline LiteralDatatypeImpl<xsd_decimal>::cpp_type LiteralDatatypeImpl<xsd_decima
...
     std::ostringstream str_os;
     str_os << std::fixed;
-    if (fract_part == 0) str_os << std::setprecision(1);
+    if (fract_part == 0) {
```
bigerl, line 58: add comments to every branch explaining what it does

bigerl (changes requested):
looks good. 👍🏻 

The requested changes are just supposed to increase code readability.

bigerl (approved):

PR #34, opened 2022-06-21
Supports IEEE Floating Point Number Datatypes

kaimal11:
Implements xsd::Float and xsd::Double datatypes

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/xsd/Double.hpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/Double.hpp
@@ -1,9 +1,10 @@
+
 /**
- * @file Registers xsd:float with DatatypeRegistry
+ * @file Registers xsd:double with DatatypeRegistry
  */
 
-#ifndef RDF4CPP_XSD_FLOAT_HPP
-#define RDF4CPP_XSD_FLOAT_HPP
+#ifndef RDF4CPP_XSD_DOUBLE_HPP
+#define RDF4CPP_XSD_DOUBLE_HPP
 
 #include <rdf4cpp/rdf/datatypes/registry/DatatypeMapping.hpp>
 #include <rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp>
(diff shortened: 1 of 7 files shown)


src/rdf4cpp/rdf/datatypes/xsd/Double.hpp
```diff
@@ -0,0 +1,91 @@
...
+ */
+template<>
+inline std::string LiteralDatatypeImpl<xsd_double>::to_string(const cpp_type &value) {
+
+    if (isnan(value)) {
+        //checks whether value is "nan", then returns rdf representation of nan
+        return "NaN";
+    } else if (isinf(value)) {
+        //checks whether value is "inf", then returns rdf representation of inf
+        return "INF";
```
nkaralis: We need to also handle -INF.

src/rdf4cpp/rdf/datatypes/xsd/Float.hpp
```diff
@@ -41,17 +47,40 @@ inline LiteralDatatypeImpl<xsd_float>::cpp_type LiteralDatatypeImpl<xsd_float>::
...
-    str_os << value;
-    // Get string from output string stream
-    std::string str = str_os.str();
-    return str;
+    if (isnan(value)) {
+        //checks whether value is "nan", then returns rdf representation of nan
+        return "NaN";
+    } else if (isinf(value)) {
+        //checks whether value is "inf", then returns rdf representation of inf
+        return "INF";
```
nkaralis, line 58: As in Double.hpp, we need to also handle -INF.

tests/datatype/tests_Double.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    auto lit8 = Literal{"NaN", type_iri};
+    CHECK(isnan(lit8.value<datatypes::xsd::Double>()));
+
+    auto lit9 = Literal{"INF", type_iri};
+    CHECK(isinf(lit9.value<datatypes::xsd::Double>()));
+
+    value = INFINITY;
+    auto lit10 = Literal::make<datatypes::xsd::Double>(value);
+    CHECK(isinf(lit10.value<datatypes::xsd::Double>()));
```
nkaralis, line 62: We need to also check for -INF

nkaralis (changes requested):

bigerl:
we should have another look at precision before we merge 💡

PR #37, opened 2022-06-30
Support for Miscellaneous XSD Datatypes

kaimal11:
implements anyURI, language, normalizedString, token, NMToken, name, and NCName datatypes

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/xsd/NCName.hpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/NCName.hpp
@@ -1,0 +1,50 @@
+
+/**
+ * @file Registers xsd:NCName with DatatypeRegistry
+ */
+
+#ifndef RDF4CPP_XSD_NCNAME_HPP
+#define RDF4CPP_XSD_NCNAME_HPP
+
+#include <rdf4cpp/rdf/datatypes/registry/DatatypeMapping.hpp>
+#include <rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp>
+
+#include <cstdint>
+#include <ostream>
+#include <regex>
+
+namespace rdf4cpp::rdf::datatypes::registry {
+/*
+ * Name of the datatype. This is kept so that we won't need to type it over and over again.
+ */
+constexpr static registry::ConstexprString xsd_NCName{"http://www.w3.org/2001/XMLSchema#NCName"};
+
+/**
+ * Defines the mapping between the LiteralDatatype IRI and the C++ datatype.
+ */
+template<>
+struct DatatypeMapping<xsd_NCName> {
+    using cpp_datatype = std::string;
+};
+
+/**
+ * Specialisation of from_string template function.
+ */
+template<>
+inline LiteralDatatypeImpl<xsd_NCName>::cpp_type LiteralDatatypeImpl<xsd_NCName>::from_string(std::string_view s) {
+    const std::regex non_ncName_regex("(\\n)|(\\t)|(\\r)|([ ]{2,})|(^[ ].*)|(.*[ ]$)|:");
+    if (std::regex_search(s.data(), non_ncName_regex)) {
+        throw std::runtime_error("XSD Parsing Error");
+    } else {
+        return s.data();
+    }
+}
+}  // namespace rdf4cpp::rdf::datatypes::registry
+namespace rdf4cpp::rdf::datatypes::xsd {
+/**
+ * Implementation of xsd::NCName
+ */
+using NCName = registry::LiteralDatatypeImpl<registry::xsd_NCName>;
+}  // namespace rdf4cpp::rdf::datatypes::xsd
+
+#endif  //RDF4CPP_XSD_NCNAME_HPP
--- a/src/rdf4cpp/rdf/datatypes/xsd/NMTOKEN.hpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/NMTOKEN.hpp
@@ -1,0 +1,50 @@
+
+/**
+ * @file Registers xsd:NMTOKEN with DatatypeRegistry
+ */
+
+#ifndef RDF4CPP_XSD_NMTOKEN_HPP
+#define RDF4CPP_XSD_NMTOKEN_HPP
+
+#include <rdf4cpp/rdf/datatypes/registry/DatatypeMapping.hpp>
+#include <rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp>
+
+#include <cstdint>
+#include <ostream>
+#include <regex>
+
+namespace rdf4cpp::rdf::datatypes::registry {
+/*
+ * Name of the datatype. This is kept so that we won't need to type it over and over again.
+ */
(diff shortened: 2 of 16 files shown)


src/rdf4cpp/rdf/datatypes/xsd/NCName.hpp
```diff
@@ -0,0 +1,50 @@
...
+struct DatatypeMapping<xsd_NCName> {
+    using cpp_datatype = std::string;
+};
+
+/**
+ * Specialisation of from_string template function.
+ */
+template<>
+inline LiteralDatatypeImpl<xsd_NCName>::cpp_type LiteralDatatypeImpl<xsd_NCName>::from_string(std::string_view s) {
+    const std::regex non_ncName_regex("(\\n)|(\\t)|(\\r)|([ ]{2,})|(^[ ].*)|(.*[ ]$)|:");
```
bigerl, line 35: double check if the regex is fine and covers all cases. Add comments that explain what happens here

src/rdf4cpp/rdf/datatypes/xsd/NMTOKEN.hpp
```diff
@@ -0,0 +1,50 @@
...
+struct DatatypeMapping<xsd_NMTOKEN> {
+    using cpp_datatype = std::string;
+};
+
+/**
+ * Specialisation of from_string template function.
+ */
+template<>
+inline LiteralDatatypeImpl<xsd_NMTOKEN>::cpp_type LiteralDatatypeImpl<xsd_NMTOKEN>::from_string(std::string_view s) {
+    const std::regex non_nmToken_regex("(\\n)|(\\t)|(\\r)|([ ]{2,})|(^[ ].*)|(.*[ ]$)");
```
bigerl, line 35: double check if the regex is fine and covers all cases. Add comments that explain what happens here

src/rdf4cpp/rdf/datatypes/xsd/Name.hpp
```diff
@@ -0,0 +1,50 @@
...
+struct DatatypeMapping<xsd_name> {
+    using cpp_datatype = std::string;
+};
+
+/**
+ * Specialisation of from_string template function.
+ */
+template<>
+inline LiteralDatatypeImpl<xsd_name>::cpp_type LiteralDatatypeImpl<xsd_name>::from_string(std::string_view s) {
+    const std::regex non_name_regex("(\\n)|(\\t)|(\\r)|([ ]{2,})|(^[ ].*)|(.*[ ]$)");
```
bigerl, line 35: double check if the regex is fine and covers all cases. Add comments that explain what happens here

src/rdf4cpp/rdf/datatypes/xsd/NormalizedString.hpp
```diff
@@ -0,0 +1,50 @@
...
+struct DatatypeMapping<xsd_normalizedString> {
+    using cpp_datatype = std::string;
+};
+
+/**
+ * Specialisation of from_string template function.
+ */
+template<>
+inline LiteralDatatypeImpl<xsd_normalizedString>::cpp_type LiteralDatatypeImpl<xsd_normalizedString>::from_string(std::string_view s) {
+    const std::regex non_ns_regex("(\\n)|(\\t)|(\\r)");
```
bigerl, line 35: check that in a for loop not with a regex

src/rdf4cpp/rdf/datatypes/xsd/Token.hpp
```diff
@@ -0,0 +1,50 @@
...
+struct DatatypeMapping<xsd_token> {
+    using cpp_datatype = std::string;
+};
+
+/**
+ * Specialisation of from_string template function.
+ */
+template<>
+inline LiteralDatatypeImpl<xsd_token>::cpp_type LiteralDatatypeImpl<xsd_token>::from_string(std::string_view s) {
+    const std::regex non_token_regex("(\\n)|(\\t)|(\\r)|([ ]{2,})|(^[ ].*)|(.*[ ]$)");
```
bigerl, line 35: check in a for loop

tests/datatype/tests_NCName.cpp
```diff
@@ -0,0 +1,94 @@
...
+    auto lit2 = Literal::make<datatypes::xsd::NCName>(value);
+    CHECK(lit2.value<datatypes::xsd::NCName>() == value);
+    CHECK(lit2.lexical_form() == value);
+
+    value = "Bob";
+    auto lit3 = Literal::make<datatypes::xsd::NCName>(value);
+    CHECK(lit3.value<datatypes::xsd::NCName>() == value);
+    CHECK(lit3.lexical_form() == value);
+
+    value = "12sdfs2df";
```
bigerl, line 36: this must not work https://stackoverflow.com/questions/1631396/what-is-an-xsncname-type-and-when-should-it-be-used

bigerl:
We do not plan to support these data types ATM. They are not relevant for RDF. If someone thinks they are important please update an issue.

AtesComp:
In partial support for inclusion, see https://www.w3.org/2011/rdf-wg/wiki/XSD_Datatypes

Several are well supported in many platforms, such as
xsd:anyURI
xsd:byte
xsd:hexBinary

This also affect issue #128

bigerl:
xsd:byte and xsd:hexBinary are already supported by rdf4cpp with native C++ types backing them. The remaining types seem to be mostly targeted towards XML document structure. Are you aware of any RDF related use-cases that actually use, e.g., xsd:anyURI?

AtesComp:
I didn't see a mapping of these XSD type to the native C++ types. As long as there is a read/write path to and from a data store to recognized and reproduce these types (round trip), all is good.

The `anyURI` type (as well as the others) is used even within RDF standards.
See https://www.w3.org/TR/rdf12-xml/
See https://www.w3.org/TR/rdf11-concepts/ section 5.1
There are many RDF documents that use `anyURI` as a data type.

bigerl:
they are here:
- [xsd:hexBinary](https://github.com/rdf4cpp/rdf4cpp/blob/develop/src/rdf4cpp/datatypes/xsd/HexBinary.hpp) 
- [xsd:byte](https://github.com/rdf4cpp/rdf4cpp/blob/develop/src/rdf4cpp/datatypes/xsd/integers/signed/Byte.hpp)

We just discussed `xsd:anyURI` and will add it which includes value validation. SPARQL actually requires the cast to boolean operator. No idea, though, who would ever use that operator. 

Maybe a final note regarding what "adding" means here. It is still possible to instantiate the other types. The constructor will just not validate the lexical form and no type-specific functions will be available.

liss-h:
@AtesComp 

> The `anyURI` type (as well as the others) is used even within RDF standards. See https://www.w3.org/TR/rdf12-xml/ See https://www.w3.org/TR/rdf11-concepts/ section 5.1 There are many RDF documents that use `anyURI` as a data type.

We've had a look through the specs again and decided that it makes sense to explicitly support `xsd:anyURI` with a datatype specialization as it is also mentioned in the SPARQL spec. The other datatypes will probably not get a datatype specialization from us as they are either only relevant for XML or basically just a string. They can still be loaded and they do roundtrip to serialization, there is just no validation or extra functionality.

If you require some additional functionalities on the other datatypes, feel free to open a PR to implement them.

PR #49, opened 2022-07-18
Fixing buffer overread UB in from_string implementations

liss-h:


liss-h:
The failing tests are yet again GCC-10. For some reason std::from_chars is incorrectly implemented there.

bigerl (changes requested):
Looks good. 👍🏻  Only question remaining: Why is [Int.hpp](https://github.com/rdf4cpp/rdf4cpp/blob/feature/fix-from-string-ub/src/rdf4cpp/rdf/datatypes/xsd/Int.hpp) not using `std::from_chars`?

liss-h:
> Looks good. 👍🏻 Only question remaining: Why is [Int.hpp](https://github.com/rdf4cpp/rdf4cpp/blob/feature/fix-from-string-ub/src/rdf4cpp/rdf/datatypes/xsd/Int.hpp) not using `std::from_chars`?

Because Int.hpp isn't even implemented correctly, and also not used anywhere as far as I can tell. So the whole file needs to be fixed first. I just thought that is a seperate issue but I can fix that.

bigerl (approved):

PR #50, opened 2022-08-02
Feature: Literal ops and conversions

liss-h:


src/rdf4cpp/rdf/datatypes/registry/Util.hpp
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Map each element in a tuple using a unary operation f
+ * and store the results in a another tuple.
+ *
+ * @param tup a std::tuple like object to map over
+ * @param f a unary operation to map the elements
+ * @return the tuple generated by applying f to each element in tup
+ */
+template<typename F, typename TupleLike>
+constexpr auto tuple_map(TupleLike &&tup, F &&f) {
```
bigerl: A concept for tuple would make compiler errors more readable.

src/rdf4cpp/rdf/datatypes/registry/Util.hpp
```diff
@@ -0,0 +1,112 @@
...
+ * The provided function f will be repeatedly called as f(acc, element) for each element in the
+ * tuple and is expected to return the acc for the next invocation.
+ *
+ * @tparam Acc the type of the accumulator
+ * @param tup a std::tuple like object to fold
+ * @param init the initial value of the accumulator
+ * @param f a binary operation that will be called with the accumulator and the current element and should return the new accumulator
+ * @return the final accumulator value
+ */
+template<typename Acc, typename F, typename TupleLike>
```
bigerl: same as above, TupleLike could be a Concept

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
```diff
@@ -13,20 +13,23 @@
...
+ */
+template<typename>
+using always_false = std::false_type;
+template<typename T>
+static constexpr bool always_false_v = always_false<T>::value;
 
+} // namespace detail
 
-public:
+template<ConstexprString type_iri_t>
```
bigerl, line 43: Add documentation for Default

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
```diff
@@ -13,20 +13,23 @@
...
-private:
-    /**
-     * static_assert would always trigger if it wasn't dependent on a template parameter.
-     * With this helper template, it only triggers if the function is instantiated.
-     */
-    template<typename>
-    using always_false = std::false_type;
-    template<typename T>
-    static constexpr bool always_false_v = always_false<T>::value;
+namespace capabilities {
```
bigerl, line 15: Have you consider super class `Capabilities<ConstexprString type_iri_t>`? You could also make Default superclass to the others. Please share your opinion in this.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -153,5 +155,225 @@ Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::
...
+    }
+
+    assert(this->handle_.is_literal() && other.handle_.is_literal());
+
+    auto const after_conversion = [&]() -> std::optional<std::tuple<std::string_view, std::any, std::any>> {
+        auto const this_lit_type = this->handle_.literal_backend().datatype_id;
+        auto const other_lit_type = other.handle_.literal_backend().datatype_id;
+
+        if (this_lit_type == other_lit_type) {
+            return std::make_tuple(this->datatype().identifier(), this->value(), other.value());
```
bigerl: This and subsequent calls to `this->datatype().identifier()` are expensive because a lookup is made in the backend. But I don't see any portable way right now how to handle it differently.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -153,5 +155,225 @@ Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::
...
+        } else {
+            auto const lhs_datatype = this->datatype().identifier();
+            auto const rhs_datatype = other.datatype().identifier();
```
bigerl: Sale here, use {this,other}_lit_type ti retrieve the identifier IRI

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -153,5 +155,225 @@ Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::
...
+        }
+    }();
+
+    if (!after_conversion.has_value()) {
+        return Literal{}; // datatype mismatch and not in same promotion hierarchy
+    }
+
+    auto const &[res_identifier, lhs_val, rhs_val] = *after_conversion;
+
+    auto const result_entry = datatypes::registry::DatatypeRegistry::get_entry(res_identifier).value().get();
```
bigerl: Please double check that during every operation the *BackendView for each operand, and the type of each operand and, - if different - the type of the result is  retrieved at most once.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -153,5 +155,225 @@ Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::
...
 
+template<typename OpSelect>
+Literal Literal::numeric_binop_impl(OpSelect op_select, Literal const &other, NodeStorage &node_storage) const {
+    if (this->null() || other.null()) {
+        return Literal{};
+    }
+
+    assert(this->handle_.is_literal() && other.handle_.is_literal());
+
+    auto const after_conversion = [&]() -> std::optional<std::tuple<std::string_view, std::any, std::any>> {
```
bigerl: That might be something for a seperate TODO: It would be better to use here the the datatype_id directly. 

Currently, the datatype_id is fixed only from some types but not in general. We would need to enforce a fixed datatype_id for all types that support non-trivial operatotions.

bigerl: How about a sperate for AfterConversion? From the types alone it's not directly clear what the positions stand for.

tests/nodes/tests_Literal_ops.cpp
```diff
@@ -0,0 +1,429 @@
...
+
+    // simple subtype substitution test cases
+    GENERATE_BINOP_TESTCASE(Integer, 6, +, Int, 7, Integer, 13);
+    GENERATE_BINOP_TESTCASE(Int, 8, *, Integer, 9, Integer, 72);
+
+    // operator assignment test cases
+    GENERATE_BINOP_TESTCASE(Integer, 1, +, Integer , 2, Integer, 3);
+    GENERATE_BINOP_TESTCASE(Integer, 12, -, Integer, 1, Integer, 11);
+    GENERATE_BINOP_TESTCASE(Integer, 3, *, Integer, 6, Integer, 18);
+    GENERATE_BINOP_TESTCASE(Integer, 12, /, Integer, 4, Integer, 3); // TODO: use idiv or div? currently using idiv
```
bigerl: it must be div. idiv is not used. Instead: 

> but xsd:decimal if both operands are xsd:integer

My understanding would be to promote to `xsd:decimal` and do div there.

bigerl (changes requested):
thx for the splendid work. 
- I think the main point that needs to be adressed is minimizing the calls into NodeStorage. 
- Also, div for xsd:integer with  promotion to xsd:decimal, might need some more work. I have not yet double checked the actual behaviour.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -153,5 +155,225 @@ Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::
...
+
+    if (!after_conversion.has_value()) {
+        return Literal{}; // datatype mismatch and not in same promotion hierarchy
+    }
+
+    auto const &[res_identifier, lhs_val, rhs_val] = *after_conversion;
+
+    auto const result_entry = datatypes::registry::DatatypeRegistry::get_entry(res_identifier).value().get();
+
+    if (!result_entry.numeric_ops.has_value()) {
```
bigerl: move up because we do not expect type hierarchies where we have non-numeric types which have might be cast to numeric types.

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
liss-h: I thought about it but I don't see what value that adds, since typedefs are not "inherited" by default, so
you have to put in a using decl anyways (if that was your intention).

src/rdf4cpp/rdf/datatypes/registry/DatatypeConversion.hpp
```diff
@@ -0,0 +1,355 @@
...
+        // The same problem exists when a supertype (or promoted type) has a subtype (promotion) rank that is not exactly one less than the origin type's rank.
+        static_assert(RankRule<Type, typename next::converted>::value,
+                      "detected invalid hierarchy, would not be able to discover all conversions");
+
+        // conversion must preserve numericity, so:
+        // NumericLiteralDatatype<Type> must imply NumericLiteralDatatype<typename next::converted>
+        // This is because the numeric operations assume that if any of the operands is not numeric
+        // the common type and the result type will also not be numeric to allow for an early return
+        // without doing the actual conversion.
+        static_assert(!NumericLiteralDatatype<Type> || NumericLiteralDatatype<typename next::converted>,
```
liss-h: Technically this assertion is not strong enough because it means even if the result would be comparable
if one of the source types is not we do an early (null) return in the literal operations, which is potentially unexpected behaviour.

Should I change the one-sided implication to a two sided one? (So from -> to <->)
I mean it is technically not required for it to be two sided but maybe the behaviour would be more consistent(?).

Please share your thoughts.

bigerl: I would say better be too strict than having undefined behavior.

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -40,12 +40,17 @@ class Literal : public Node {
...
+     * @return the effective-boolean-value of this literal as a TriStateBool
+     */
     TriStateBool get_ebv_impl() const;
 
     /**
      * @brief the implementation for all logical, binary operations
      *
-     * @param logic_table the logic table for the binary operation accessed via logic_table[static_cast<size_t>(this->get_ebv_impl())][static_cast<size_t>(other.get_ebv_impl())]
+     * @param logic_table the logic table for the binary operation. It is accessed via
+     *                      logic_table[static_cast<size_t>(this->get_ebv_impl())][static_cast<size_t>(other.get_ebv_impl())].
```
bigerl: any technical reason for the indentation?

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -29,25 +29,30 @@ class DatatypeRegistry {
     using to_string_fptr_t = std::string (*)(const std::any &);
     using ebv_fptr_t = bool (*)(std::any const &);
 
-    using unop_fptr_t = std::any (*)(std::any const &);
-    using binop_fptr_t = std::any (*)(std::any const &, std::any const &);
+    struct NumericOpResult {
+        std::string_view result_type_iri;
+        std::any result_value;
+    };
```
bigerl, line 34: Good idea to encapsulate both together.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -157,78 +161,116 @@ Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::
...
+    auto const entry = DatatypeRegistry::get_entry(this->datatype().identifier());
+    assert(entry.has_value());
 
-    if (!entry.numeric_ops.has_value()) {
+    if (!entry->get().numeric_ops.has_value()) {
         return Literal{}; // datatype not numeric
     }
 
-    auto op_res = op_select(*entry.numeric_ops)(this->value());
+    DatatypeRegistry::NumericOpResult const op_res = op_select(*entry->get().numeric_ops)(this->value());
```
bigerl: `std::optional<std::reference_wrapper<X>>` is very verbose. Why not a raw pointer? Especially considering that the optional is not checked anyway.

src/rdf4cpp/rdf/datatypes/registry/Util.hpp
```diff
@@ -43,10 +91,10 @@ constexpr Acc tuple_type_fold_impl(Acc init, F &&f, std::index_sequence<Ixs...>)
  * @param f a unary operation to map the elements
  * @return the tuple generated by applying f to each element in tup
  */
-template<typename F, typename TupleLike>
-constexpr auto tuple_map(TupleLike &&tup, F &&f) {
-    constexpr size_t tup_size = std::tuple_size_v<std::remove_cvref_t<TupleLike>>;
-    return tuple_util_detail::tuple_map_impl(std::forward<TupleLike>(tup), std::forward<F>(f), std::make_index_sequence<tup_size>{});
+template<typename F, typename Tuple> requires AccessibleTuple<std::remove_cvref_t<Tuple>>
```
bigerl: `std::remove_csref_t` would still leave e.g. volatile. I would always use `std::decay_t`. Same below

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -104,8 +108,8 @@ std::partial_ordering Literal::operator<=>(const Literal &other) const {
     }
 }
 bool Literal::operator==(const Literal &other) const {
-    if (this->null() && other.null()) {
-        return true;
+    if (this->null() || other.null()) {
```
bigerl, line 111: good idea.

bigerl (changes requested):
I would say it is almost ready. Only some small things left to address. Please run format and optimize when everything else is done.

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
bigerl: yes, I forgot about the thing with the using declarations. I agree.

src/rdf4cpp/rdf/datatypes/registry/Util.hpp
liss-h: I don't think that is correct
see: https://en.cppreference.com/w/cpp/types/remove_cvref

src/rdf4cpp/rdf/Literal.cpp
liss-h: Yes, you are right. I was thinking in rust not c++ there 😅 (in rust Option<&T> is the idiomatic thing)

src/rdf4cpp/rdf/datatypes/registry/Util.hpp
bigerl: Right, v is for volatile. After reading through the docs again, I would prefer `std::remove_csref_t` over `std:decay_t`. Easier to forsee what we will get out of applying it.

src/rdf4cpp/rdf/Literal.hpp
liss-h: nope, that was just an accident

bigerl (approved):
time to merge

PR #51, opened 2022-08-05
Feature/namespaces

bigerl:
add support for generating IRIs from Namespace. This is inspired by https://rdflib.readthedocs.io/en/stable/apidocs/rdflib.namespace.html


src/rdf4cpp/rdf/Namespace.hpp
```diff
@@ -0,0 +1,87 @@
...
+     */
+    mutable NodeStorage node_storage_;
+
+    // TODO: a faster, less memory efficient map would be better.
+    /**
+     * Cache storing the <div>NodeBackendHandle</div> for prefixes. This saves roundtrips to NodeStorage.
+     */
+    storage::util::tsl::sparse_map<std::string, storage::node::identifier::NodeBackendHandle,
+                                   storage::util::robin_hood::hash<std::string>,
+                                   std::equal_to<std::string>>
```
liss-h: Using the transparent `std::equal_to<>` and `hash<std::string_view>` would be better because it allows the cache to be searched without constructing an `std::string` which also means `operator+` could take a `std::string_view` instead of a `std::string const &`.

see: https://en.cppreference.com/w/cpp/utility/functional/equal_to_void

```c++
storage::util::tsl::sparse_map<std::string, storage::node::identifier::NodeBackendHandle,
                                   storage::util::robin_hood::hash<std::string_view>,
                                   std::equal_to<>> cache_;
```

Maybe the hash functor would also require an `using is_transparent = void;` typedef but I'm not entirely certain of that. If you agree with the change but it turns out to be annoying to implement with unreadable compiler messages I can also try to make that work.

src/rdf4cpp/rdf/Namespace.hpp
```diff
@@ -0,0 +1,87 @@
...
+     * @param namespace_iri namespace IRI string. This will be used as prefix. IRI must not be encapsulated in <..>.
+     * @param node_storage where the IRIs will live
+     */
+    explicit Namespace(std::string_view namespace_iri,
+                       NodeStorage &node_storage = NodeStorage::default_instance());
+
+    /**
+     * @return IRI string of the Namespace
+     */
+    std::string const &names_space() const noexcept;
```
liss-h: very minor detail: this should probably return `std::string_view`. It doesn't really matter since it's a string anyways, but using `std::string_view` would prevent accidental copies.

Also: is the function name a typo?

src/rdf4cpp/rdf/namespaces/BFLC.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace rdf4cpp::rdf::namespaces {
+
+/**
+ * A ClosedNamespace for http://id.loc.gov/ontologies/bflc/ version http://id.loc.gov/ontologies/bflc-1-2-0/
+ */
+class BFLC : public ClosedNamespace {
+public:
+    static const std::string NAMESPACE;
+    static const std::vector<std::string> SUFFIXES;
```
liss-h: Could these be `static constexpr std::string_view NAMESPACE` and `static constexpr std::array<std::string_view, N> SUFFIXES`?

This also applies in other classes inheriting ClosedNamespace.

src/rdf4cpp/rdf/namespaces/RDF.cpp
```diff
@@ -0,0 +1,26 @@
...
+namespace rdf4cpp::rdf::namespaces {
+
+const std::string RDF::NAMESPACE = "http://www.w3.org/1999/02/22-rdf-syntax-ns#";
+const std::vector<std::string> RDF::SUFFIXES = {"HTML", "langString", "PlainLiteral", "type", "Property", "Statement", "subject", "predicate", "object", "Bag", "Seq", "Alt", "value", "List", "nil", "first", "rest", "XMLLiteral", "JSON", "CompoundLiteral", "language", "direction"};
+RDF::RDF(Namespace::NodeStorage &node_storage) : ClosedNamespace(NAMESPACE, SUFFIXES, node_storage) {}
+IRI RDF::operator+(const std::string &suffix) {
+    if (suffix.starts_with('_')) {
+        size_t id;
+        auto from_chars_result = std::from_chars(suffix.data() + 1, suffix.data() + suffix.size(), id);
+        if (from_chars_result.ec == std::errc()) {
```
liss-h: Potentially unwanted behaviour. `std::from_chars` can exit successfully without having consumed the whole string.

see: https://en.cppreference.com/w/cpp/utility/from_chars section Return value
example: https://godbolt.org/z/5vPf9crbf

Solution: the following must be true ```from_chars_result.ptr == suffix.data() + suffix.size()```

examples/namespaces_example.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include <iostream>
+
+int main() {
+    using namespace rdf4cpp::rdf;
+
+    Namespace ex("http://example.com/");
+
+    auto ex_A = ex + "A";
+    std::cout << ex_A << std::endl;
```
liss-h: Personal Opinion (yours might be different):

A general suggestion for examples: maybe replace the printing of example values by asserts.
The benefit this has is that users don't have to actually execute the code to see what the expected output is.
At least I personally would never actually bother to execute the example so it would be not helpful to me; but seeing what is does at a glance makes it useful to me.

Example:
```c++
Namespace ex{"http://example.com/"};

auto ex_A = ex + "A";
assert(std::string{ex_A} == "http://example.com/A"); // or something simmilar that shows what the expected output is
```

liss-h (commented):
Looks good in general.
Just found some minor things that could be improved performance- or usability-wise.

This is my first "bigger" code review so I hope it's not too nitpicky

src/rdf4cpp/rdf/namespaces/RDF.cpp
liss-h: Solution: the following must be true: ```from_chars_result.ptr == suffix.data() + suffix.size()```

src/rdf4cpp/rdf/Namespace.hpp
bigerl: very good point. Together with using `emplace`, it works 👍🏻

src/rdf4cpp/rdf/Namespace.hpp
bigerl: Good point. What we loose, is that `auto x = rdfs.std::string{"debug: "} + x.name_space();` won't work anymore. That seems OK given that we do have a interface that is string implementation agnostic. 

name is a typo, should be `name_space` 👍🏻.

liss-h: I think the standard committee just forgot to provide an operator+ for string and string_view
because operator+= on string has a string_view overload; so you can emulate the missing operator+ by:

```c++
auto s = (std::string{"debug: "} += x.name_space());
```

which is kinda ugly but at least we don't loose that behaviour

src/rdf4cpp/rdf/namespaces/RDF.cpp
bigerl: 🙀

examples/namespaces_example.cpp
bigerl: good point

bigerl:
Thank you for the review. All very good and valid points you got. Do not worry, you are not too nit-picky. :)

Issue #55, opened 2022-08-10
dynamic Literal constructors behave badly


liss-h:
Many of the Literal constructors behave badly as they allow the user to create a literal without the type being registered
which leads to triggered asserts in comparisons, ebv and numeric ops.

How to reproduce (used branch feature/literal-ops-comparisons for this example, but can be triggered in develop with num ops):

1. never mention `rdf4cpp::rdf::datatypes::xsd::String` anywhere in your code (because this would trigger the template instantiation and register the type, thus preventing this bug).

2. 
```c++
Literal lit1 = Literal{"test1"};
Literal lit2 = Literal{"test2"};
bool res = lit1 < lit2; // assertion triggered here
```

Constructors that have this behaviour:
- `Literal::Literal(std::string_view lexical_form, Node::NodeStorage &node_storage)`
- `Literal::Literal(std::string_view lexical_form, const IRI &datatype, Node::NodeStorage &node_storage)`
- `Literal::Literal(std::string_view lexical_form, std::string_view lang, Node::NodeStorage &node_storage)`
- `Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::NodeStorage &node_storage)`

liss-h:
Ideas:
- explicit template instantiations in cpp file
- static variables

liss-h:
Fixed by #59

PR #56, opened 2022-08-10
Feature: literal comparisons

### LINKED ISSUE #61: Add rdf:langString datatype to make "graphs_and_datasets" example work
Currently rdf:langString does not exist as a LiteralDatatype, but it is used in the mentioned example as though it is one.
This provokes a bug kind of related to #55 when #56 is merged as during comparisons the type is required to exist in the registry (which it does not) to fetch the corresponding function to perform the actual comparison.

To Fix:

- [x] "Just" define it as a LiteralDatatype
- [x]  Extend #56 to fallback to lexical form comparisons in the extended (order by) comparison operators (which are used for Node comparisons)

liss-h:
Initial draft for literal comparisons.
I marked this as a draft because there are some things to be discussed first:

- [x] correct filter compare behaviour?
  - [x] should filters also always return less-than instead of unordered for null <=> non-null value? -> no
- [x] correct order by compare behaviour?
- [x] return value of compare operators (should it be a Literal instead?)  => TriStateBool and ordering
- [x] see #55 as this creates problems here as well
- [x] comparison operators with TriStateBools
  - [x] find suitable modelling for TriStateBool
    - enum struct doesn't work; does not support `operator bool()` because must be memberfunction
- [x] fix #61

### DIFF
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -280,16 +280,16 @@ std::partial_ordering Literal::compare_impl(Literal const &other, std::strong_or
     auto const this_entry = DatatypeRegistry::get_entry(this_datatype);
     assert(this_entry != nullptr);
 
-    if (this_entry->compare_fptr == nullptr) {
-        return std::partial_ordering::unordered; // not comparable
-    }
-
     auto const other_datatype = other.datatype().identifier();
 
     std::strong_ordering const datatype_cmp_res = this_datatype <=> other_datatype;
 
     if (out_type_ordering != nullptr) {
         *out_type_ordering = datatype_cmp_res;
+    }
+
+    if (this_entry->compare_fptr == nullptr) {
+        return std::partial_ordering::unordered; // this not comparable
     }
 
     if (datatype_cmp_res == std::strong_ordering::equal) {
@@ -299,13 +299,13 @@ std::partial_ordering Literal::compare_impl(Literal const &other, std::strong_or
         assert(other_entry != nullptr);
 
         if (other_entry->compare_fptr == nullptr) {
-            return std::partial_ordering::unordered; // not comparable
+            return std::partial_ordering::unordered; // other not comparable
         }
 
         auto const equalizer = DatatypeRegistry::get_common_type_conversion(this_datatype, other_datatype);
 
         if (!equalizer.has_value()) {
-            return std::partial_ordering::unordered; // not convertible
+            return std::partial_ordering::unordered; // not convertible to common type
         }
 
         auto const equalized_compare_fptr = [&]() {
@@ -338,7 +338,7 @@ bool Literal::operator==(Literal const &other) const {
 }
 
 std::weak_ordering Literal::compare_with_extensions(Literal const &other) const {
-    auto type_cmp_res  = std::strong_ordering::equal;
+    std::strong_ordering type_cmp_res = std::strong_ordering::equivalent; // choose any here, will be overwritten anyway
     auto const cmp_res = this->compare_impl(other, &type_cmp_res);
 
     if (cmp_res == std::partial_ordering::equivalent) {
@@ -356,9 +356,11 @@ std::weak_ordering Literal::compare_with_extensions(Literal const &other) const 
             return std::weak_ordering::greater;
         }
 
-        // return equal on not comparable, since it's the only stable option
-        // => this makes this comparator weak
-        return std::weak_ordering::equivalent;
+        // return type comparison instead
+        // this is reached either:
+        //      - if one of the arguments is not comparable
+        //      - there is no viable conversion to a common type
+        return type_cmp_res;
     } else if (cmp_res == std::partial_ordering::less) {
         return std::weak_ordering::less;
     } else if (cmp_res == std::partial_ordering::greater) {
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -254,42 +254,54 @@ util::TriBool Literal::get_ebv_impl() const {
     return ebv(this->value()) ? util::TriBool::True : util::TriBool::False;
 }
 
-std::partial_ordering Literal::compare_impl(Literal const &other, std::strong_ordering *out_type_ordering) const {
+std::partial_ordering Literal::compare_impl(Literal const &other, std::strong_ordering *out_alternative_ordering) const {
     using datatypes::registry::DatatypeRegistry;
 
     if (this->handle_.null() || other.handle_.null()) {
-        return this->handle_ == other.handle_
-                       ? std::partial_ordering::equivalent
-                       : std::partial_ordering::unordered;
+        if (this->handle_ == other.handle_) {
+            return std::partial_ordering::equivalent;
+        } else {
+            if (out_alternative_ordering != nullptr) {
+                *out_alternative_ordering = this->handle_.null()
+                                             ? std::strong_ordering::less
+                                             : std::strong_ordering::greater;
+            }
+            return std::partial_ordering::unordered;
+        }
     }
 
     auto const this_datatype = this->datatype().identifier();
-    auto const this_entry = DatatypeRegistry::get_entry(this_datatype);
-    assert(this_entry != nullptr);
-
     auto const other_datatype = other.datatype().identifier();
 
     std::strong_ordering const datatype_cmp_res = this_datatype <=> other_datatype;
 
-    if (out_type_ordering != nullptr) {
-        *out_type_ordering = datatype_cmp_res;
-    }
-
-    if (this_entry->compare_fptr == nullptr) {
-        return std::partial_ordering::unordered; // this not comparable
-    }
+    auto const this_entry = DatatypeRegistry::get_entry(this_datatype);
 
     if (datatype_cmp_res == std::strong_ordering::equal) {
-        return this_entry->compare_fptr(this->value(), other.value());
-    } else {
-        auto const other_entry = DatatypeRegistry::get_entry(other_datatype);
-        assert(other_entry != nullptr);
-
-        if (other_entry->compare_fptr == nullptr) {
-            return std::partial_ordering::unordered; // other not comparable
+        if (out_alternative_ordering != nullptr) {
+            // types equal, fallback to lexical form ordering
+            *out_alternative_ordering = this->lexical_form() <=> other.lexical_form();
         }
 
-        auto const equalizer = DatatypeRegistry::get_common_type_conversion(this_datatype, other_datatype);
+        if (this_entry == nullptr || this_entry->compare_fptr == nullptr) {
+            return std::partial_ordering::unordered;
+        }
+
+        return this_entry->compare_fptr(this->value(), other.value());
+    } else {
+        if (out_alternative_ordering != nullptr) {
+            // types are different, the only useful alternative ordering is the type ordering
+            *out_alternative_ordering = datatype_cmp_res;
+        }
+
+        auto const other_entry = DatatypeRegistry::get_entry(other_datatype);
+
+        if (this_entry == nullptr || this_entry->compare_fptr == nullptr || other_entry == nullptr || other_entry->compare_fptr == nullptr) {
+            return std::partial_ordering::unordered;
+        }
+
+        auto const equalizer = DatatypeRegistry::get_common_type_conversion(this_entry->conversion_table,
+                                                                            other_entry->conversion_table);
 
         if (!equalizer.has_value()) {
             return std::partial_ordering::unordered; // not convertible to common type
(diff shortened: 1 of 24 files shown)


src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -300,6 +267,87 @@ Literal Literal::logical_binop_impl(std::array<std::array<TriStateBool, 3>, 3> c
...
+
+    auto const other_datatype = other.datatype().identifier();
+
+    auto const datatype_cmp_res = this_datatype <=> other_datatype;
+
+    if (out_type_ordering != nullptr) {
+        *out_type_ordering = datatype_cmp_res;
+    }
+
+    if (datatype_cmp_res == std::strong_ordering::equivalent) {
```
bigerl: equivalent or equal?

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -145,9 +155,32 @@ class Literal : public Node {
...
+
+    /**
+     * @return returns true if: *this <=> other == equivalent
+     * @note this only exists to hide operator== of base class Node
+     */
+    bool operator==(Literal const &other) const;
 
-    std::partial_ordering operator<=>(const Literal &other) const;
+    /**
+     * The comparison function with sparql operator extensions.
```
bigerl:
```suggestion
     * The comparison function with SPARQL operator extensions. 
     * Behavior fulfills requirements for SPARQL's ORDER BY
```

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -300,6 +267,87 @@ Literal Literal::logical_binop_impl(std::array<std::array<TriStateBool, 3>, 3> c
     return Literal::make<datatypes::xsd::Boolean>(res == TriStateBool::True, node_storage);
 }
 
+std::partial_ordering Literal::compare_impl(Literal const &other, std::partial_ordering *out_type_ordering) const {
```
```suggestion
std::partial_ordering Literal::compare_impl(Literal const &other, std::partial_ordering *out_type_ordering = nullptr) const {
```

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -300,6 +267,87 @@ Literal Literal::logical_binop_impl(std::array<std::array<TriStateBool, 3>, 3> c
...
+
+        assert(equalized_compare_fptr != nullptr);
+
+        return equalized_compare_fptr(equalizer->convert_lhs(this->value()),
+                                      equalizer->convert_rhs(other.value()));
+    }
+}
+
+std::partial_ordering Literal::compare(Literal const &other) const {
+    return this->compare_impl(other, nullptr);
```
```suggestion
    return this->compare_impl(other);
```

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -300,6 +267,87 @@ Literal Literal::logical_binop_impl(std::array<std::array<TriStateBool, 3>, 3> c
...
+
+bool Literal::operator==(Literal const &other) const {
+    return this->compare(other) == std::partial_ordering::equivalent;
+}
+
+std::partial_ordering Literal::compare_with_extensions(Literal const &other) const {
+    auto type_cmp_res = std::partial_ordering::unordered;
+    auto const cmp_res = this->compare_impl(other, &type_cmp_res);
+
+    if (cmp_res == std::partial_ordering::equivalent) {
```
```suggestion
    if (cmp_res == std::partial_ordering::equal) {
```

Computationally, not difference. Semantically, I would say two types are equal not equivalent.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -300,6 +267,87 @@ Literal Literal::logical_binop_impl(std::array<std::array<TriStateBool, 3>, 3> c
...
+
+std::partial_ordering Literal::operator<=>(Literal const &other) const {
+    return this->compare(other);
+}
+
+bool Literal::operator==(Literal const &other) const {
+    return this->compare(other) == std::partial_ordering::equivalent;
+}
+
+std::partial_ordering Literal::compare_with_extensions(Literal const &other) const {
```
bigerl: Isn't this ultimately a `std::strong_ordering`?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -300,6 +267,87 @@ Literal Literal::logical_binop_impl(std::array<std::array<TriStateBool, 3>, 3> c
...
+    auto const this_entry = DatatypeRegistry::get_entry(this_datatype);
+    assert(this_entry != nullptr);
+
+    if (this_entry->compare_fptr == nullptr) {
+        return std::partial_ordering::unordered; // not comparable
+    }
+
+    auto const other_datatype = other.datatype().identifier();
+
+    auto const datatype_cmp_res = this_datatype <=> other_datatype;
```
```suggestion
    std::strong_ordering const datatype_cmp_res = this_datatype <=> other_datatype;
```

We are comparing two std::string_views. Result is a [`std::strong_ordering`](https://en.cppreference.com/w/cpp/string/basic_string_view/operator_cmp#Notes). I would suggest to add the type explicitly here. It helps to understand why `compare_with_extensions` is `std::string_ordering`.

bigerl (changes requested):
Thx for the PR. In general, it looks very good. I add some small points for improvement. I will point out additional features required in a separate review.

bigerl:
We need a class `OrderByComparator` that applies `std::strong_ordering` and `operator==`. It should do so transparently for Node, Literal, IRI, BNonde and Variable (not really needed but for the sake of completeness; otherwise we would violate the strong ordering). 

Special case:
- `null()` Nodes are the smallest possible element.
- `null()` Nodes are ordered by `NodeBackendHandle::type()` (`NodeType`)
- For comparison it does not matter of what class they are. They are compared only by the two features above.

bigerl (changes requested):
We need a class `OrderByComparator` that applies `std::strong_ordering` and `operator==`. It should do so transparently for `Node`, `Literal`, `IRI`, `BNonde` and `Variable` (not really needed but for the sake of completeness; otherwise we would violate the strong ordering). 

Special case:
- `null()` Nodes are the smallest possible element.
- `null()` Nodes are ordered by `NodeBackendHandle::type()` (`NodeType`)
- For comparison it does not matter of what class they are. They are compared only by the two features above.

This Comparator must be used in a tests that must contain usage with a ordered and unordered container type.

src/rdf4cpp/rdf/Literal.cpp
liss-h: partial ordering does not have an equal variant

src/rdf4cpp/rdf/Literal.cpp
liss-h: I need to think about this, because
- if one of the parameters is not comparable we currently return unordered
- if the result type is not comparable we currently return unordered

Not sure what the correct mapping to strong/weak_ordering for each are yet

liss-h:
Tests failing because I changed the ordering function with extensions to return equivalent if types not comparable
Probably not optimal but need to think about the correct solution here

src/rdf4cpp/rdf/Literal.cpp
bigerl: Right, this was a left over from when I was considering if we should changes the whole Parameter to strong_ordering

bigerl:
For the TriStateBool:
@Clueliss how about a wrapper around a enum class?

liss-h:
> For the TriStateBool: @Clueliss how about a wrapper around a enum class?

Yes, that's probably what it's going to be. I had a look at the `std::*_ordering` types and they are also implemented this way.

liss-h:
There is one edge case where I'm not sure if this implementation is legal. I read the spec but it's not entirely clear to me if it's ok.
Currently this holds (for ORDER BY comparisons):

```c++
auto cmp_res = Literal::make<Double>(NAN).compare_with_extensions(Literal::make<Double>(NAN));
assert(cmp_res == std::weak_ordering::equivalent);
```

because: 
1. same datatype
2. registered
3. `NAN <=> NAN` returns std::partial_ordering::unordered
4. required to fallback to lexical form ordering to produce a `std::weak_ordering` which then returns `std::strong_ordering::equal`

Therefore: `NAN == NAN` in ORDER BY comparisons
In filters this would obviously be illegal, but the part about operator extensions is not clear enough to me that I can decide wether this is allowed in ORDER BY.

Btw: `Literal::make<Float>(NAN).compare_with_extensions(Literal::make<Double>(NAN))` does return `std::weak_ordering::greater` because `Float != Double` and therefore the fallback is type ordering. That behaviour is easily changeable to return `std::weak_ordering::equal` instead by falling back on lexical form ordering in case both are convertible+comparable but the comparison returns `std::partial_ordering::unordered`.

bigerl:
From what I can see, this case is not specified. I think, your approach is understandable and produces expectable results.
> `Literal::make<Float>(NAN).compare_with_extensions(Literal::make<Double>(NAN))` does return `std::weak_ordering::greater`

👍

src/rdf4cpp/rdf/util/TriBool.hpp
```diff
@@ -0,0 +1,96 @@
+#ifndef RDF4CPP_UTIL_TRIBOOL_HPP
+#define RDF4CPP_UTIL_TRIBOOL_HPP
+
+#include <array>
+
+namespace rdf4cpp::rdf::util {
+
+struct TriBool {
+private:
+    enum struct Value {
```
bigerl:
```suggestion
    enum struct Value : int8_t {
```

src/rdf4cpp/rdf/util/TriBool.hpp
```diff
@@ -0,0 +1,96 @@
+#ifndef RDF4CPP_UTIL_TRIBOOL_HPP
+#define RDF4CPP_UTIL_TRIBOOL_HPP
+
+#include <array>
+
+namespace rdf4cpp::rdf::util {
+
+struct TriBool {
```
bigerl, line 12: Implementation is good. Only some docstrings for the class, public members and public methods are missing.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -273,31 +240,186 @@ Literal Literal::numeric_unop_impl(OpSelect op_select, NodeStorage &node_storage
...
-    auto const other_ebv = other.get_ebv_impl();
+std::partial_ordering Literal::compare_impl(Literal const &other, std::strong_ordering *out_alternative_ordering) const {
+    using datatypes::registry::DatatypeRegistry;
+
+    if (this->handle_.null() || other.handle_.null()) {
+        if (this->handle_ == other.handle_) {
+            return std::partial_ordering::equivalent;
+        } else {
+            if (out_alternative_ordering != nullptr) {
+                *out_alternative_ordering = this->handle_.null()
```
bigerl, line 292: Add an explanatory comment

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -273,31 +240,186 @@ Literal Literal::numeric_unop_impl(OpSelect op_select, NodeStorage &node_storage
...
+std::weak_ordering Literal::compare_with_extensions(Literal const &other) const {
+    std::strong_ordering alternative_cmp_res = std::strong_ordering::equivalent; // choose any here, will be overwritten anyway
+    auto const cmp_res = this->compare_impl(other, &alternative_cmp_res);
+
+    if (cmp_res == std::partial_ordering::equivalent || cmp_res == std::partial_ordering::unordered) {
+        // return alternative ordering instead
+        return alternative_cmp_res;
+    } else if (cmp_res == std::partial_ordering::less) {
+        return std::weak_ordering::less;
+    } else if (cmp_res == std::partial_ordering::greater) {
```
bigerl: why not:

```suggestion
    } else { // cmp_res == std::partial_ordering::greater
```

?

bigerl (changes requested):

bigerl:
> should filters also always return less-than instead of unordered for null <=> non-null value?

no, in 17.2 [Filter Evaluation](https://www.w3.org/TR/sparql11-query/#evaluation) it says:

> Apart from [BOUND](https://www.w3.org/TR/sparql11-query/#func-bound), [COALESCE](https://www.w3.org/TR/sparql11-query/#func-coalesce), [NOT EXISTS](https://www.w3.org/TR/sparql11-query/#func-filter-exists) and [EXISTS](https://www.w3.org/TR/sparql11-query/#func-filter-exists), all functions and operators operate on RDF Terms and will produce a type error if any arguments are unbound.

which means we must to use the `<`, `<=`, etc. and get with this PR already the correct result as `TriBool::Error`.

It feel more consistent to me to map an error to unordered for <=>.

src/rdf4cpp/rdf/Literal.cpp
liss-h: good point

src/rdf4cpp/rdf/util/TriBool.hpp
liss-h: added comments to the conversions and partial_ordering_eq (also moved that out of the struct; as it didn't really belong there)

liss-h: Also if you have a better name for `partial_ordering_eq` that would be appreciated. I was just unable to find a fitting name 😅

bigerl: Nothing beautiful but some candidates to pick from:
- `pw_ordering_eq`
- `partial_weak_ordering_eq`

BTW: it should also be marked as `noexcept`. Implicit `noexcept` from `constexpr` might be lost when it is called at runtime AFAIK.

bigerl: same for the two constructors I think (`noexcept`)

bigerl (approved):

PR #59, opened 2022-08-15
Fix dynamic Literal ctors behaving badly

### LINKED ISSUE #55: dynamic Literal constructors behave badly

liss-h:
This PR fixes #55

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/xsd/Boolean.hpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/Boolean.hpp
@@ -59,11 +59,14 @@ inline bool capabilities::Logical<xsd_boolean>::effective_boolean_value(cpp_type
 
 
 namespace rdf4cpp::rdf::datatypes::xsd {
+
 /**
  * Implementation of xsd::boolean
  */
-using Boolean = registry::LiteralDatatypeImpl<registry::xsd_boolean,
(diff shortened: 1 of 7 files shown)


bigerl (approved):
Looks like it was easier than we anticipated in the end. :)

PR #63, opened 2022-08-26
Feature: datatype for rdf:langString

liss-h:
This PR adds a dummy LiteralDatatype for rdf:langString and integrates it into Literal.


bigerl (approved):
splendid. thx :)

PR #65, opened 2022-09-07
Feature: fixed datatype ids

liss-h:
implements fixed ids for datatypes in registry and node storage

note by @bigerl: API changed. This breaks existing custom implementations of `INodeStorageBackend` implementations.


src/rdf4cpp/rdf/datatypes/registry/DatatypeIRI.hpp
```diff
@@ -0,0 +1,126 @@
...
+    template<typename S>
+    auto operator()(S &&other) -> std::invoke_result_t<O, decltype(std::forward<S>(other))> {
+        return std::invoke(this->map_other, std::forward<S>(other));
+    }
+};
+
+template<typename F, typename O>
+DatatypeIRIVisitor(F, O) -> DatatypeIRIVisitor<F, O>;
+
+struct DatatypeIRIView {
```
liss-h: Probably need a better name for DatatypeIRIView and DatatypeIRI. Maybe you have ideas? (and classdoc missing)

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
```diff
@@ -181,17 +182,41 @@ struct Comparable {
...
  *
  * @tparam Capabilities all capabilities this instantiation should have
  */
-template<ConstexprString type_iri, template<ConstexprString> typename... Capabilities>
+template<util::ConstexprString type_iri, template<util::ConstexprString> typename... Capabilities>
 struct LiteralDatatypeImpl : capabilities::Default<type_iri>, Capabilities<type_iri>... {
     using typename capabilities::Default<type_iri>::cpp_type;
 
+    // todo: find better name
+    static constexpr DatatypeIRIView const datatype_iri = []() {
```
liss-h: LiteraldatatypeImpl::datatype_iri probably also needs a better name

src/rdf4cpp/rdf/datatypes/registry/util/StaticFlatMap.hpp
```diff
@@ -0,0 +1,95 @@
...
+        return this->storage.begin();
+    }
+
+    constexpr const_iterator end() const noexcept {
+        return this->storage.begin() + this->max_init;
+    }
+
+    constexpr const_iterator find(key_type const &search_key) const noexcept {
+        // todo: maybe do a linear find to reduce compiletime overhead?
+        auto res = std::ranges::lower_bound(this->begin(), this->end(), search_key, {},
```
liss-h: Should this be a linear search instead? Maps of this type are expected to be small anyways I think

src/rdf4cpp/rdf/storage/util/StaticVec.hpp
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * A vector-like type with a capacity
+ * that is bounded by a compiletime constant.
+ * The data is stored in the StaticVec itself and not allocated on the heap.
+ *
+ * @tparam T element_type
+ * @tparam MAX_SIZE upper bound for the capacity (the vec cannot grow past this)
+ */
+template<typename T, size_t MAX_SIZE>
+struct StaticVec {
```
liss-h: This type was needed in an early version, but no longer needed. Do we want to keep it?

liss-h (commented):
Need some input on some things :)

src/rdf4cpp/rdf/datatypes/registry/util/StaticFlatMap.hpp
bigerl: would be fine. My guess would be that the linear search is faster for very small maps because of vectorization and cache lines. Question would only be where the break even point is. My uninformed guess would be around 32 elements on modern server hardware.

src/rdf4cpp/rdf/storage/util/StaticVec.hpp
bigerl: I would remove it from this code base. The implementation looks nice, though. Please make a PR with it to this library: https://github.com/dice-group/dice-template-library

src/rdf4cpp/rdf/datatypes/registry/DatatypeIRI.hpp
bigerl: I tend to say that just DatatypeIRI, or Datatype, or DatatypeIdentifier would be fine. 

I stumbled upon sth else here: I did not get why in there is those two [DatatypeIRIView](https://github.com/rdf4cpp/rdf4cpp/blob/feature/fixed-datatype-ids/src/rdf4cpp/rdf/datatypes/registry/DatatypeIRI.hpp#L30) and [DatatypeIRI](https://github.com/rdf4cpp/rdf4cpp/blob/feature/fixed-datatype-ids/src/rdf4cpp/rdf/datatypes/registry/DatatypeIRI.hpp#L68).

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
bigerl: I would probably make that dependent on the name of the type

src/rdf4cpp/rdf/datatypes/registry/DatatypeIRI.hpp
liss-h: There is one owned version that can contain `std::string` (DatatypeIRI) and one borrowing version that can contain `std::string_view` (DatatypeIRIView)

src/rdf4cpp/rdf/storage/util/StaticVec.hpp
liss-h: :+1:

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -184,7 +185,7 @@ Literal Literal::numeric_binop_impl(OpSelect op_select, Literal const &other, No
         assert(to_string_fptr != nullptr);
 
         return Literal{to_string_fptr(op_res.result_value),
-                       IRI{op_res.result_type_iri, node_storage},
+                       IRI::from_datatype_iri(op_res.result_type_iri, node_storage), // todo translate
```
bigerl: What is this todo referring to?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -217,51 +218,44 @@ Literal Literal::numeric_binop_impl(OpSelect op_select, Literal const &other, No
...
                                                                                                   equalizer->convert_rhs(other.value()));
 
-        auto const to_string_fptr = [&]() {
-            if (op_res.result_type_iri == equalized_entry->datatype_iri) [[likely]] {
-                return equalized_entry->to_string_fptr;
-            } else [[unlikely]] {
-                return DatatypeRegistry::get_to_string(op_res.result_type_iri);
-            }
-        }();
+        auto const to_string_fptr = op_res.result_type_iri == equalized_iri
```
bigerl: I would stick with the lambda because it supports `[[likely]]`/`[[unlikely]] `

src/rdf4cpp/rdf/datatypes/registry/DatatypeConversion.hpp
```diff
@@ -8,13 +8,12 @@
 #include <type_traits>
 #include <utility>
 
+#include <rdf4cpp/rdf/datatypes/registry/util/Tuple.hpp>
```
bigerl, line 11: Tuple doesn't seem to be required here

src/rdf4cpp/rdf/datatypes/registry/DatatypeConversionTyping.hpp
```diff
@@ -4,9 +4,11 @@
 #include <concepts>
 #include <tuple>
 #include <type_traits>
+#include <variant>
```
bigerl: not required

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -206,12 +248,16 @@ class DatatypeRegistry {
      * @param datatype_iri datatype IRI string
      * @return if available a structure containing function pointers for all numeric ops
      */
-    inline static std::optional<NumericOps> get_numerical_ops(std::string_view const datatype_iri) {
-        auto const res = find_map_entry(datatype_iri, [](auto const &entry) {
-            return entry.numeric_ops;
+    inline static NumericOps const *get_numerical_ops(DatatypeIRIView const datatype_iri) {
+        auto const res = find_map_entry(datatype_iri, [](auto const &entry) -> NumericOps const * {
+            if (entry.numeric_ops.has_value()) {
+                return &*entry.numeric_ops;
```
bigerl:
```suggestion
                return &entry.numeric_ops.value();
```

for readability

src/rdf4cpp/rdf/storage/node/identifier/NodeID.hpp
```diff
@@ -5,32 +5,38 @@
...
 class __attribute__((__packed__)) NodeID {
 public:
-    static std::pair<NodeID, std::string> const default_graph_iri;
-    static std::pair<NodeID, std::string> const xsd_string_iri;
-    static std::pair<NodeID, std::string> const rdf_langstring_iri;
+    static std::pair<NodeID, std::string_view> const default_graph_iri;
+    static std::pair<NodeID, std::string_view> const xsd_string_iri;
+    static std::pair<NodeID, std::string_view> const rdf_langstring_iri;
+
+    //static util::StaticVec<std::pair<NodeID, std::string_view>, 64> const predefined_iris;
```
bigerl:
```suggestion
```

bigerl (changes requested):
Beautifully done. Again, only some minor changes I would suggest. 

Almost forgot to mention: 

- [x] documentation :)

src/rdf4cpp/rdf/datatypes/registry/DatatypeIRI.hpp
bigerl: After the review, I would vote for `DatatypeIdentifier` / `DatatypeIdentifierView`

src/rdf4cpp/rdf/Literal.cpp
liss-h: Yes, I agree and wanted to do that the problem is just that lambdas cannot really capture structured bindings by reference or value it needs really ugly workarounds for that to work. The reason is something about structured bindings not being variables. This is how it would look

```
auto const to_string_fptr = [&op_res, equalized_id = std::ref(equalized_id), equalized_entry = equalized_entry]() {
    if (op_res.result_type_id == equalized_id.get()) [[likely]] {
        return equalized_entry->to_string_fptr;
    } else [[unlikely]] {
        return DatatypeRegistry::get_to_string(op_res.result_type_id);
    }
}();
```

src/rdf4cpp/rdf/Literal.cpp
liss-h: Just forgot to remove that comment

src/rdf4cpp/rdf/datatypes/registry/DatatypeConversion.hpp
liss-h: The file is using util::tuple_map, so that is required. Ok technically it is already included in DatatypeConversionTyping.hpp; is that what you meant?

src/rdf4cpp/rdf/Literal.cpp
bigerl: Ah, right. the structured binding - lambda thing is really annoying. 

I would still stick with the solution with `std::ref` you proposed. I think the likely/unlikely really makes sense here.

src/rdf4cpp/rdf/datatypes/registry/DatatypeConversion.hpp
bigerl: I must have overlooked tuple_map. Never mind and keep it.

src/rdf4cpp/rdf/datatypes/registry/util/StaticFlatMap.hpp
```diff
@@ -68,8 +68,7 @@ struct StaticFlatMap {
     }
 
     constexpr const_iterator find(key_type const &search_key) const noexcept {
-        // todo: maybe do a linear find to reduce compiletime overhead?
-        auto res = std::ranges::lower_bound(this->begin(), this->end(), search_key, {},
+        auto const res = std::ranges::lower_bound(this->begin(), this->end(), search_key, {},
```
bigerl: for small amounts of entries, linear search is faster. See https://quick-bench.com/q/fJYSIzHlH6oOjpMDSP_BHhHYuwM

The flipping point seems to be between 64 and 96 words. (see https://quick-bench.com/q/TEKyLMUO8eNtuatvw1hXJ2APmEs)  It's probably architecture and CPU dependent but I would expect that we are on the safe side with 64 for most if not all relevant targets. 

My suggestion would be 
```suggestion
        if (this->size() <= 64 * sizeof(size_t))
            return std::find_if(this->begin(), this->end(), [x](auto entry) noexcept {return entry.first == search_key;});
        else
        auto const res = std::ranges::lower_bound(this->begin(), this->end(), search_key, {},
```

bigerl: flipping point for this map is even lower at 16 entries. see: https://quick-bench.com/q/f1Rw_mdgVMJZxpwA2DviGuZ1mL8 (16) vs https://quick-bench.com/q/VR_Onb8H-0B4O8uWFTMWDPsoQYs (32) vs https://quick-bench.com/q/aEZsZ2Ox8IxK0i8Fxyk2arMnPXM (64)

bigerl (changes requested):
I am undecided if we should implement linear search for flat_maps with <= 16 elements. I would leave that to you. Otherwise everything is ready to merge.

src/rdf4cpp/rdf/datatypes/registry/util/StaticFlatMap.hpp
liss-h: As discussed I have now implemented it as a linear search

src/rdf4cpp/rdf/datatypes/registry/util/StaticFlatMap.hpp
```diff
@@ -0,0 +1,76 @@
...
+    using const_iterator = value_type const *;
+
+private:
+    using storage_t = std::array<value_type, capacity>;
+    storage_t storage;
+    size_t max_init;
+
+public:
+    constexpr StaticFlatMap(std::initializer_list<std::pair<Key, Value>> const inits) noexcept(std::is_nothrow_copy_constructible_v<value_type>)
+        : max_init{inits.size()} {
```
bigerl: may contain duplicates. just documenting it is fine.

src/rdf4cpp/rdf/datatypes/registry/util/StaticFlatMap.hpp
```diff
@@ -0,0 +1,76 @@
...
+        });
+    }
+
+    constexpr bool contains(key_type const &search_key) const noexcept {
+        return this->find(search_key) != this->end();
+    }
+
+    constexpr mapped_type const &operator[](key_type const &key) const noexcept {
+        auto it = this->find(key);
+        return it->second;
```
bigerl, line 71: out-of-bounds read if key not contained. `std::optional<std::reference_wrapper<mapped_type const &>>` or just use find?

bigerl (changes requested):
new changes look good. unfortunately, I overlooked these two issues in staticflatmap before.

src/rdf4cpp/rdf/datatypes/registry/util/StaticFlatMap.hpp
liss-h: I have an idea that is kinda weird and kinda hacky; but it prevents it from compiling if the key is not found and since we are only using it at compiletime anyways it might be an idea(?) I would really like to use `if consteval` here but thats sadly c++23 so I think it's the best I can do to make it not compile

```c++
...
private:
    consteval const_iterator find_ub(key_type const &search_key) const noexcept {
        auto const it = this->find(search_key);

        if (it == this->end()) {
            return this->storage.end();
        }

        return it;
    }

public:
    consteval mapped_type const &operator[](key_type const &key) const noexcept {
        auto const it = this->find_ub(key);
        return it->second;
    }
...

// compiletime error; array index out of bounds
// because undefined behaviour is not allowed in constexpr contexts and has to be reported
inline constexpr LiteralType t = reserved_datatype_ids["hello"]; 
```

src/rdf4cpp/rdf/datatypes/registry/util/StaticFlatMap.hpp
liss-h: good point

src/rdf4cpp/rdf/datatypes/registry/util/StaticFlatMap.hpp
liss-h: and for some reason you cannot just static_assert that even if the function is consteval.. so yeah

bigerl: looks good 👍🏻 
maybe add to the class doc that it is only safe to use at compile time.

bigerl: This seems to break support for clang 12 and 13. Please double check that it cannot (easily) worked around.

liss-h: Sorry, didn't notice. Is fixed now

bigerl (approved):

PR #66, opened 2022-09-07
Feature: numeric ops errors

### LINKED ISSUE #64: Add error support to numeric literal ops
The (sparql) standard requires to raise "dynamic errors" under certain conditions in numeric operations which is currently not possible.
The problem is that the functions are marked noexcept and the return type does not allow for error value return.

There are two solutions:
1. remove the noexcept specification and just throw exceptions in the appropriate cases
    - trivial to implement
    - possibly bad performance
2. change the return type to be something like `std::expected<op_result_type, op_error>`
    - requires either: a. finding a library implementing something like it; or b. implementing our own until c++23 arrives

A decision needs to be made before merging some parts of the code for #45 (mainly related to one-side-bounded ap integer types, as there is no clear alternative to erroring on bound violation).

liss-h:
Fixes #64

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ endif ()
 # find packages
 find_package(serd 0 QUIET)
 find_package(fmt 9 COMPONENTS fmt-header-only QUIET)
-find_package(expected-lite)
+find_package(expected-lite 0.6 QUIET)
 
 # find system libraries
 if (NOT serd_FOUND)
@@ -41,13 +41,13 @@ endif ()
 
 if (NOT expected-lite_FOUND)
     FetchContent_Declare(
-            fetchexpected-lite
+            expected-lite
             GIT_REPOSITORY https://github.com/martinmoene/expected-lite
-            GIT_TAG v0.6.2
+            GIT_TAG 95b9cb015fa17baa749c2b396b335906e1596a9e # v0.6.2
     )
 
-    FetchContent_MakeAvailable(fetchexpected-lite)
-    message(DEBUG "No CMake package for expected-lite found. expected-lite was retrieved via FetchContent.")
+    FetchContent_MakeAvailable(expected-lite)
+    message("No CMake package for expected-lite found. expected-lite was retrieved via FetchContent.")
 endif ()
 
 ## Create the main rdf4cpp library target
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ if (NOT fmt_FOUND)
     )
 
     FetchContent_MakeAvailable(fetchfmt)
-    message(DEBUG "Non CMake package for fmt found. fmt was retrieved via FetchContent.")
+    message(DEBUG "No CMake package for fmt found. fmt was retrieved via FetchContent.")
 endif ()
 
 if (NOT expected-lite_FOUND)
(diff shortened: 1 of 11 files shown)


CMakeLists.txt
```diff
@@ -37,6 +39,17 @@ if (NOT fmt_FOUND)
     message(DEBUG "Non CMake package for fmt found. fmt was retrieved via FetchContent.")
 endif ()
 
+if (NOT expected-lite_FOUND)
```
liss-h, line 42: Need some help here cannot figure out the correct cmake magic to make it work 😅
That's also why the (fetchContent) tests fail

bigerl: see https://github.com/rdf4cpp/rdf4cpp/pull/66/commits/9ca09f45cc3ae8d5d55863fe00a5c3497a07ca3c

You normally add only your own headers to `target_include_directories`. CMake takes care that  headers of dependencies are visible within the library and within the interface if required. 

If assets from dependencies are exposed via the public headers of our library, the dependencies must be in the `PUBLIC` section of `target_link_libraries`.

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -332,6 +332,15 @@ struct SelectOpRes<std::false_type, Fallback> {
     using type = Fallback;
 };
 
+template<typename T>
+[[nodiscard]] nonstd::expected<std::any, NumericOpError> map_expected(nonstd::expected<T, NumericOpError> const &e) noexcept {
+    if (e.has_value()) {
+        return *e;
+    } else {
+        return e.error();
```
bigerl: Is this safe? Couldn't NumericOpError fall into `std::any`?

```suggestion
        return nonstd::make_unexpected(e.error());
```

bigerl (changes requested):
Nice work. The code is very easy to follow and well understandable. 

Besides the one comment I have two more things that need to be taken care of before we merge:
- [x] double check that there is no portable implementation for `__builtin_x_overflow`
- [x] add tests that raise the errors

liss-h:
> double check that there is no portable implementation for __builtin_x_overflow

I don't think there is, most compilers seem to have something with simmilar functionality but it's always slightly different.

Most people seem to just resort to a wall of macros that detects the compiler and then conditionally uses the correct function.
  - https://github.com/nemequ/portable-snippets/blob/master/safe-math/safe-math.h
  - https://github.com/dcleblanc/SafeInt

I guess we could use this kind of library for that but tbh it does not seem very robust to me. But I think this is probably a discussion for next meeting

bigerl:
> Most people seem to just resort to a wall of macros that detects the compiler and then conditionally uses the correct function.
> 
>     * https://github.com/nemequ/portable-snippets/blob/master/safe-math/safe-math.h
> 
>     * https://github.com/dcleblanc/SafeInt

I tend towards safe-math as it makes a more mature impression to me. 

Edit: Boost also seems to have something: https://www.boost.org/doc/libs/1_80_0/libs/safe_numerics/doc/html/index.html

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
liss-h: Yes, you are right. I have made that exact mistake multiple times now

liss-h:
Since we are going to stick with builtins for overflow for now, there is nothing more to be done.

bigerl (approved):

PR #70, opened 2022-09-19
Fix: integer ops will now always return xsd:integer

liss-h:
Introduces a new capability called `NumericStub`, which makes the type behave like it was numeric, without actually implementing any numeric ops. It instead relies on it's `numeric_impl_type` to perform the actual calculations.

Example: `xsd:int` is now a numeric-stub that delegates to `xsd:integer`.

This is implemented by simply calculating the required subtype offset to the impl-type (at compiletime) and then
always starting the common-type-conversion search at that offset (when searching for numeric type conversions).

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/LiteralDatatype.hpp
+++ b/src/rdf4cpp/rdf/datatypes/LiteralDatatype.hpp
@@ -29,6 +29,10 @@ enum struct NumericOpError {
     OverOrUnderFlow   // https://www.w3.org/TR/xpath-functions/#ERRFOAR0002
 };
 
+
+/**
+ * A type that is not explicitly a LiteralDatatype but fulfills the requirements for being impl-numeric (see NumericStubLiteralDatatype)
+ */
 template<typename LiteralDatatypeImpl>
 concept NumericImpl = requires(typename LiteralDatatypeImpl::cpp_type const &lhs, typename LiteralDatatypeImpl::cpp_type const &rhs) {
                           requires LiteralDatatypeOrUndefined<typename LiteralDatatypeImpl::add_result>;
@@ -59,6 +63,9 @@ concept NumericImpl = requires(typename LiteralDatatypeImpl::cpp_type const &lhs
 template<typename LiteralDatatypeImpl>
 concept NumericImplLiteralDatatype = LiteralDatatype<LiteralDatatypeImpl> && NumericImpl<LiteralDatatypeImpl>;
 
+/**
+ * A type that is not explicitly a LiteralDatatype but fulfills the requirements for being stub-numeric (see NumericStubLiteralDatatype)
+ */
 template<typename LiteralDatatypeImpl>
 concept NumericStub = requires {
                           requires NumericImplLiteralDatatype<typename LiteralDatatypeImpl::numeric_impl_type>;
@@ -72,6 +79,9 @@ concept NumericStub = requires {
 template<typename LiteralDatatypeImpl>
 concept NumericStubLiteralDatatype = LiteralDatatype<LiteralDatatypeImpl> && NumericStub<LiteralDatatypeImpl>;
 
+/**
+ * A LiteralDatatype that is either stub-numeric (see NumericStubLiteralDatatype) or impl-numeric (see NumericImplLiteralDatatype).
+ */
 template<typename LiteralDatatypeImpl>
 concept NumericLiteralDatatype = LiteralDatatype<LiteralDatatypeImpl> && (NumericImpl<LiteralDatatypeImpl> || NumericStub<LiteralDatatypeImpl>);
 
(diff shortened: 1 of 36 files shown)


src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -206,12 +236,12 @@ class DatatypeRegistry {
...
      */
-    inline static std::optional<NumericOps> get_numerical_ops(std::string_view const datatype_iri) {
+    inline static NumericOps const *get_numerical_ops(std::string_view const datatype_iri) {
         auto const res = find_map_entry(datatype_iri, [](auto const &entry) {
-            return entry.numeric_ops;
+            return &entry.numeric_ops;
         });
 
-        return res.has_value() ? *res : std::nullopt;
+        return res.has_value() && (**res).has_value() ? &***res : nullptr;
```
liss-h: I have a feeling that you are not going to like this, but the reason is that `res` has type `std::optional<std::optional<NumericOps> const *>` and it needs to become `NumericOps const *` to avoid a copy and still have a sensible return type.

Suggestions are of course appreciated :)

bigerl: A bit. It's a little hard to read. I would try to handle pointer and optional differently to make clear what happens here.

```cpp
if (res.has_value()){
   if (auto const res_value = res.value(); res_value->has_value()){
      return &res_value->value();
   }
}
return nullptr;
```

how about ↑ plus some doc strings why each of the `has_value()` calls could return false.

src/rdf4cpp/rdf/datatypes/LiteralDatatype.hpp
```diff
@@ -26,27 +26,50 @@ enum struct NumericOpError {
...
+/**
+ * A LiteralDatatype that is numeric but does not itself have an impl for any numeric op
+ * it instead names another NumericImplLiteralDatatype that the operations should be delegated to.
+ * This property is also referred to as being "stub-numeric".
+ */
+template<typename LiteralDatatypeImpl>
+concept NumericStubLiteralDatatype = LiteralDatatype<LiteralDatatypeImpl> && NumericStub<LiteralDatatypeImpl>;
+
+template<typename LiteralDatatypeImpl>
+concept NumericLiteralDatatype = LiteralDatatype<LiteralDatatypeImpl> && (NumericImpl<LiteralDatatypeImpl> || NumericStub<LiteralDatatypeImpl>);
```
bigerl, line 86: doc

src/rdf4cpp/rdf/datatypes/LiteralDatatype.hpp
```diff
@@ -26,27 +26,50 @@ enum struct NumericOpError {
...
 
 template<typename LiteralDatatypeImpl>
 concept LogicalLiteralDatatype = LiteralDatatype<LiteralDatatypeImpl> && requires(typename LiteralDatatypeImpl::cpp_type const &value) {
```
bigerl, line 89: doc

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -48,6 +49,35 @@ class DatatypeRegistry {
...
+         * Determines the number of subtype levels to immediately skip up in
+         * the hierarchy, if searching for a _numeric_ type conversion.
+         *
+         * @example if start_s_off = 1 then the search will _start_ at the immediate supertype
+         *      of the type this NumericOpsStub belongs to.
+         */
+        size_t start_s_off;
+    };
+
+    struct NumericOps : std::variant<NumericOpsStub, NumericOpsImpl> {
```
bigerl:
```suggestion
    struct NumericOps : public std::variant<NumericOpsStub, NumericOpsImpl> {
```

It's just for readability. But I had to look up if parent classes of structs are public by default.

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -206,12 +236,22 @@ class DatatypeRegistry {
      * @param datatype_iri datatype IRI string
      * @return if available a structure containing function pointers for all numeric ops
```
bigerl:
```suggestion
     * @return if available a pointer to a structure containing function pointers for all numeric ops; otherwise a null pointer
```

bigerl (changes requested):
Looks good and, albeit I tend to repeat myself, great work. Code looks even after this additional feature quite clean and architecture is clear. 
A little detail: I like that `get_numerical_ops` returns now a pointer. I think here it makes the code easier understandable avoiding that you have both a pointer that might be a null pointer wrapped into an optional that might not have a value. Both modeling the same thing. I can imagine that the  rusty part of your heart was bleeding a bit :)

bigerl (approved):

PR #72, opened 2022-09-23
merge develop into fix/literal ops 

bigerl:


bigerl:
@Clueliss One thing I came across while merging: should this https://github.com/rdf4cpp/rdf4cpp/blob/248a646576483e5fd70ac4819c91eb5bae63fc08/src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp#L85 be `DatatypeID`?

liss-h:
> @Clueliss One thing I came across while merging: should this
> 
> https://github.com/rdf4cpp/rdf4cpp/blob/248a646576483e5fd70ac4819c91eb5bae63fc08/src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp#L85
> be `DatatypeID`?

Yes, that is intended to be `std::string` I thought it would be kind of useless to store a datatypeId there since you can know the fixed id already by what index it is in. So I think having the actual IRI there is more useful.

liss-h:
Merge looks good. Thanks for doing that :).
I forgot that I needed to do that last week and was out of time when I noticed. Sorry

liss-h (approved):

PR #76, opened 2022-09-27
Feature: Serd-iterator

liss-h:
An `std::istream_iterator` like type that iterates over Quads in turtle files.
Uses serd (`serd_reader_read_chunk`) as a backend but tries to skip errors when they are encountered to try to continue parsing.

Marked as draft so you can look at it already but needs more testing and maybe a little cleanup.


src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
```diff
@@ -0,0 +1,162 @@
...
+#include <nonstd/expected.hpp>
+#include <rdf4cpp/rdf/Quad.hpp>
+
+namespace rdf4cpp::rdf::parser {
+
+/**
+ * An error that might occur while parsing a turtle file
+ */
+struct ParsingError {
+    enum struct Type {
```
bigerl:
```suggestion
    enum struct Type : uint8_t {
```

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
```diff
@@ -0,0 +1,162 @@
...
+        EofReached,
+        BadSyntax,
+        BadCurie,
+    };
+
+    Type error_type;
+    uint32_t line;
+    uint32_t col;
```
bigerl:
```suggestion
    uint64_t line;
    uint64_t col;
```

RDF files tend to be huge. 2^32 lines is not enough and 2^32 in a line might get hit even earlier, e.g. if a turtle file is stored in minimal form without \n.

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
```diff
@@ -0,0 +1,162 @@
...
+        return os;
+    }
+};
+
+enum struct ParsingFlag : uint8_t {
+    Strict = 1 << 0,
+    StopOnFirstError = 1 << 1,
+};
+
+struct ParsingFlags {
```
bigerl: move to own file

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+#include <nonstd/expected.hpp>
+#include <rdf4cpp/rdf/Quad.hpp>
+
+namespace rdf4cpp::rdf::parser {
+
+/**
+ * An error that might occur while parsing a turtle file
+ */
+struct ParsingError {
```
bigerl: move to own file

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Constructs the end-of-stream iterator
+     */
+    IStreamQuadIterator(std::default_sentinel_t) noexcept;
+
+    IStreamQuadIterator(IStreamQuadIterator const &) = delete;
+    IStreamQuadIterator(IStreamQuadIterator &&) = default;
+
+    explicit IStreamQuadIterator(std::istream &istream, ParsingFlags flags = ParsingFlags::none(), storage::node::NodeStorage &node_storage = storage::node::NodeStorage::default_instance());
```
bigerl: noexcept?

src/rdf4cpp/rdf/parser/SerdIteratorImpl.ipp
```diff
@@ -0,0 +1,336 @@
...
+
+#include <rdf4cpp/rdf/storage/util/robin-hood-hashing/robin_hood_hash.hpp>
+#include <rdf4cpp/rdf/Quad.hpp>
+
+namespace rdf4cpp::rdf::parser {
+
+struct IStreamAdaptor {
+    std::reference_wrapper<std::istream> istream;
+    uint32_t line_off = 0;
+    uint32_t col_off = 0;
```
bigerl:
```suggestion
    uint64_t line_off = 0;
    uint64_t col_off = 0;
```

see above

src/rdf4cpp/rdf/parser/SerdIteratorImpl.ipp
```diff
@@ -0,0 +1,336 @@
...
+            if (static_cast<char const *>(buf)[0] == '\n') {
+                self->line_off += 1;
+                self->col_off = 0;
+            }
+        }
+
+        return self->istream.get().gcount();
+    }
+
+    static int is_ok(void *voided_self) {
```
bigerl: noexcept?

src/rdf4cpp/rdf/parser/SerdIteratorImpl.ipp
```diff
@@ -0,0 +1,336 @@
...
+#include <serd-0/serd/serd.h>
+#endif
+
+
+#include <rdf4cpp/rdf/storage/util/robin-hood-hashing/robin_hood_hash.hpp>
+#include <rdf4cpp/rdf/Quad.hpp>
+
+namespace rdf4cpp::rdf::parser {
+
+struct IStreamAdaptor {
```
bigerl: move to owl file in subfolder. Namespace rdf4cpp::rdf::parser::util

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
```diff
@@ -0,0 +1,162 @@
...
+struct IStreamQuadIterator {
+    using value_type = nonstd::expected<Quad, ParsingError>;
+    using reference = value_type const &;
+    using pointer = value_type const *;
+    using difference_type = std::ptrdiff_t;
+    using iterator_category = std::input_iterator_tag;
+    using istream_type = std::istream;
+
+private:
+    struct Impl;
```
bigerl, line 45: Don't predeclare here. Will work after putting ParsingFlag and ParsingError into seperate files.

src/rdf4cpp/rdf/parser/SerdIteratorImpl.ipp
```diff
@@ -0,0 +1,336 @@
...
+        return self->istream.get().gcount();
+    }
+
+    static int is_ok(void *voided_self) {
+        auto *self = reinterpret_cast<IStreamAdaptor *>(voided_self);
+        return self->istream.get() ? 0 : 1;
+    }
+};
+
+struct IStreamQuadIterator::Impl {
```
bigerl: rename to e.g. `IStreamQuadIteratorSerdBackend`. Rename file accordingly.  Move it to namespace/folder `backend` (I am open for better name suggestions.

src/rdf4cpp/rdf/parser/SerdIteratorImpl.ipp
```diff
@@ -0,0 +1,336 @@
...
+private:
```
bigerl: <del>Separate source and header +</del> add noexcepts

bigerl: We should not export the headers of serd what we currently would need to do. Right now, we would need to move serd in the CMake file at target_link_libraries to the PUBLIC section and remove the private mark in the conanfile.

There basically two options here:
a. move all serd imports to cpp files
b. use  a private header for IStreamQuadIteratorSerdBackend ( see https://github.com/dice-group/tentris-private/blob/develop/libs/triple-store/private-include/dice/triple-store/SerdLoad.hpp for an example ). 

Please check what's feasible here. I assume that only option b. will work.

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
```diff
@@ -0,0 +1,162 @@
...
+        return (cpy |= flag);
+    }
+};
+
+constexpr ParsingFlags operator|(ParsingFlag const f1, ParsingFlag const f2) noexcept {
+    return ParsingFlags::none() | f1 | f2;
+}
+
+/**
+ * Simmilar to std::istream_iterator<>.
```
bigerl:
```suggestion
 * Similar to std::istream_iterator<>.
```

bigerl (changes requested):
Thank you for the PR. I like the general design with a flags type and a error type. That should help to make the primary API very stable. Function-wise I everything looks nice and neat. 

File-wise I would suggest a bit of restructuring. We also need to take care of visibility of headers and dependencies. 

Please find my detailed comments below.

After all other issues are addressed:
- [x] replace all other uses of serd in rdf4cpp with the stream iterator 🪚
- [x] documentation 📖 
- [x] code formatting 🤖

src/rdf4cpp/rdf/parser/SerdIteratorImpl.ipp
liss-h: It's already not exposing serd. Notice that the impl part is in an .ipp file (not sure if it's the correct extension) that is only included in the .cpp file. Basically the file is not supposed to be included anywhere public

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
liss-h: I'm predeclaring here so that the backend implementation can be entirely private (pimpl) and can be replaced at a later stage without having visible changes to the outside world

src/rdf4cpp/rdf/parser/SerdIteratorImpl.ipp
liss-h: But I can certainly move it into a "private-include" folder

bigerl: Ah, right. I overlooked that. Then I would reduce this comment to just add noexcepts ;)

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
liss-h: Good point just one problem is that serd itself uses 32bit ints 😅 (see [SerdError](https://drobilla.gitlab.io/serd/c/html/api/serd_event.html#c.SerdError)). I guess with the manual line tracking implemented currently it's not a problem but I had an idea yesterday that might make that unecessary (also faster). So I guess keep manual tracking for correctness?

src/rdf4cpp/rdf/parser/SerdIteratorImpl.ipp
bigerl: I have no experience with using ipp files. They should be omitted from the installed library because we filter for *.hpp and *.h. I would expect that they are visible if you pull in the library with FetchContent.

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
liss-h: No longer a problem since I also patched that in serd

src/rdf4cpp/rdf/parser/ParsingFlags.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+struct ParsingFlags {
+private:
+    using flag_u_type = std::underlying_type_t<ParsingFlag>;
+    flag_u_type flags;
+
+    constexpr ParsingFlags(uint8_t const flags) : flags{flags} {}
+
+public:
+    constexpr ParsingFlags(ParsingFlag const flag)
```
bigerl: noexcept

src/rdf4cpp/rdf/parser/ParsingFlags.hpp
```diff
@@ -0,0 +1,60 @@
...
+enum struct ParsingFlag : uint8_t {
+    Strict = 1 << 0,
+};
+
+struct ParsingFlags {
+private:
+    using flag_u_type = std::underlying_type_t<ParsingFlag>;
+    flag_u_type flags;
+
+    constexpr ParsingFlags(uint8_t const flags) : flags{flags} {}
```
bigerl: noexcept

src/rdf4cpp/rdf/parser/private/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -0,0 +1,247 @@
...
+                                              SerdNode const *obj_datatype,
+                                              SerdNode const *obj_lang) noexcept {
+
+    auto *self = reinterpret_cast<Impl *>(voided_self);
+
+    auto const graph_node = [&]() -> nonstd::expected<Node, SerdStatus> {
+        if (graph != nullptr) {
+            switch (graph->type) {
+                case SERD_CURIE:
+                    return self->get_prefixed_uri(pred);
```
bigerl: all `get_*` functions can throw if something malformed slipped through, e.g. "abc"^^xsd:int

That should be handled. I'd suggest to do it here in `on_stmt`.

bigerl (changes requested):
Great work. Only some exceptions need additional handling (and 2x noexcept) then we should be done here.

src/rdf4cpp/rdf/parser/private/IStreamQuadIteratorSerdImpl.cpp
liss-h: Ah right, totally forgot that

liss-h: But as far as I can tell only `get_literal` can throw right?
I mean the rest (IRI, BlankNode) is also not marked `noexcept` but as far as I can tell they don't validate anything

liss-h: Fixed it for literal and curies

bigerl: IRI and BlankNode don't validate yet but should at some point in the future.

src/rdf4cpp/rdf/parser/private/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -72,19 +73,57 @@ nonstd::expected<IRI, SerdStatus> IStreamQuadIterator::Impl::get_prefixed_uri(Se
...
+        } else {
+            return Literal{literal_value, this->node_storage};
+        }
+    } catch (std::runtime_error const &e) {
+        // NOTE: line, col not entirely accurate as this function is called after a triple was parsed
+        this->last_error = ParsingError{
+                .error_type = ParsingError::Type::BadLiteral,
+                .line = serd_reader_get_current_line(this->reader.get()),
+                .col = serd_reader_get_current_col(this->reader.get()),
+                .message = e.what()};
```
bigerl: The NOTE in the comment is rather important. Please append a similar statement to `.message`.

src/rdf4cpp/rdf/parser/private/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -72,19 +73,57 @@ nonstd::expected<IRI, SerdStatus> IStreamQuadIterator::Impl::get_prefixed_uri(Se
         oss << prefix_it->second << suffix;
         return IRI{oss.view(), this->node_storage};
     } else {
+        // NOTE: line, col not entirely accurate as this function is called after a triple was parsed
+        this->last_error = ParsingError{
+                .error_type = ParsingError::Type::BadCurie,
+                .line = serd_reader_get_current_line(this->reader.get()),
+                .col = serd_reader_get_current_col(this->reader.get()),
+                .message = "unknown prefix"};
```
bigerl: see below. The message should state that the curor is at the end of the triple

bigerl (changes requested):

src/rdf4cpp/rdf/parser/private/IStreamQuadIteratorSerdImpl.cpp
bigerl: Having said that: we can take care of IRIs and BlankNodes when they got validation in place.

liss-h: I implemented it, because I didn't see your latest comment and left TODO comments for IRI and BNodes

bigerl (approved):

PR #77, opened 2022-09-30
Per NodeStorage Namespaces

bigerl:
It is currently fairly expensive to construct a Namespace because the Hashtable must be populated on each usage. 
This renders it basically unusable for usage in functions where the signature is not extendable and only a single IRI is created via a Namespace. 

This is a first draft to address this problem by allowing namespace instances per NodeStorage. 

One open question for me is whether it works fine with persistent node storages. My idea is: At least with metall, they need to have their actual implementation separate from the INodeStorageBackend (because metall doesn't allow virtual inheritance). INodeStorageBackend handles the DependentAssetCleaners, so nothing will remain at next startup of a Namespace from the last executable run.


liss-h (commented):
So in general I think this is a good idea and looks good.
The one thing I am a little (much) worried about are the dependent asset cleaners, I am wondering multiple things now

- Why is this needed? I would assume that a node storage can only be destroyed if the refcount is 0
- If it is needed, then it would potentially break even trivial multithreaded applications because now multiple threads have mutable pointers to the same data (the instances map) without the users control
- Is this potentially an ownership design mistake in the node storage that needs deeper analysis?

Idk maybe you can help me here. But this feels very unsafe 😰 (not this PR but the node storage stuff). But maybe I'm worrying too much about this

bigerl:
First of all, thread safety is a good point that is not yet addressed in this design. `instance()` and `operator+` would need to be synchronized to make sure the design is thread safe. 

Secondly, regarding the general design. 
There are conflicting desired behaviors that need to be balanced:
- From a functional point of view of a node it would be ideal if it would use reference counts and live e.g. in a shared pointer. 
- From a interface point of view regarding storage point of view, a Node should be able to live in any type of storage but still expose the same interface. 
- From 

The idea behind the asset cleaner was to make 
```

**Comment by bigerl** [comment]
First of all, thread safety is a good point that is not yet addressed in this design. `instance()` and `operator+` would need to be synchronized to make sure the design is thread safe. 

Secondly, regarding the general design. 
There are conflicting desired behaviors that need to be balanced:
- From a functional point of view of a node it would be ideal if it would use reference counts and live e.g. in a shared pointer. 
- From a interface point of view regarding storage point of view, a Node should be able to live in any type of storage but still expose the same interface. 
- From a storage efficiency point of view it is important that a Node takes as little memory as possible. 
- For Node functions it would be ideal if all data resides directly in the Node object. 
- The library should as easy usable as possible without the need to think about implementation details. 

Result is:
- In the current implementation, the library is fairly easy to use if you do only "standard" things. You use the Standard NodeStorage and may not even realize it. 
- We do not have reference counts but you have the possibility to use a separate NodeStorages both "long living" nodes and nodes that are used only temporarily during a one-shot computation (e.g. parsing a SPARQL query). 
- The NodeStorage comes with responsibilities similar to iterators on container types: The user is responsible that after a NodeStorage is destroyed no Nodes or other assets bound to it remain or are -- even worse -- used. 

The idea behind the asset cleaner was to make this safe:
```cpp
#include <rdf4cpp/rdf.hpp>

#include <iostream>

int main() {
    using namespace rdf4cpp::rdf;

    [[maybe_unused]] IRI iri0{"http://first.iri/"};
    // default node storage is initialized now
    {
        auto volatile_node_storage = storage::node::NodeStorage::new_instance();
        assert(volatile_node_storage.id().value == 1);
        auto &rdf = namespaces::RDF::instance(volatile_node_storage);
        auto rdf_property = rdf + "Property";
        assert(rdf_property.identifier() == std::string{namespaces::RDF::NAMESPACE} + "Property");
    }
    // volatile_node_storage and namespaces::RDF instance for node storage with node_storage_id 1 are destroyed


    {
        auto volatile_node_storage2 = storage::node::NodeStorage::new_instance();
        // volatile_node_storage2 should be able to have the same node_storage_id without any problem
        assert(volatile_node_storage2.id().value == 1);
        auto &rdf = namespaces::RDF::instance(volatile_node_storage2);
        // the NodeID used by Bag
        const IRI &rdf_bag = rdf + "Bag";
        assert(rdf_bag.identifier() == std::string{namespaces::RDF::NAMESPACE} + "Bag");
    }
    // volatile_node_storage2 and namespaces::RDF instance for node storage with node_storage_id 1 are destroyed
}
```

bigerl:
code above is not working yet. Need to look into it tomorrow.

liss-h:
I think I might have an idea that achieves the same thing without the dependent asset cleaner and would (I think) fix all of these lifetime related problems. Instead of giving out pointers or nodestore ids directly we instead give out handles that contain an additional generation.

(I assume backends were supposed to be stored in the array in NodeStorage, but they currently aren't. So the following assumes that they were supposed to be stored there)

A rough sketch:
```c++
// internal type for storage
struct NodeStorageBackendSlot {
    INodeStorageBackend *backend;
    uint32_t generation;
};

// this is given to the outside
struct NodeStorageHandle {
    NodeStorageId index;
    uint32_t generation;
};

struct NodeStorage {
    ....stuff

    NodeStorageHandle id() { return handle_; }
private:
    static inline std::array<NodeStorageBackendSlot, 1024> node_context_instances;
    NodeStorageHandle handle_;

    // lookup of the backend for this NodeStorage
    INodeStorageBackend *backend() {
         auto &instance = node_context_instances[handle_.index];

         if (instance.generation != handle_.generation) [[unlikely]] {
              assert(false);
              std::abort() or exception or something else
         }

         return instance.backend;
    }
};
```

Then whenever a slot in the array gets deleted the generation is increased by 1.
When you try to look up the node storage you can just check if the generation match, and if they don't you know that this can't be the same node storage you originally meant.

```c++
struct NamespaceSlot {
     NamespaceClass namespace;
     uint32_t node_storage_generation;
};

struct NamespaceInstance {
     static NamespaceClass &instance(NodeStorage &node_storage) {
         static map<NodeStorageId, NamespaceSlot> instances;

         if (auto it = instances.find(node_storage.id().index); it != instances.end()) {
              if (it->second.node_storage_generation != node_storage.id().generation) {
                   // this can't be the namespace we wanted; old node storage doesn't exist anymore
                   // erase and replace
              }
              return it->second.namespace;
         } else {
              // insert
         }
     }
};
```

Hope it's at least kinda clear what I mean. I am not 100% sure that this would work with the current architecture, you can probably assess that better. But I personally would much prefer something like this over handing out mutable pointers (if it's possible to implement that). But I think this would require a little rework of the way refcounting and deleting of backends is handled

bigerl:
example from https://github.com/rdf4cpp/rdf4cpp/pull/77#issuecomment-1270274411 works now

bigerl:
in answer to: https://github.com/rdf4cpp/rdf4cpp/pull/77#issuecomment-1270447655

> (I assume backends were supposed to be stored in the array in NodeStorage, but they currently aren't. So the following assumes that they were supposed to be stored there)

What executable do you run where they aren't? For example in `tests_Namespace_life_cycle` they are definitely stored there.

I think your approach should work. I like it. I would introduce it in a separate PR though. Or more precisely, ask you to introduce it. :)

Please do a re-review on the things I fixed. Then I would merge this one.

src/rdf4cpp/rdf/storage/node/NodeStorage.cpp
```diff
@@ -67,17 +67,9 @@ NodeStorage::~NodeStorage() {
...
-            this->backend_->dec_use_count();
-        this->backend_ = other.backend_;
-        if (this->backend_ != nullptr)
-            this->backend_->inc_use_count();
-    } else {
-        if (other.backend_ != nullptr) {
-            other.backend_->dec_use_count();
-        }
-    }
+    if (this->backend_ != nullptr and this->backend_ != other.backend_)
```
liss-h: Is constructor so `this->backend_` will always be nullptr

src/rdf4cpp/rdf/storage/node/NodeStorage.cpp
```diff
@@ -96,18 +88,9 @@ NodeStorage &NodeStorage::operator=(const NodeStorage &other) noexcept {
...
-            if (this->backend_ != nullptr)
-                this->backend_->inc_use_count();
-        } else {
-            if (other.backend_ != nullptr) {
-                other.backend_->dec_use_count();
-            }
-        }
-
+        if (this->backend_ != nullptr and this->backend_ != other.backend_)
+            this->backend_->dec_use_count();
```
liss-h, line 89: Shouldn't this always decrease the refcount even if they have the same backend. It's still 2 instances pointing to the same backend and one of them will not point to it afterwards.

Example
```c++
auto n1 = storage::node::NodeStorage::new_instance();
auto n2 = n1; // copy to have the same backend, refcount == 2

n2 = std::move(n1); // move with node storages of same backend
// refcount still 2
```

liss-h (changes requested):
Looks fine to me, just some refcount related thingies in NodeStorage :)

PR #79, opened 2022-10-13
Fix: eliminate unecessary overhead of literal construction from template make

liss-h:
`Literal::make<>()` had a bug where it would call `Literal::make` which would then unecessarily canonicalize the literal a second time, which is expensive.


bigerl (approved):
looks good. thx. It's also good in general that the unchecked make functions for Literal are now at one place.

PR #82, opened 2022-10-17
Feature: Numeric Datatypes

liss-h:
Implements the remaining xsd and owl datatypes and makes the appropriate datatype arbitrary precision.
Additionally fixes the performance of numeric serialization by utilizing `<charconv>`

Some tests missing.


liss-h:
Tests failing because didn't implement fetchcontent stuff yet

liss-h:
Also: still need to integrate the test from the other older numeric types PR

liss-h:
Btw, compile times are really suffering. Maybe we should include explicit template specializations in a cpp file afterall. Although I'm not entirely certain that this would help

liss-h:
TODOs:
- [x] fix compile times
- [x] fix float,double to_string
- [x] make build without conan work
- [x] do cmake stuff properly

.github/workflows/tests_and_examples.yml
```diff
@@ -118,10 +118,10 @@ jobs:
           # build and test package
           conan remove -f $(conan inspect . --raw name)/$(conan inspect . --raw version)
           conan create . $(conan inspect . --raw name)/$(conan inspect . --raw version) -pr:b=default
-      - name: Install systemlib serd for FetchContent
+      - name: Install systemlib serd and boost for FetchContent
         run: |
           # FetchContent does not handle transitive dependencies and doesn't run conan.
-          sudo apt-get install libserd-dev
+          sudo apt-get install libserd-dev libboost-dev
```
liss-h: has nothing to do with this PR but this install of libserd-dev shouldnt be there

CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ endif ()
 # find packages
 find_package(fmt 9 COMPONENTS fmt-header-only QUIET)
 find_package(expected-lite 0.6 QUIET)
+find_package(Boost QUIET COMPONENTS)
```
liss-h: Please look if I did the boost dependency correctly, there are some things in the cmake logs that make it seem like its dragging in (maybe) unnecessary dependencies but idk how to fix that

bigerl: should be fine now. I've made the find_package REQUIRED. Otherwise, if we used cmake to get boost_multiprecision, we would be forced to export it what I don't want to do. 

Also, finding out which boost component includes what libraries is always a pain. Header seems. Side comment: header is  a bit odd. If you use the system package and cmake's own find_package definition, this is fine:
```
find_package(Boost REQUIRED COMPONENTS headers)
```

With conan it doesn't find headers here but linking against Boost::headers is fine.

src/rdf4cpp/rdf/datatypes/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+        PUBLIC
+        nonstd::expected-lite
+        Boost::headers
+        )
+
+set_target_properties(rdf4cpp_datatypes PROPERTIES
+        CXX_STANDARD 20
+        CXX_EXTENSIONS OFF
+        CXX_STANDARD_REQUIRED ON
+        POSITION_INDEPENDENT_CODE ON)
```
bigerl: PIC should be set cmd line argument `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` 

is it necessary here or can it be removed?

src/rdf4cpp/rdf/datatypes/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+        xsd/Decimal.cpp
+        xsd/Double.cpp
+        xsd/Float.cpp
+        xsd/String.cpp
+        )
+
+target_include_directories(
+        rdf4cpp_datatypes
+        PRIVATE
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>/../../.."
```
bigerl: If I understand the rdf4cpp_datatypes library right, it exposes no headers and just compiles the datatype classes. 
You could even make the target_link_libraries private then. 

What is the rational behind this separate cmake file? Does the private include speed up the build time? If not the effective make files resulting from cmake should be pretty much the same as if the datatype cpp would be directly part of rdf4cpp

src/rdf4cpp/rdf/datatypes/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+        )
+
+target_link_libraries(rdf4cpp_datatypes
+        PUBLIC
```
bigerl: can be made private. We don't have any public headers on this target so we also do not depend on any dependent headers to be visible.

CMakeLists.txt
```diff
@@ -85,6 +86,32 @@ add_library(rdf4cpp
...
+        src/rdf4cpp/rdf/datatypes/xsd/integers/non_negative/UnsignedShort.cpp
+        src/rdf4cpp/rdf/datatypes/xsd/integers/non_positive/NegativeInteger.cpp
+        src/rdf4cpp/rdf/datatypes/xsd/integers/non_positive/NegativeInteger.cpp
+        src/rdf4cpp/rdf/datatypes/xsd/integers/non_positive/NonPositiveInteger.cpp
+        src/rdf4cpp/rdf/datatypes/xsd/integers/non_positive/NonPositiveInteger.cpp
+        src/rdf4cpp/rdf/datatypes/xsd/integers/signed/Byte.cpp
+        src/rdf4cpp/rdf/datatypes/xsd/integers/signed/Int.cpp
+        src/rdf4cpp/rdf/datatypes/xsd/integers/signed/Integer.cpp
+        src/rdf4cpp/rdf/datatypes/xsd/integers/signed/Long.cpp
+        src/rdf4cpp/rdf/datatypes/xsd/integers/signed/Short.cpp
```
bigerl, line 115: these seem to be duplicates with the separate object library rdf4cpp_datatypes

CMakeLists.txt
```diff
@@ -126,24 +150,24 @@ add_library(rdf4cpp
...
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/serd/src>")
 
 target_link_libraries(rdf4cpp
         PUBLIC
         nonstd::expected-lite
+        Boost::headers
         PRIVATE
-        fmt::fmt-header-only
-        )
+        fmt::fmt-header-only)
```
bigerl: ah, the object library is not used at all 💡

CMakeLists.txt
```diff
@@ -126,24 +150,24 @@ add_library(rdf4cpp
         src/rdf4cpp/rdf/writer/NNodeWriter.cpp
         src/rdf4cpp/rdf/writer/NQuadsWriter.cpp
         src/rdf4cpp/rdf/writer/NTriplesWriter.cpp
-        ${serd_source_files}
-        )
+        ${serd_source_files})
+
```
bigerl, line 156: more a style comment: I prefer here trailing `)` in a new line for nicer git diffs.

bigerl (changes requested):
Review 1: CMake

Summed up, I think we can just remove the extra CMake file for datatypes, right?

CMakeLists.txt
liss-h: I just forgot to remove the CMakeLists.txt in datatypes, that was from a previous attempt. Oops 😅

src/rdf4cpp/rdf/datatypes/rdf/LangString.hpp
```diff
@@ -38,15 +32,18 @@ inline std::string capabilities::Default<rdf_lang_string>::to_string(cpp_type co
     __builtin_unreachable();
 }
 
+extern template struct LiteralDatatypeImpl<rdf_lang_string,
```
bigerl, line 35: good idea with extern

liss-h:
> Summed up, I think we can just remove the extra CMake file for datatypes, right?

Yes, I just forgot to delete it, sorry 😅

src/rdf4cpp/rdf/datatypes/CMakeLists.txt
liss-h: Oh, is that the only thing `PUBLIC` means in this context I thought it also had something to do with exporting symbols in the binary, good to know

src/rdf4cpp/rdf/datatypes/CMakeLists.txt
liss-h: I think it was, because it didn't compile otherwise but I didn't quite understand it. This was one of the reasons I scrapped that approach

src/rdf4cpp/rdf/datatypes/owl/Real.hpp
```diff
@@ -0,0 +1,43 @@
...
+};
+
+template<>
+capabilities::Default<owl_real>::cpp_type capabilities::Default<owl_real>::from_string(std::string_view s);
+
+template<>
+bool capabilities::Logical<owl_real>::effective_boolean_value(cpp_type const &value) noexcept;
+
+template<>
+std::partial_ordering capabilities::Comparable<owl_real>::compare(cpp_type const &lhs, cpp_type const &rhs);
```
bigerl, line 25: could we make `capabilities::Comparable<T>::compare` in general `noexcept`?

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
```diff
@@ -0,0 +1,54 @@
...
+template<>
+capabilities::Default<owl_rational>::cpp_type capabilities::Default<owl_rational>::from_string(std::string_view s) {
+    if (auto pos = s.find_last_of('-'); pos != std::string_view::npos && pos != 0) {
+        // owl:rational only allows - at beginning
+        throw std::runtime_error{"owl:rational parsing error: invalid sign position"};
+    }
+
+    if (s.find_first_not_of("0123456789/-") != std::string_view::npos) {
+        throw std::runtime_error{"owl:rational parsing error: invalid character in string"};
+    }
```
bigerl: why are those checks done here? does the underlying type allow more than what is permitted by owl:rational?

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
```diff
@@ -0,0 +1,54 @@
...
+    try {
+        return cpp_type{s};
+    } catch (std::runtime_error const &e) {
+        throw std::runtime_error{std::string{"owl:rational parsing error:"} + e.what()};
+    }
+}
+
+template<>
+std::string capabilities::Default<owl_rational>::to_string(cpp_type const &value) {
+    if (auto den = denominator(value); den < 0) {
```
bigerl, line 26: add small example that explains what happens here

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -0,0 +1,166 @@
...
+    // +1 for minus in mantissa
+    // +1 for integral part
+    // +1 for dot
+    // +1 for E
+    // +1 for minus in exponent
+    // at least 2 for exponent because the (c++) standard says so (https://en.cppreference.com/w/cpp/utility/to_chars)
+    static constexpr size_t buf_sz = 5 + std::numeric_limits<F>::max_digits10 + std::max(2ul, detail::log10ceil(std::numeric_limits<F>::max_exponent10));
+    std::array<char, buf_sz> buf;
+
+    std::to_chars_result res = [&]() {
```
bigerl, line 127: is noexcept automatically applied here because the enclosing function is noexcept? otherwise you could also make the lambda noexcept?

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -0,0 +1,166 @@
...
+            return std::to_chars(buf.data(), buf.data() + buf.size(), value, std::chars_format::scientific);
+        }
+    }();
+    assert(res.ec == std::errc{});
+
+    auto *e_ptr = std::find(buf.data(), res.ptr, 'e');
+    assert(e_ptr != res.ptr); // serializing in scientific notation, there must be an 'e'
+    *e_ptr = 'E'; // convert 'e' to 'E' as required by the SPARQL standard
+
+    auto const shift_amt = [&]() {
```
bigerl, line 145: see above

src/rdf4cpp/rdf/datatypes/xsd/integers/signed/Integer.hpp
```diff
@@ -0,0 +1,60 @@
...
+                                           capabilities::Comparable,
+                                           capabilities::Subtype,
+                                           capabilities::FixedId>;
+
+}  // namespace rdf4cpp::rdf::datatypes::registry
+
+
+namespace rdf4cpp::rdf::datatypes::xsd {
+
+struct Integer : registry::LiteralDatatypeImpl<registry::xsd_integer,
```
bigerl, line 52: I understand that during development it is handy to have the datatypes sorted and separated. For usage I would think it is easier to have them all in one place (i..e xsd/). What do you think?

src/rdf4cpp/rdf/datatypes/owl.hpp
```diff
@@ -0,0 +1,7 @@
+#ifndef RDF4CPP_OWL_HPP
+#define RDF4CPP_OWL_HPP
```
bigerl: add another header for rdf (for the sake of consistency)

src/rdf4cpp/rdf/datatypes/owl/Rational.hpp
```diff
@@ -0,0 +1,46 @@
...
+template<>
+struct DatatypeMapping<owl_rational> {
+    using cpp_datatype = boost::multiprecision::cpp_rational;
+};
+
+template<>
+capabilities::Default<owl_rational>::cpp_type capabilities::Default<owl_rational>::from_string(std::string_view s);
+
+template<>
+std::string capabilities::Default<owl_rational>::to_string(cpp_type const &value);
```
bigerl, line 22: could the `capabilities::Default<T>::to_string` be made noexcept?

bigerl (changes requested):
review 2/3: actual code

nicely done 👍

src/rdf4cpp/rdf/datatypes/owl/Rational.hpp
liss-h: yes I think so, added it to #84

src/rdf4cpp/rdf/datatypes/owl/Real.hpp
liss-h: Yes, I think so, added it to #84

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
liss-h: basically, the reason is that `boost::multiprecision::cpp_rational` allows values such as `1/-5`. Which is invalid for `owl:rational`.

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
liss-h: boost rational allows hex during parsing and i don't think there is any way to disable that

liss-h: and it also allows "1/-5" which is also invalid for `owl:rational`

liss-h: added comments

src/rdf4cpp/rdf/datatypes/xsd/integers/signed/Integer.hpp
liss-h: Do you mean moving all the integers into `xsd/` instead of their own directories?

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
liss-h: Sadly, no: https://godbolt.org/z/fz3MbxdMe
Thing is: those are immediatly invoked lambdas, not sure how to test how the compiler treats them

Added it to #84

src/rdf4cpp/rdf/datatypes/owl.hpp
liss-h: not sure if you meant that but found and fixed a include guard clash with namespaces/owl

src/rdf4cpp/rdf/datatypes/xsd/integers/signed/Integer.hpp
bigerl: yes. but I made up my mind and would keep it now as it is

src/rdf4cpp/rdf/datatypes/owl.hpp
bigerl: ah, it's there already. the rdf.hpp file

bigerl (approved):
tests are fine

PR #83, opened 2022-10-17
Feature: Literal casting

liss-h:
As the feature name implies, allows casting of literals to other types.


bigerl (changes requested):
- [x] Please add tests that cover the table of allowed casts here: https://www.w3.org/TR/sparql11-query/#FunctionMapping

liss-h:
Added the tests. Two things:
- not sure if I should actually implement the TODO with "?" in it
- the TODOs with "!" in them need to be fixed in the hierarchy as they require inverse conversion functions

liss-h:
- [x] figure out how to cast in opposite direction of hierarchy
- [x] "IRI -> str" possible with IRI::identifier but IRI would need cast function (=> use IRI::identifier)

Edit: Idea: inverse functions in hierarchy

tests/datatype/tests_NumOpResults.cpp
```diff
@@ -17,7 +17,7 @@ TEST_SUITE("numeric op results") {
             Integer::cpp_type const zero{0};
             Integer::cpp_type const one{1};
 
-            CHECK(Integer::div(one, zero) == nonstd::make_unexpected(NumericOpError::DivideByZero));
+            CHECK(Integer::div(one, zero) == nonstd::make_unexpected(DynamicError::DivideByZero));
         }
 
         // TODO: comment-in when decimal is properly implemented
```
bigerl, line 23: what pieces are missing here?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -165,6 +170,97 @@ Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::
...
+
+    auto const *this_e = DatatypeRegistry::get_entry(this_dtid);
+    if (this_e == nullptr) {
+        // this datatype not registered
+        return Literal{};
+    }
+
+    if (auto const common_conversion = DatatypeRegistry::get_common_type_conversion(this_e->conversion_table, target_e->conversion_table); common_conversion.has_value()) {
+        // general cast
+        // TODO: if performance is bad split into separate cases for up-, down- and cross-casting to avoid one set of std::any wrapping and unwrapping for the former 2
```
bigerl, line 249: I would cover this definitely in a separate PR. How much work is it? What's your estimate?

src/rdf4cpp/rdf/datatypes/registry/DatatypeConversion.hpp
```diff
@@ -91,6 +93,10 @@ consteval ConversionLayer auto make_conversion_layer_impl(LayerAcc const &table_
                 inline static typename target_type::cpp_type convert(typename source_type::cpp_type const &value) noexcept {
```
bigerl, line 92: Please resolve the TODO in the method documentation (LayerAcc template parameter)

bigerl (changes requested):
I tend to repeat my self but: great work. The new code is clean and given the functionality it implements, well readable. Tests cover all important cases I see. 

I have only some remarks and doc requests left.

tests/datatype/tests_NumOpResults.cpp
liss-h: basically #89 (the other open PR)

src/rdf4cpp/rdf/Literal.cpp
liss-h: almost no work, just looks uglier. Which is why I went with this version

src/rdf4cpp/rdf/datatypes/registry/DatatypeConversion.hpp
liss-h: just removed it as it has no relevant meaning except that it is the type of ```table_acc```

src/rdf4cpp/rdf/Literal.cpp
liss-h: should I change it or leave it as is?

bigerl: leave it for now.

PR #85, opened 2022-10-17
Fix: Effective boolean value handling

liss-h:
Adds a public function to extract the ebv in case you want to save node storage accesses for logical operators.


bigerl:
@nkaralis how should we design the interface to handle stuff like this:

```
SELECT ((<agd> && true) AS ?x) WHERE {}
```

```
SELECT ((<agd> + <asd>) AS ?x) WHERE {}
```

Both queries return en empty result.
Should the RDF nodes already take care of that? Or is it better handled in the expression evaluation?

bigerl (approved):
looks good. thx :)

PR #88, opened 2022-11-07
Fix: noexcept and nodiscard specs in literal, registry and misc functions

liss-h:


liss-h:
Please double check that I have not accidentally marked something as noexcept that isn't, and if you can think of anything else that can be noexcept feel free to mention it

bigerl (changes requested):
The placed `noexcept`s seem all fine. 

Further, I would suggest to make the following methods `noexcept`: 

Node (and overridden methods if applicable):
- to_node_storage
- operator std::string
- operator BlankNode, IRI, Literal, Variable

Literal
- lambdas passed in implementation of operators (as you pointed out)

bigerl (approved):
good idea with the requires clause for OpSelect

PR #91, opened 2022-11-21
Fix: decimal over- and underflow behaviour

liss-h:


bigerl (approved):
looks good. thx :)
note: I haven't checked for completeness. This review covers the extended behavior an its tests.

bigerl:
@Clueliss please merge develop or re-base. then we can proceed with merging into develop

PR #92, opened 2022-11-23
fix issue with parsing prefixed objects

nkaralis:


bigerl (approved):
tests are green, changes and new test case are clear

PR #96, opened 2022-11-28
Fix packaging and deploy problems

liss-h:


conanfile.py
```diff
@@ -35,6 +35,7 @@ def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
         self.options["fmt"].header_only = True
+        self.options["boost"].header_only = True
```
bigerl: Is the `boost::header_only` option absolutely necessary? For conan center packages it is required that only absolutely necessary options to other packages are set. 

Reason is: The option is transitively applied and must be explicitly override in consuming conan files. E.g., if some one wants to use `boost::log=True` they must also explicitly set `boost::header_only=False`.

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
```diff
@@ -312,19 +312,14 @@ struct LiteralDatatypeImpl : capabilities::Default<type_iri>, Capabilities<type_
                   "Mismatch between declared and actual fixed id mapping state. "
                   "Hint: maybe you forgot declare the fixed id or to add the FixedId capability. "
                   "Note: this would cause inconsistency between registry and node storage");
-protected:
-    static std::nullptr_t init() noexcept;
-    inline static std::nullptr_t const dummy = init();
+private:
```
bigerl, line 315: What's the rational behind this change? How is it different from what was there before?

bigerl (changes requested):
Looks good and would be great if that works. 

- [x] Please add builds for the static version (without the SHARED option in the CMake args) to CI so that we can be sure  both versions build.
- [x] In addition, the time for the execution of actions has increased somewhat. But could also be due to the runners, which do not always perform the same. Please check locally if the build has become slower.

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
liss-h: Well, it simly doesn't work without that change. Tbh I have no idea why it doesn't. But it seems the object is not enough for the linker to no drop the symbol for the static member. With a non-static member it doesn't drop
it for some reason.

conanfile.py
liss-h: Oh, I didn't know that. I'll just revert that then

liss-h:
Yes, the build times increased quite significantly:
Before: ~00:49 elapsed
After: ~01:27 elapsed

But if I remember the original time correctly, at least that's still much better than 3 minutes

bigerl (changes requested):
Please build in CI some executables, e.g. examples, additionally with static libraries. 
Errors occurred during linking or executing, if I remember correctly; not during building the static library.

bigerl (approved):

PR #102, opened 2022-11-29
Feature: Additional Parser functionality needed for tentris 

liss-h:
This PR adds two features to IStreamQuadIterator

1. The ability to provide custom prefixes during construction
```c++
std::istream is{...};
IStreamQuadIterator qit{is, ParsingFlags::none(), {{"xsd", "http://some_url.org"}, {"asd", "http://other.com"}}};
```

2. The ability to disable prefix parsing
```c++
std::istringstream iss{"@prefix xsd: <http://some_url.org#>"};
IStreamQuadIterator qit{is, ParsingFlag::NoParsePrefix, {{"xsd", "http://some_url.org"}, {"asd", "http://other.com"}}};
assert(qit->has_exception());
```

Note:
The interface to provide prefixes is not quite optimal since it suffers from two limitations:
1. The constructor cannot take template arguments since the type is pimpl. Therefore no ranges.
2. Something like `qit.add_prefix("a", "b")` wouldn't work since the iterator starts on the first triple, and therefore the prefix wouldnt work for the first triple. Fixing this would require breaking the interface and it would make it weird to use.

### DIFF
--- a/src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
+++ b/src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
@@ -15,10 +15,12 @@ IStreamQuadIterator::IStreamQuadIterator(std::default_sentinel_t) noexcept
     : impl{nullptr} {
 }
 
-IStreamQuadIterator::IStreamQuadIterator(std::istream &istream, ParsingFlags flags, storage::node::NodeStorage node_storage) noexcept
-    : impl{std::make_unique<Impl>(istream, flags.contains(ParsingFlag::Strict), std::move(node_storage))} {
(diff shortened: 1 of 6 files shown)


liss-h:
If you have a better idea for the prefix interface, feel free to suggest something :)

liss-h:
Oh and another thing, I noticed that it maybe makes sense to include `robin_hood` as a dependency
for the parser prefixes, since tentris and hypertrie also use `robin_hood` for high-perf temporary maps.

bigerl (approved):
looks perfectly fine. Do you still plan changes or is there another reason why this is still a draft?

I have no better idea for the interface. I think it is sufficient that way. We typically have only a hand full of prefixes. So it shouldn't be a bottle neck.

liss-h:
> Do you still plan changes or is there another reason why this is still a draft?

No, I just thought the interface for passing prefixes was kinda ugly and was hoping for a better solution

PR #103, opened 2022-12-01
Feature/complete revamp of github actions

bigerl:
This PR brings order to the CI actions. 
- Tests and examples are now separated from packaging and deployment tests with FetchContent and conan
- For the sake of efficiency coverage stays with tests and examples (but is executed only for gcc-12, shared)
- rdf4cpp is tested as a shared and as a static library
- Packaging and deployment tests are executed on one compiler only (for now clang-14)

Checkout the actions views of [Packaging and deployment tests](https://github.com/rdf4cpp/rdf4cpp/actions/runs/3593178760/jobs/6049800603) and [Run unit tests and examples](https://github.com/rdf4cpp/rdf4cpp/actions/runs/3593178749/jobs/6049803368). They are much cleaner now. It should be easier to track down where something went wrong and what is still working.

Open question: 
- [ ] @nkaralis  Do you have any idea why the coverall integration is no longer working? I changed the used compiler from gcc-11 to gcc-12. Might that be a problem? The error can be found e.g. here: https://github.com/rdf4cpp/rdf4cpp/actions/runs/3593178749/jobs/6049803368#step:10:14


nkaralis:
I do not know why the coverage stopped working. I will have a look at it as soon as possible and get back to you.

.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
```diff
@@ -0,0 +1,97 @@
...
+          ${{ fromJSON('["", "-DCMAKE_CXX_FLAGS=--coverage"]')[inputs.with-coverage] }}  
+          cmake --build build_dir
+        env:
+          CC: ${{ steps.install_cc.outputs.cc }}
+          CXX: ${{ steps.install_cc.outputs.cxx }}
+
+      - name: Run tests
+        working-directory: build_dir
+        run: |
+          ctest --parallel 2
```
liss-h, line 80: More out of interest: why specifically 2?

.github/workflows/test_packaging_and_deployment.yml
```diff
@@ -0,0 +1,12 @@
+name: Packaging and deployment tests
+on: [ 'pull_request' ]
+
+
+jobs:
+  packaging:
```
liss-h: It could maybe makes sense to include one GCC release here as well, just in case we come across some compiler specific behaviour that specifically breaks something in packaging. Tbh I am not sure what that would be but considering the fiddly nature of the datatypes atm it probably makes sense.

liss-h (changes requested):
Looks good to me and seems to run fine.
One potential thingy for the deploy tests :)

.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
bigerl: runners have two cores: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

.github/workflows/test_packaging_and_deployment.yml
bigerl: has now builds for gcc-12 and clang-14

bigerl:
I would ignore for now the problem that github actions seem to populate reusable actions via matrices not always correctly. See https://github.com/rdf4cpp/rdf4cpp/actions/runs/3620679876/attempts/1 where we have  `clang-14, ubuntu-22.04 / test-and-examples-static / static lib` twice.

liss-h (approved):

PR #106, opened 2022-12-12
Improvement: Replace std::tuple with mz::type_list in hierarchy calculations to decrease compiltime

liss-h:
Compiletime: ~1:50 -> 1:05 
Not as good as it was before we needed to fix the packaging stuff but still much better.


bigerl (approved):
Looks good. Cutting the compile time almost have is quite good. 

We could think about to make a PR for your additions to the TypeList. Especially a compile time type map seems like something others might also need. Also, a conan package would be handy. If we happen to need it in more projects we should think about how we could best contribute what we need.

PR #107, opened 2022-12-12
Feature: node inlining

liss-h:
TODOs:

- [x] make other types (other than xsd:int) inlineable where applicable
- [x] think about inlining IRIs, BNodes => IRIs: no, not enough space, BNodes: no
- [x] rdf:langString inlining? => no, not enough space to be useful
- [x] some comments
- [x] resolve todos in code

### DIFF
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -114,6 +114,13 @@ std::any Literal::value() const {
         return registry::LangStringRepr{
             .lexical_form = std::string{lit.lexical_form},
             .language_tag = std::string{lit.language_tag}};
+    }
+    else if (this->handle_.is_inlined()) {
+        auto const fi = registry::DatatypeRegistry::get_inlining_ops(datatype);
+        assert(fi != nullptr);
+
+        auto const inlined_value = this->handle_.node_id().literal_id().value;
+        return fi->from_inlined_fptr(inlined_value);
     } else if (auto const factory = registry::DatatypeRegistry::get_factory(datatype); factory != nullptr) {
         return factory(this->lexical_form());
     } else {
@@ -146,6 +153,19 @@ Literal Literal::make_lang_tagged_unchecked(std::string_view lexical_form, std::
                                              .language_tag = lang}),
                                      storage::node::identifier::RDFNodeType::Literal,
                                      node_storage.id()}};
+}
+
+Literal Literal::make_inlined_unchecked(uint64_t inlined_value, IRI const &datatype, Node::NodeStorage &node_storage) noexcept {
+    using namespace storage::node::identifier;
+
+    assert(inlined_value >> 42 == 0);
+    auto const type = static_cast<datatypes::registry::DatatypeIDView>(datatype);
+
+    return Literal{NodeBackendHandle{
+            NodeID{LiteralID{inlined_value}, type.is_fixed() ? type.get_fixed() : LiteralType::other()},
+            RDFNodeType::Literal,
+            node_storage.id(),
+            1}}; // TODO extract 1 into some constant
 }
 
 Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::NodeStorage &node_storage) {
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -23,12 +23,31 @@ Literal::Literal(std::string_view lexical_form, std::string_view lang, Node::Nod
     : Literal{make_lang_tagged_unchecked(lexical_form, lang, node_storage)} {}
 
 IRI Literal::datatype() const noexcept {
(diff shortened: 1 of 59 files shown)


liss-h:
- [x] Revalidate that we want the inlining tagging bit in the free tagging bits and not the literal type

comment by @bigerl: 3 tagging bits should still be enough and this way we keep more bits for payload

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -148,7 +187,19 @@ Literal Literal::make_lang_tagged_unchecked(std::string_view lexical_form, std::
...
 }
 
-Literal Literal::make(std::string_view lexical_form, const IRI &datatype, Node::NodeStorage &node_storage) {
+Literal Literal::make_inlined_unchecked(uint64_t inlined_value, storage::node::identifier::LiteralType fixed_id, Node::NodeStorage &node_storage) noexcept {
+    using namespace storage::node::identifier;
+
+    assert(inlined_value >> LiteralID::width == 0);
+    assert(fixed_id != LiteralType::other());
+
+    return Literal{NodeBackendHandle{NodeID{LiteralID{inlined_value}, fixed_id},
```
liss-h, line 196: This does not work if a custom node storage implementation does not automatically insert the fixed id types on creation. Should I fix this or is such a node storage implementation considered wrong?

bigerl: A correct NodeStorage implementation must adhere to the rule that certain values of certain types are not stored. This should be clearly stated in the documentation.

tests/util/tests_Inlining.cpp
```diff
@@ -0,0 +1,104 @@
...
+        CHECK(lit2.backend_handle().is_inlined());
+        CHECK(lit1 == lit2);
+
+        auto const extracted1 = lit1.template value<xsd::Long>();
+        auto const extracted2 = lit2.value();
+        CHECK(extracted1 == i);
+        CHECK(std::any_cast<xsd::Long::cpp_type>(extracted2) == i);
+    }
+
+    TEST_CASE("negative 64bit int not-inlining") {
```
bigerl: Small negative numbers should also be inlined. I would suggest 42 bit two's complement for storing it, or, alternatively storing the sign in the MSB. Both should still be much faster convertible to the target type than a hash map lookup.

This should be implemented for Integer as well.

src/rdf4cpp/rdf/datatypes/registry/util/Inlining.hpp
```diff
@@ -56,9 +66,83 @@ constexpr P pack(T value) noexcept {
...
+    if constexpr (sizeof(T) * 8 <= bits) {
+        return pack<P>(value);
+    } else if constexpr (std::unsigned_integral<T>) {
+        if (value >= (T{1} << bits)) [[unlikely]] {
+            return std::nullopt;
+        }
+
+        return pack<P>(value);
+    } else {
+        if (auto const boundary = T{1} << (bits - 1); value >= boundary || value < -boundary) [[unlikely]] {
```
bigerl:
```
auto constexpr boundary = T{1} << (bits - 1);
```

could probably be constexpr

tests/nodes/tests_Literal.cpp
```diff
@@ -14,7 +14,8 @@ TEST_CASE("Literal - Check for only lexical form") {
     CHECK(lit1.is_literal());
     CHECK(not lit1.is_variable());
     CHECK(not lit1.is_iri());
-    CHECK(lit1.lexical_form() == "Bugs Bunny");
+    lit1.lexical_form() == "Bugs Bunny";
```
bigerl: I think that is a left over from the PG tests. Please remove Bugs Bunny and any other similar names from the tests.

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -63,26 +64,41 @@ class Literal : public Node {
...
 
     /**
      * Creates a language-tagged Literal directly without any safety checks
      */
     [[nodiscard]] static Literal make_lang_tagged_unchecked(std::string_view lexical_form, std::string_view lang, NodeStorage &node_storage) noexcept;
 
+    /**
+     * Creates an inlined Literal without any safety checks
+     */
+    [[nodiscard]] static Literal make_inlined_typed_unchecked(uint64_t inlined_value, storage::node::identifier::LiteralType fixed_id, NodeStorage &node_storage) noexcept;
```
bigerl, line 99: document parameter inlined_value. It should explain what's expected here because a uint64_t is used

bigerl (changes requested):
Looks good. Very nice that you managed to inline all number types. Some little tiny things and we are ready to merge.

tests/nodes/tests_Literal.cpp
liss-h: Not sure I understand what you mean. I see the error and fixed that but idk what that has to do with the name

bigerl: I would like to have the mention of the Warner compic character "Bugs Bunny" removed and replaced by sth neutral

bigerl (approved):

PR #108, opened 2022-12-14
Fix Coverage

nkaralis:


nkaralis:
This PR fixes the issues related to:

- calculating coverage statistics
- sending the report to coveralls (https://coveralls.io/builds/55194184)

Coveralls has not commented on this PR (it should appear in the checks). I would like for us to give it a try in a fresh  PR and in case it does not work, I will investigate it further. Should I close this and open a new PR?

bigerl:
Would make sense. If it also doen't work, would codecov.io with this [github action](https://github.com/codecov/codecov-action) also be an option? What do you think.

Issue #99, opened 2022-11-28
Feature: 17.4.3 Functions on Strings


liss-h:
https://www.w3.org/TR/sparql11-query/#func-strings

- [x] 17.4.3.2 STRLEN
- [x] 17.4.3.3 SUBSTR
- [x] 17.4.3.4 UCASE
- [x] 17.4.3.5 LCASE
- [x] 17.4.3.6 STRSTARTS
- [x] 17.4.3.7 STRENDS
- [x] 17.4.3.8 CONTAINS
- [x] 17.4.3.9 STRBEFORE
- [x] 17.4.3.10 STRAFTER
- [x] 17.4.3.11 ENCODE_FOR_URI
- [x] 17.4.3.12 CONCAT
- [x] 17.4.3.13 langMatches
- [x] 17.4.3.14 REGEX
  - should preferably be implemented with [hyperscan](https://github.com/intel/hyperscan/tree/develop)
- [x] 17.4.3.15 REPLACE

PR #115, opened 2023-01-10
Feature: misc functions on literals

### LINKED ISSUE #97: Feature: 17.4.4 Functions on Numerics
https://www.w3.org/TR/sparql11-query/#func-numerics

- [x] 17.4.4.1 abs
- [x] 17.4.4.2 round
- [x] 17.4.4.3 ceil
- [x] 17.4.4.4 floor
- [x] 17.4.4.5 RAND

### LINKED ISSUE #134: Enhancement: provide an efficient way to compare the datatype of literals with another datatype
Currently to compare the datatype of literals one has to

```c++
Literal lit = ...unknown literal...;
bool dt_eq = lit.datatype() == IRI{datatypes::xsd::String::identifier};
```
Which potentially involves 2 unecessary backend calls. Instead a function should provided that makes use of DatatypeIds to compare.

```c++
struct Literal {
    template<LiteralDatatype T>
    bool is() const noexcept;

    bool is(IRI const &datatype) const noexcept;
};
```

### LINKED ISSUE #136: Perf Bug: Literal::compare_impl should check backend handles for early exit
if the handles are equal we can do an early exit with `equal` which is currently not done

liss-h:
This PR resolves #97 and #99 (at least partially).

Currently missing:
- [`ENCODE_FOR_URI`](https://www.w3.org/TR/sparql11-query/#func-encode)
- ~~[`langMatches`](https://www.w3.org/TR/sparql11-query/#func-langMatches)~~

Other notes:
- ~~the regex matching is currently implemented with `std::regex`~~

Also added some user defined literal/iri/bnode operators to make literal construction less verbose


src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -277,6 +313,255 @@ class Literal : public Node {
...
+     * @warning 1-based indexing
+     *
+     * @param start position of the first character
+     * @param len requested length of the substring
+     * @return the characters in this whose position P satisfy: P >= round(start) && P < round(start) + round(len) or the null literal if
+     *      - this is not string-like
+     *      - start is not xsd:double or derived from it
+     *      - len is not xsd:double or derived from it
+     */
+    [[nodiscard]] Literal substr(Literal const &start,
```
liss-h, line 752: What's your opinion on the two `substr` overloads having different behaviour?
- `substr(size_t, size_t, NodeStorage &)` behaves like `std::string_view::substr`
- `substr(Literal const &, Literal const &, NodeStorage &)` behaves as per [XPATH spec](https://www.w3.org/TR/xpath-functions/#func-substring) (you should read it, its weird 😅)

Note: my intention for having two overloads for each function was to have one version that is useful when working in "C++-land" and one version that is useful in "RDF-land" e.g. for query evaluation.

Argument for keeping is this way:
The XPATH spec is weird and the first overload is supposed to be useful in "C++-land" and a `substr(double, double)` is not what anyone would expect from a `substr` function in the context of C++. Especially not the 1-based indexing and handling of negative values.

Argument for making them consistent:
Well they are consistent then, but then we have to choose one behaviour and either move the XPATH weirdness into query evaluation or live with not having a proper C++y interface here.

bigerl: Let's keep both substr versions. I agree the definition is wild. Overhead seems little.

bigerl:
Question: Does substr in xsd count unicode characters or `char`? Unicode characters can span multiple `char`s.

liss-h:
> Question: Does substr in xsd count unicode characters or `char`? Unicode characters can span multiple `char`s.

"Definition: A character is an atomic unit of text as specified by ISO/IEC 10646:2000 [[ISO/IEC 10646]](https://www.w3.org/TR/REC-xml/#ISO10646). Legal characters are tab, carriage return, line feed, and the legal characters of Unicode and ISO/IEC 10646. [...]"

"any Unicode character, excluding the surrogate blocks, FFFE, and FFFF."

So I guess it counts `Unicode Scalar Values` (?)

Not sure but maybe this function works: [mblen](https://en.cppreference.com/w/c/string/multibyte/mblen)

liss-h:
~~Currently waiting on #122 merge to be able to implement `langMatches`.~~ 👍

liss-h:
Using utfcpp for unicode stuff now.
Also considered icu but it had problems with building on clang and the interface is very different to stdlib and also ineffiecient for our usecase.

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -282,6 +295,44 @@ class DatatypeRegistry {
...
+            // conversion does not convert to correct type
+            return nullptr;
+        }
+
+        return &conv;
+    }
+
+    inline static RuntimeConversionEntry const *get_supertype_conversion(DatatypeIDView const sub, DatatypeIDView const super) noexcept {
+        auto const sub_table = get_conversion_table(sub);
+        if (sub_table == nullptr) {
```
bigerl: Is this used anywhere?

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -282,6 +295,44 @@ class DatatypeRegistry {
...
                                                                                      size_t rhs_init_soff = 0) noexcept;
 
+    /**
+     * Tries to find a conversion to convert a subtype to the given supertype
+     *
+     * @param sub table of type to convert
+     * @param super table of conversion target
+     * @return the conversion if there is one, otherwise if sub is not a subtype of super nullptr
+     */
+    inline static RuntimeConversionEntry const *get_supertype_conversion(RuntimeConversionTable const &sub, RuntimeConversionTable const &super) noexcept {
```
bigerl: Is this used anywhere (except in get_supertype_conversion below)?

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -273,6 +278,14 @@ class DatatypeRegistry {
      */
     [[nodiscard]] static InliningOps const *get_inlining_ops(DatatypeIDView datatype_id) noexcept;
 
+    inline static RuntimeConversionTable const *get_conversion_table(DatatypeIDView const datatype_id) noexcept {
```
bigerl, line 281: Is this used anywhere (except in get_supertype_conversion below)?

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
```diff
@@ -2,6 +2,15 @@
 
 namespace rdf4cpp::rdf::datatypes::registry {
 
+static owl::Rational::cpp_type simplify(owl::Rational::cpp_type const &r) noexcept {
+    auto const divisor = gcd(numerator(r), denominator(r));
```
bigerl: This looks strange. Given r = 2/4 (=.5); divisor would be gcd(2,4) = 2- so 2/4 simplified would be 2/4/2 (= .25). 

Or am I overlooking sth?

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
```diff
@@ -48,6 +59,35 @@ std::partial_ordering capabilities::Comparable<owl_rational>::compare(cpp_type c
...
+
+template<>
+nonstd::expected<capabilities::Numeric<owl_rational>::floor_result_cpp_type, DynamicError> capabilities::Numeric<owl_rational>::floor(cpp_type const &operand) noexcept {
+    return static_cast<boost::multiprecision::cpp_int>(operand);
+}
+
+template<>
+nonstd::expected<capabilities::Numeric<owl_rational>::ceil_result_cpp_type, DynamicError> capabilities::Numeric<owl_rational>::ceil(cpp_type const &operand) noexcept {
+    auto const integral = static_cast<boost::multiprecision::cpp_int>(operand);
+    auto const rest = numerator(operand) % denominator(operand);
```
bigerl, line 90: I don't get why the rest thing is necessary. Shouldn't casting to integral simply result in an ceil already?

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
```diff
@@ -48,6 +59,35 @@ std::partial_ordering capabilities::Comparable<owl_rational>::compare(cpp_type c
...
 }
 
+template<>
+nonstd::expected<capabilities::Numeric<owl_rational>::abs_result_cpp_type, DynamicError> capabilities::Numeric<owl_rational>::abs(cpp_type const &operand) noexcept {
+    return boost::multiprecision::abs(operand);
+}
+
+template<>
+nonstd::expected<capabilities::Numeric<owl_rational>::round_result_cpp_type, DynamicError> capabilities::Numeric<owl_rational>::round(cpp_type const &operand) noexcept {
+    auto const integral_part = numerator(operand) / denominator(operand);
```
bigerl:
```c++
auto const integral_part = static_cast<boost::multiprecision::cpp_int>(operand);
```

as below?

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
```diff
@@ -48,6 +59,35 @@ std::partial_ordering capabilities::Comparable<owl_rational>::compare(cpp_type c
     }
 }
 
+template<>
+nonstd::expected<capabilities::Numeric<owl_rational>::abs_result_cpp_type, DynamicError> capabilities::Numeric<owl_rational>::abs(cpp_type const &operand) noexcept {
+    return boost::multiprecision::abs(operand);
+}
+
+template<>
+nonstd::expected<capabilities::Numeric<owl_rational>::round_result_cpp_type, DynamicError> capabilities::Numeric<owl_rational>::round(cpp_type const &operand) noexcept {
```
bigerl, line 71: Why do rational round and ceil have custom implementations instead of using boost functions like in real?

bigerl (changes requested):
Partial review: abs, round, floor, ceil 

Thx for the PR. Some  things are unclear to me. Please review the comments below.

src/rdf4cpp/rdf/regex/Regex.hpp
```diff
@@ -0,0 +1,94 @@
...
+    struct Impl;
+    std::unique_ptr<Impl> impl;
+
+public:
+    /**
+     * Tries to compile a regex.
+     * The syntax follows google's RE2 library.
+     *
+     * @param regex the regex to compile
+     * @param flags compilation flags
```
bigerl: Compilation Flag sounds like this decides if the regex is (pre-)compiled/optimized. I would suggest something like "flags to configure the regex mode".

src/rdf4cpp/rdf/regex/private/RegexReplacerImpl.cpp
```diff
@@ -0,0 +1,86 @@
...
+ * equivalent construct in the RE2 replacement syntax
+ *
+ * @param s rewrite string in libpcre syntax
+ * @return rewrite string in RE2 syntax
+ */
+static std::string translate_rewrite(std::string_view const s) {
+    std::string res{s};
+
+    auto pos = res.find_first_of('$');
+    while (pos != std::string_view::npos) {
```
bigerl: unsigned underflow is well defined in C++ and string_view's npos builds on it. So I would suggest to use it here. 

```c++
while (pos != std::string_view::npos) {
        if (pos == 0UL) {
            // is at start, cannot be escaped => replace
            replace(res, 0UL);
        } else {
            // find start of potential backslash sequence preceding the '$'
            auto const end = pos;
            auto before_start = end - 1UL;
            while (before_start != std::string_view::npos && res[before_start] == '\\') {
                --before_start;
            }

            if (auto const count = end - (before_start + 1UL); count % 2UL == 0UL) {
                // even number of '\' preceding '$' => not escaped => replace
                replace(res, pos);
            } else {
                // uneven number of '\' preceding '$' => escaped => remove escape char as RE2 does not require to escape '$'
                res.erase(pos - 1UL, 1UL);
                pos -= 1UL;
            }
        }

        pos = res.find_first_of('$', pos + 1UL);
    }
```

src/rdf4cpp/rdf/util/private/Utf8.hpp
```diff
@@ -0,0 +1,6 @@
+// Work around different directory structure for utfcpp when using conan vs fetch content
```
bigerl: I guess it is time now to introduce an folder private at the root level that contains the (relevant parts of the) same folder structure. Things that are used only in cpp files should live there. This also applies to parts of the parser, I think.

src/rdf4cpp/rdf/BlankNode.cpp
```diff
@@ -23,4 +23,12 @@ std::ostream &operator<<(std::ostream &os, const BlankNode &node) {
     return os;
 }
 
+
+inline namespace literals {
```
bigerl: I like the idea of string literals but the name literals is here a bit ambiguous. How about shorthands?

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -314,6 +381,319 @@ class Literal : public Node {
...
+     * @param len requested length of the substring
+     * @return the characters in this whose position P satisfy: P >= start && P < start + len or the null literal if
+     *      - this is not string-like
+     */
+    [[nodiscard]] Literal substr(size_t start,
+                                 size_t len = std::string_view::npos,
+                                 NodeStorage &node_storage = NodeStorage::default_instance()) const noexcept;
+
+    /**
+     * TODO: should both substr version behave identically?
```
bigerl: No

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
liss-h: No, you are right. The idea was dividing both numerator and denominator by divisor. Don't ask me why I thought these were equivalent 😅

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
liss-h: because boost does not provide specializations of these functions for `rational`

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
liss-h: ~~Casting to integral is usually a floor not a ceil. Thats why the rest thingy is there~~ Oh ok I think you meant that and just wrote ceil by accident. Yes makes sense

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
liss-h: I'm not sure what I originally used that for but it appears unused now.

tests/nodes/tests_Literal.cpp
```diff
@@ -457,3 +459,237 @@ TEST_CASE("Literal - casting") {
...
+        SUBCASE("same thread") {
+            auto const l1 = Literal::generate_random_double();
+            auto const l2 = Literal::generate_random_double();
+
+            CHECK(l1 >= 0.0_xsd_double);
+            CHECK(l1 < 1.0_xsd_double);
+
+            CHECK(l2 >= 0.0_xsd_double);
+            CHECK(l2 < 1.0_xsd_double);
+            CHECK(l1 != l2);
```
bigerl: please add a short comment that this check is non-deterministic. Will most probably never fail but should be documented.

tests/nodes/tests_Literal.cpp
```diff
@@ -457,3 +459,237 @@ TEST_CASE("Literal - casting") {
...
+            auto const l1 = Literal::generate_random_double();
+            Literal l2;
+
+            std::thread t{[&]() {
+                l2 = Literal::generate_random_double();
+            }};
+
+            t.join();
+
+            CHECK(l1 != l2);
```
bigerl: same as above

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
liss-h: But I guess its a maybe useful accessor if you only want the table. If every other capability gets it's own accessor I guess this makes sense to have (?)

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
liss-h: Not sure what the original usage for that was, but appears unused.

tests/nodes/tests_Literal.cpp
```diff
@@ -457,3 +459,237 @@ TEST_CASE("Literal - casting") {
...
+        CHECK("abracadabra"_xsd_string .regex_replace("bra"_xsd_string, "*"_xsd_string) == "a*cada*"_xsd_string);
+        CHECK("abracadabra"_xsd_string .regex_replace("a.*a"_xsd_string, "*"_xsd_string) == "*"_xsd_string);
+        CHECK("abracadabra"_xsd_string .regex_replace("a.*?a"_xsd_string, "*"_xsd_string) == "*c*bra"_xsd_string);
+        CHECK("abracadabra"_xsd_string .regex_replace("a"_xsd_string, ""_xsd_string) == "brcdbr"_xsd_string);
+        CHECK("abracadabra"_xsd_string .regex_replace("a(.)"_xsd_string, "a$1$1"_xsd_string) == "abbraccaddabbra"_xsd_string);
+        CHECK("AAAA"_xsd_string .regex_replace("A+"_xsd_string, "b"_xsd_string) == "b"_xsd_string);
+        CHECK("AAAA"_xsd_string .regex_replace("A+?"_xsd_string, "b"_xsd_string) == "bbbb"_xsd_string);
+        CHECK("darted"_xsd_string .regex_replace("^(.*?)d(.*)$"_xsd_string, "$1c$2"_xsd_string) == "carted"_xsd_string);
+
+        // 'The expression fn:replace("abracadabra", ".*?", "$1") raises an error, because the pattern matches the zero-length string'
```
bigerl, line 683: What exactly are the problems here? If it turns out to be complex, I would suggest to create an issue, fix the other things in ths PR and merge.

tests/util/tests_Regex.cpp
```diff
@@ -0,0 +1,76 @@
...
+            {
+                std::string s = "Hello 99 World";
+                r.make_replacer(R"(\\\\$0)").regex_replace(s);
+                CHECK(s == R"(Hello \\9\\9 World)");
+            }
+        }
+
+        SUBCASE("stray backslash") {
+            // TODO
+            // CHECK_THROWS(r.make_replacer(R"(Hello \ World)"));
```
bigerl, line 73: Also something for an issue?

bigerl (changes requested):
Review of the rest. 
This covers quite a lot: Regex, functions on strings, C++ literals (very nice idea BTW) and additions to CowString. Great work. All looks pretty well already. Some points that still need a little attention are listed below.

src/rdf4cpp/rdf/regex/Regex.hpp
liss-h: But the regex does get pre-compiled. See https://github.com/google/re2/wiki/CplusplusAPI#pre-compiled-regular-expressions

tests/util/tests_Regex.cpp
liss-h: This shouldn't work because it's an invalid escape (even in RE2 syntax as far as I can tell).
RE2 just accepts it and does some weird things here. I probably need to change the translation somehow to catch these cases manually.

liss-h: I.e. with regex r = `[0-9]` this passes:

```
std::string s = "Hello 1 World";
r.make_replacer(R"(Hello \ World)").regex_replace(s);
CHECK(s == "Hello Hello  World");
```

tests/nodes/tests_Literal.cpp
liss-h: This is basically behaviour coming from RE2 and I'm not sure how to fix it from the outside.

liss-h: opened an issue #132

tests/util/tests_Regex.cpp
bigerl: I would guess it is an escaped space. Like in Latex `\ `.

src/rdf4cpp/rdf/regex/Regex.hpp
bigerl: Yes but to me the phrase "compilation flag" sounds as if this flag would control if the regex gets pre-compiled.

The relevant part is that this flag decides if the resulting rdf4cpp Regex object will work in the modes DotAll, CaseInsensitive and/or Literal. The following is a guess: The flags would also be applicable if the backend regex implementation would not be used in pre-compiled mode.

tests/util/tests_Regex.cpp
liss-h: Do you think the behaviour is fine this way or is it important that we catch that?

liss-h: I just noticed something. Everything after a `\` which is not part of a reference to the match in the replacement string just gets deleted. You can see it in the example I gave above and also:

```
std::string s = "0";
r.make_replacer(R"(Hello \hgfhf World)").regex_replace(s);
CHECK(s == "Hello ");
```

It doesn't seem to matter what comes after the backslash if it isnt another backslash or a number the rest of the replacement just gets ignored. Why

bigerl: That is very odd. I would say we have for now some preliminary regex support. We state that the syntax follows google RE2. So users are warned. I would first complete features and then come back to this kind of fine-tuning. So, please document it in a issue and then mark this conversation as resolved.

src/rdf4cpp/rdf/util/private/Utf8.hpp
liss-h: Do you mean

```
- src
    - rdf4cpp
    - private
```

Or
```
- src
    - rdf4cpp
        - private
```

?

bigerl: I mean 
./src/rdf4cpp/ ...
./private/rdf4cpp/ ...

liss-h:
This PR now also addresses #136 #134

bigerl (approved):
Changes for https://github.com/rdf4cpp/rdf4cpp/issues/136 and https://github.com/rdf4cpp/rdf4cpp/issues/134 look fine

bigerl (changes requested):
Ready to merge after conflicts with develop are resolved

tests/util/tests_Regex.cpp
liss-h: #141

bigerl (approved):

PR #116, opened 2023-01-10
Fix: small misc fixes

liss-h:
small improvements

- move functions of registry into cpp file (hpp was getting quite large)
- optimize the regex for decimal (increased construction cost but faster search)
- some include fixes
- some more noexcept


bigerl (approved):
looks good

PR #118, opened 2023-01-11
Feature: hexBinary and base64Binary datatypes

liss-h:
This PR implements xsd:hexBinary and xsd:base64Binary.

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
@@ -26,7 +26,7 @@ static constexpr std::array<int8_t, 128> decode_lut{-1,  -1,  -1,  -1,  -1,  -1,
                                                     -1,  26,  27,  28,  29,  30,  31,  32,  33,  34,  35,  36,  37,  38,  39,  40,
                                                     41,  42,  43,  44,  45,  46,  47,  48,  49,  50,  51,  -1,  -1,  -1,  -1,  -1};
 
-static constexpr std::array<char, 4> base64_encode(std::byte const h1, std::byte const h2, std::byte const h3) noexcept {
+static std::array<char, 4> base64_encode(std::byte const h1, std::byte const h2, std::byte const h3) noexcept {
     uint32_t const combined = static_cast<uint32_t>(h1) << 16| static_cast<uint32_t>(h2) << 8 | static_cast<uint32_t>(h3);
 
     return {encode_lut[(combined >> 18) & 0b0011'1111],
@@ -35,7 +35,7 @@ static constexpr std::array<char, 4> base64_encode(std::byte const h1, std::byte
             encode_lut[combined & 0b0011'1111]};
 }
 
-static constexpr uint8_t base64_decode_single(char const c) {
+static uint8_t base64_decode_single(char const c) {
     auto const decoded = decode_lut[static_cast<size_t>(c)];
     if (decoded < 0) {
         throw std::runtime_error{"xsd:base64Binary parsing error: invalid digit"};
@@ -44,7 +44,7 @@ static constexpr uint8_t base64_decode_single(char const c) {
     return static_cast<uint8_t>(decoded);
 }
 
-static constexpr std::array<std::byte, 3> base64_decode(char const c1, char const c2, char const c3, char const c4) {
+static std::array<std::byte, 3> base64_decode(char const c1, char const c2, char const c3, char const c4) {
     uint32_t const combined = (base64_decode_single(c1) << 18) | (base64_decode_single(c2) << 12) | (base64_decode_single(c3) << 6) | base64_decode_single(c4);
 
     uint8_t const b1 = (combined >> 16) & 0b1111'1111;
--- a/src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
@@ -8,7 +8,7 @@ namespace rdf4cpp::rdf::datatypes::registry {
 
 namespace encode_decode_detail {
 
-// value -> ascii
+// integer value -> ascii base64 character
 static constexpr std::array<char, 64> encode_lut{'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K',
                                                  'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V',
                                                  'W', 'X', 'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f', 'g',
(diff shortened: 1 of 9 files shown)


src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
```diff
@@ -0,0 +1,184 @@
...
+// value -> ascii
+static constexpr std::array<char, 64> encode_lut{'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K',
+                                                 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V',
+                                                 'W', 'X', 'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f', 'g',
+                                                 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r',
+                                                 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', '0', '1', '2',
+                                                 '3', '4', '5', '6', '7', '8', '9', '+', '/'};
+
+// ascii -> value
+static constexpr std::array<int8_t, 128> decode_lut{-1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,
```
bigerl: Is the singed int8_t necessary? Woudn't an unsigned be sufficient while avoiding static_casts to unsigned? (127 instead of -1 then)

src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
```diff
@@ -0,0 +1,184 @@
...
+                                                    -1,  26,  27,  28,  29,  30,  31,  32,  33,  34,  35,  36,  37,  38,  39,  40,
+                                                    41,  42,  43,  44,  45,  46,  47,  48,  49,  50,  51,  -1,  -1,  -1,  -1,  -1};
+
+static std::array<char, 4> base64_encode(std::byte const h1, std::byte const h2, std::byte const h3) noexcept {
+    uint32_t const combined = static_cast<uint32_t>(h1) << 16| static_cast<uint32_t>(h2) << 8 | static_cast<uint32_t>(h3);
+
+    return {encode_lut[(combined >> 18) & 0b0011'1111],
+            encode_lut[(combined >> 12) & 0b0011'1111],
+            encode_lut[(combined >> 6) & 0b0011'1111],
+            encode_lut[combined & 0b0011'1111]};
```
bigerl: add a short code documentation that illustrates what happens here.

src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
```diff
@@ -0,0 +1,184 @@
...
+    return static_cast<uint8_t>(decoded);
+}
+
+static std::array<std::byte, 3> base64_decode(char const c1, char const c2, char const c3, char const c4) {
+    uint32_t const combined = (base64_decode_single(c1) << 18) | (base64_decode_single(c2) << 12) | (base64_decode_single(c3) << 6) | base64_decode_single(c4);
+
+    uint8_t const b1 = (combined >> 16) & 0b1111'1111;
+    uint8_t const b2 = (combined >> 8) & 0b1111'1111;
+    uint8_t const b3 = combined & 0b1111'1111;
+    return {static_cast<std::byte>(b1), static_cast<std::byte>(b2), static_cast<std::byte>(b3)};
```
bigerl: needs some documentation as well

src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
```diff
@@ -0,0 +1,184 @@
...
+}
+} // namespace encode_decode_detail
+
+
+Base64BinaryRepr Base64BinaryRepr::from_encoded(std::string_view const base64encoded) {
+    if (base64encoded.empty()) {
+        return Base64BinaryRepr{};
+    }
+
+    size_t const len = std::count_if(base64encoded.begin(), base64encoded.end(), [](auto const ch) { return ch != ' '; });
```
bigerl:
```suggestion
    size_t const len = std::count_if(base64encoded.begin(), base64encoded.end(), [](auto const ch) noexcept{ return ch != ' '; });
```

src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
```diff
@@ -0,0 +1,184 @@
...
+    return buf;
+}
+
+std::byte Base64BinaryRepr::hextet(size_t const n) const noexcept {
+    auto const triple = n / 4 * 3;
+    auto const off = (3 - (n % 4)) * 6;
+
+    uint32_t const bytes = static_cast<uint32_t>((*this)[triple]) << 16
+                           | (triple + 1 < this->size() ? static_cast<uint32_t>((*this)[triple + 1]) << 8 : 0)
+                           | (triple + 2 < this->size() ? static_cast<uint32_t>((*this)[triple + 2]) : 0);
```
bigerl:
```suggestion
    uint32_t const bytes = (triple < this->size() ? static_cast<uint32_t>((*this)[triple]) << 16 : 0)
                           | (triple + 1 < this->size() ? static_cast<uint32_t>((*this)[triple + 1]) << 8 : 0)
                           | (triple + 2 < this->size() ? static_cast<uint32_t>((*this)[triple + 2]) : 0);
```

Or should we do bound-checking here and throw an error?

src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <cstddef>
+#include <vector>
+
+namespace rdf4cpp::rdf::datatypes::registry {
+
+/**
+ * Represents a decoded base64 value as a sequence of bytes.
+ * Bytes are stored in the same order as the encoded hextets.
+ */
+struct Base64BinaryRepr : std::vector<std::byte> {
```
bigerl: make it a field. I know, I also did this in hypertrie somwhere. But I regret it. :)

e.g.: https://quuxplusone.github.io/blog/2018/12/11/dont-inherit-from-std-types/

src/rdf4cpp/rdf/datatypes/xsd/HexBinary.cpp
```diff
@@ -0,0 +1,108 @@
+#include <rdf4cpp/rdf/datatypes/xsd/HexBinary.hpp>
+
+namespace rdf4cpp::rdf::datatypes::registry {
+
+namespace encode_decode_detail {
+
+static constexpr std::array<char, 16> encode_lut{'0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
```
bigerl, line 8: short documentation what it does and how it works

src/rdf4cpp/rdf/datatypes/xsd/HexBinary.cpp
```diff
@@ -0,0 +1,108 @@
...
+                                                 'A', 'B', 'C', 'D', 'E', 'F'};
+
+static constexpr std::array<int8_t, 128> decode_lut{
```
bigerl: short documentation what it does and how it works

src/rdf4cpp/rdf/datatypes/xsd/HexBinary.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+} // namespace encode_decode_detail
+
+HexBinaryRepr HexBinaryRepr::from_encoded(std::string_view const hex_encoded) {
+    if (hex_encoded.empty()) {
+        return HexBinaryRepr{};
+    }
+
+    auto const start = hex_encoded.find_first_not_of('0');
```
bigerl, line 45: add some step-by-step documentation

src/rdf4cpp/rdf/datatypes/xsd/HexBinary.cpp
bigerl: same regarding int8_t vs. uint8_t as above (base64Binary)

src/rdf4cpp/rdf/datatypes/xsd/HexBinary.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace rdf4cpp::rdf::datatypes::registry {
+
+/**
+ * Represents a decoded xsd:hexBinary value as a byte sequence
+ * where each byte stores 2 half-octets.
+ *
+ * The bytes are stored in order least significant to most significant (i.e. index 0 is least significant).
+ */
+struct HexBinaryRepr : std::vector<std::byte> {
```
bigerl: make it a field (see above)

tests/datatype/tests_Base64Binary.cpp
```diff
@@ -0,0 +1,105 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+
+#include <doctest/doctest.h>
+#include <rdf4cpp/rdf.hpp>
+
+using namespace rdf4cpp::rdf::datatypes::xsd;
+
+// ascii -> value
+// note: modified to decode '=' to 0
```
bigerl: provide the reason why you did the modification here as well

tests/datatype/tests_Base64Binary.cpp
```diff
@@ -0,0 +1,105 @@
...
+                                                    41,  42,  43,  44,  45,  46,  47,  48,  49,  50,  51,  -1,  -1,  -1,  -1,  -1};
+
+static std::byte base64_decode_single(char const c) {
+    auto const decoded = decode_lut[static_cast<size_t>(c)];
+    REQUIRE(decoded >= 0);
+    return static_cast<std::byte>(decoded);
+}
+
+TEST_SUITE("base64Binary") {
+    TEST_CASE("canonical") {
```
bigerl: Is canonical the right name for the testcases? some seem quite non-canonical

tests/datatype/tests_HexBinary.cpp
```diff
@@ -0,0 +1,87 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+
+#include <doctest/doctest.h>
+#include <rdf4cpp/rdf.hpp>
+
+using namespace rdf4cpp::rdf::datatypes::xsd;
+using rdf4cpp::rdf::datatypes::registry::HexBinaryRepr;
+
+TEST_SUITE("xsd:hexBinary") {
+    TEST_CASE("canonical") {
```
bigerl: Is canonical the right name for the testcases? some seem quite non-canonical (see above)

bigerl (changes requested):
Thx for the PR. Looks fine in general and from what I can tell you got all the nasty bit shifting, masking, combining and slicing stuff right. I am impressed. 
Some more comments, especially in the code, would help future devs to not first fully understand all details of the encodings before touching your code.  

Requested changes concern mainly documentation and some smaller design details.

src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
liss-h: good point

tests/datatype/tests_Base64Binary.cpp
liss-h: canonicalization probbably fits better

src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp
liss-h: The reason I did it this way is that `n_hextets()` returns the number of hextets in the encoded string
meaning if there are padding chars '=' they also count towards it.

Example: `Hello World` <-> `SGVsbG8gV29ybGQ=`

In this value you can currently extract the last/padding hextet ('='). You get a value of zero for it.
But the type itself does not store the corresponding bytes that contain the padding hextets, as these bytes are by definition just zero bytes that were appended to the data to make it encodable.

Meaning: `n_hextets() == len(encoded value) == 16`  but `n_bytes() == len(decoded value) == 11`

Therefore to extract the padding hextet, the padding byte that does not exist in the decoded value needs to be
artificially inserted to not have a buffer overread.

Reason for why I chose this behaviour:
- the encoded padding hextets are required and have a purpose in the encoded value
- the padding bytes in the decoded value however do not have any meaning (apart from being padding) and were not part of the original data

Although I would agree that it is not the most efficient way to implement that (because of doing bounds checking twice). It was the least ugly way I could come up with.

bigerl: Thx for the explanation. Makes sense.

The one thing that still bothers me is what happens if triple >= this->size(). This is an index out of bound. That behavior is principally (obviously vector does the same) fine but needs to be documented.

liss-h: added warnings to index functions

Issue #121, opened 2023-01-12
Parsing of decimals


nkaralis:
Currently, Decimal::from_string throws an XSD parsing error when provided integers (e.g., ` Literal{"1", IRI("http://www.w3.org/2001/XMLSchema#decimal")}`

The spec's regular expression for decimals' allows integers. See: https://www.w3.org/TR/xmlschema11-2/#decimal

Issue #120, opened 2023-01-12
Casting operations


nkaralis:
There are some issues with casting operations

- Casting to boolean currently relies on EBV. This is not always desired. For example, when casting strings to boolean, the rules for casting from string need to be followed (see: https://www.w3.org/TR/xpath-functions/#casting-from-strings)
-  Currently, when casting to string, the code only pays attention to the lexical form. When casting numerical datatypes, and especially decimals, doubles and floats, to string the spec provides a set of rules that need to be followed (see: https://www.w3.org/TR/xpath-functions/#casting-to-string).

liss-h:
#122

Issue #117, opened 2023-01-11
Comparison of literals with language tags


nkaralis:
Currently, the expression `"DEF"@en-US == "DEF"@en-us` (consider them being literals) evaluates to false. 
[RFC 4646](https://www.ietf.org/rfc/rfc4646) states in section 2.1 (page 5):

_The tags and their subtags, including private use and extensions, are to be treated as case insensitive: there exist conventions for the capitalization of some of the subtags, but these MUST NOT be taken to carry meaning._

I believe that the above expression should evaluate to true. 
This is also the case in following test case:

- https://github.com/w3c/rdf-tests/blob/main/sparql11/data-sparql11/functions/strlang03.rq (query)
- https://github.com/w3c/rdf-tests/blob/main/sparql11/data-sparql11/functions/strlang03-rdf11.srx (results)

Note that the query uses "en-US", whereas in the results "en-us" is used.

liss-h:
#122

PR #122, opened 2023-01-16
Fix: datatype behaviour fixes

liss-h:
This PR addresses a few things. The changes are all quite small so I didn't feel like opening a PR for each one was necessary. You can use the following as orientation to see which changes belong to what issue.

- #121
    - fixed regex
    - removed unecessary `try { } catch () { }` block (value already verified by regex so `boost` constructor cannot throw)
- #120
    - some reordering of special cases in `Literal::cast` to prioritize to/from `xsd:string` casts
    - a new function for datatypes called `capabilities::Default::display` (name as in Rust's `Display` trait)) mainly for numeric datatypes (but also others in the future)

- #117 
    - Added a `std::char_traits<>` like type for case insensitive comparisions on language tags (can be extended in the future if we discover additional requirements`

- a floating point canonicalization bug that would sometimes yield incorrect values
    - the original to_string for floating point values calculated `fraction = fmod(value, &exp)` to determine the
precision for the output (fraction == 0 => 1 to include an extra .0, else => max). 
While this is correct when formatting to fixed notation.
It is definitely wrong for scientific notation because the mantissa gets normalized to (e.g. to 1.001 for 1001.0).
This resulted in a significant loss of precision in large integer-like values.

- making `LangStringRepr` a view to the backend, which saves 2 allocations when working with `rdf:langString` values
    - Values of `rdf:langString`s currently will always be equivalent to what is stored in the backend, so instead of allocating 2 additional `std::string`s every time a `rdf:langString` value (= `LangStringRepr`) needs to be materialized the value will now instead just point to the backend. In the future when we inline the language tags the `std::string_view` can instead point to an entry of an `inline constexpr StaticFlatMap<enum LangTag, std::string_view>`.

- added null checks to `Literal::as_lexical_form()` and `Literal::as_display()` to unify behaviour between these functions and other `Literal -> Literal` transformation functions. (Having this property on those functions makes them chainable without manually checking for nullness everytime)

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
+++ b/src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
@@ -69,8 +69,8 @@ struct Default {
      * @param value the value
      * @return <div>value</div>'s canonical string representation
      */
-    inline static std::string to_string(cpp_type const &value) noexcept {
-        // If not further specified, to_string is instantiated via operator<<. If operator<< is not defined for cpp_type instantiation will fail.
+    inline static std::string to_canonical_string(cpp_type const &value) noexcept {
+        // If not further specified, to_canonical_string is instantiated via operator<<. If operator<< is not defined for cpp_type instantiation will fail.
         std::stringstream str_s;
         str_s << value;
         return str_s.str();
@@ -80,8 +80,8 @@ struct Default {
      * Returns a string representation of a datatype that is intended to be shown in user friendly output.
      * E.g. when casting to xsd:string
      */
-    inline static std::string display(cpp_type const &value) noexcept {
-        return to_string(value);
+    inline static std::string to_simplified_string(cpp_type const &value) noexcept {
+        return to_canonical_string(value);
     }
 };
 
--- a/src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
+++ b/src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
@@ -75,6 +75,14 @@ struct Default {
         str_s << value;
         return str_s.str();
     }
+
+    /**
+     * Returns a string representation of a datatype that is intended to be shown in user friendly output.
+     * E.g. when casting to xsd:string
+     */
+    inline static std::string display(cpp_type const &value) noexcept {
+        return to_string(value);
+    }
 };
 
 /**
--- a/src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
+++ b/src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
@@ -95,6 +95,13 @@ constexpr size_t log10ceil(T const value) noexcept {
 }
 } // namespace detail
 
+template<size_t N>
+static void string_shift_left(std::array<char, N> &buf, size_t start, size_t end, size_t amt) {
+
+    std::shift_left(&buf[start], &buf[end], amt);
+
+}
+
 /**
  * Serializes a floating point number into its (SPARQL) canonical string representation
  * see https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#dt-float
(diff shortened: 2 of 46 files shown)


liss-h:
Found the bug, the code was missing declarations for some template specializations of `capabilities::Default::display`.
Apparently this triggers some kind of UB where the compiler can either use the specialized version or the non-specialized version.

bigerl (commented):
Partial review: https://github.com/rdf4cpp/rdf4cpp/issues/121

- regex good & unnecessary try-catch block removed

approved ✅

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -124,41 +131,46 @@ std::string to_chars(F const value) noexcept {
...
-
-        if (fraction == 0) {
-            // include one zero after dot
-            return std::to_chars(buf.data(), buf.data() + buf.size(), value, std::chars_format::scientific, 1);
-        } else {
-            // automatic precision guarantees round-trip-safety (https://en.cppreference.com/w/cpp/utility/to_chars)
-            return std::to_chars(buf.data(), buf.data() + buf.size(), value, std::chars_format::scientific);
-        }
-    }();
+    std::to_chars_result res = std::to_chars(buf.data(), buf.data() + buf.size(), value, std::chars_format::scientific);
```
bigerl, line 142: > [Unlike other formatting functions in C++ and C libraries, std::to_chars is locale-independent, non-allocating, and non-throwing.](https://en.cppreference.com/w/cpp/utility/to_chars#Notes)

Does locale=de-de_utf8 cause `,` being used as decimal separator?

bigerl, line 142: I could not find a detailed specification of the format `std::chars_format::scientific`. 

How do we know that it adheres to these two rules: 

> - If SV has an absolute value that is greater than or equal to 0.000001 (one millionth) and less than 1000000 (one million), then the value is converted to an xs:decimal and the resulting xs:decimal is converted to an xs:string according to the rules above, as though using an implementation of xs:decimal that imposes no limits on the totalDigits or fractionDigits facets. 
> - n other cases, the result consists of a mantissa, which has the lexical form of an xs:decimal, followed by the letter "E", followed by an exponent which has the lexical form of an xs:integer.

from Casting to [xs:string and xs:untypedAtomic](https://www.w3.org/TR/xpath-functions/#casting-to-string)

src/rdf4cpp/rdf/datatypes/xsd/Double.cpp
```diff
@@ -15,6 +15,25 @@ std::string capabilities::Default<xsd_double>::to_string(cpp_type const &value)
     return util::to_chars(value);
 }
 
+template<>
+std::string capabilities::Default<xsd_double>::display(cpp_type const &value) noexcept {
+    if (value == 0) {
+        return std::signbit(value) ? "-0" : "0";
+    }
+
+    if (auto const abs = std::abs(value); abs >= 0.000001 && abs < 1000000) {
```
bigerl: Ah, [Casting to xs:string and xs:untypedAtomic](https://www.w3.org/TR/xpath-functions/#casting-to-string) it's handled here. Please state in CharConvExt  that the not the full behavior is covered.

bigerl: Would it make sense to extract this into a separate function as it is used in Double and Float?

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
```diff
@@ -75,6 +75,14 @@ struct Default {
         str_s << value;
         return str_s.str();
     }
+
+    /**
+     * Returns a string representation of a datatype that is intended to be shown in user friendly output.
+     * E.g. when casting to xsd:string
+     */
+    inline static std::string display(cpp_type const &value) noexcept {
```
bigerl: I don't really get what the difference to to_string is. When are they actually different and what is each function used for? To me it looks right now a bit like there is two solutions for the same problem. 

I had a look at the rust function. If I got it right, it is used as a human-readable counterpart of fmt::Debug. I would say, our to_string is already human readable.

bigerl (changes requested):
Partial review: https://github.com/rdf4cpp/rdf4cpp/issues/122

- Please read all comments before starting to address them 
- Main point is that it is not yet clear to me why we need to two functions, to_string and display.

src/rdf4cpp/rdf/datatypes/registry/util/LangTag.hpp
```diff
@@ -0,0 +1,55 @@
+#ifndef RDF4CPP_REGISTRY_UTIL_LANGTAG_HPP
+#define RDF4CPP_REGISTRY_UTIL_LANGTAG_HPP
+
+#include <string_view>
+
+namespace rdf4cpp::rdf::datatypes::registry::util {
+
+struct LangTagCharTraits : std::char_traits<char> {
```
bigerl, line 8: typedef and using aliases being available in subclasses is non-standard behavior. You should redefine them, e.g. with 
```
using char_type = typename std::char_traits<char>::char_type;
```

bigerl (changes requested):
Partial review: #117

Only one little thing here (see comment). The solution with the std::char_traits is beautiful. 

I like the arg name 🪡

bigerl (commented):
Partial review: floating point canonicalization

- problem seems resolved

approved ✅

bigerl (commented):
Partial review: making LangStringRepr a view to the backend

- makes sense and looks good

approved ✅

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
liss-h: No it does not. See https://godbolt.org/z/dr5nYEKen

liss-h: It is specified in terms of `printf` so `std::chars_format::scientific` maps to `printf`'s `e` modifier.

https://en.cppreference.com/w/cpp/utility/to_chars At 4) "[..]  but the conversion specified for the as-if printf is [..] `e` if fmt is `std::chars_format::scientific`"

https://en.cppreference.com/w/c/io/fprintf in the table:  "converts floating-point number to the decimal exponent notation. For the `e` conversion style `[-]d.ddde±dd` is used. [..]"

bigerl: Have you checked that it prints more than 6 digits after the decimal point? The fprintf definition suggests that precision is limited to 6 digits. From what I could seen xsd there is no such limit and I am quite sure that we need more than 6 decimal digits to represent the precision of double

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
liss-h: The difference is effectively what they are used for.

- `to_string` is used when interacting with the backend as it always yields the canonical representation of a value (which is then stored in the backend, so the representations there are always canonical)
- `display` is used for casting numeric types to `xsd:string` as it does not always require the canonical version but a simplified one. You can basically also use it to output more user friendly representations, which is why I chose that name

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
liss-h: https://en.cppreference.com/w/cpp/utility/to_chars

3) "[..] parsing the representation using the corresponding `std::from_chars` function recovers value exactly." 
4) "same as (3), [but specified format]"

Would suggest that it has to, but I'll check quickly.

liss-h: Yes, behaves as expected. Added small test case

liss-h:
Added another small fix, edited the pr description to reflect that (last bullet)

src/rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp
bigerl: I would rename:
> to_string -> canonical_string

I am not sure about the name of the second function. How about 
> display -> xsd_string

liss-h: how about:
`to_canonical_string`
`to_simplified_string`
?

bigerl (approved):

Issue #119, opened 2023-01-12
Precision error when adding doubles and decimals


nkaralis:
Reference test case:

- https://github.com/w3c/rdf-tests/blob/main/sparql11/data-sparql11/aggregates/agg-sum-02.rq (query)
- https://github.com/w3c/rdf-tests/blob/main/sparql11/data-sparql11/aggregates/agg-numeric2.ttl (data)

Expected results for `:mixed2`: 4E-1; actual result: 4.0000000298023225E-1

liss-h:
fix merged, closing

PR #127, opened 2023-01-23
Fix: Precision errors when casting decimal -> double

liss-h:
As mentioned in #119 previously adding decimals and doubles resulted in a precision loss.

This way due to the cast of `xsd:double("0.2"^^xsd:decimal)` being roughly equivalent to `xsd:double(xsd:float("0.2"^^xsd:decimal))` which resulted from a limitation in the type hierarchy.

Previously only direct (neighbour) conversion functions could be specialized, i.e.  `xsd:decimal -> xsd:float`
or `xsd:float -> xsd:double` but not `xsd:decimal -> xsd:double`. As this is the order they appear in their promotion hierarchy. The (latter) indirect conversion function was instead automatically generated by composing the former two direct conversion functions, which then lead to the precision error.

This PR solves this problem by allowing types to explicitly specialize non-direct conversion functions further downstream, i.e. `xsd:decimal` can now specialize the promotion to `xsd:double` directly.

The mechanism to achieve this consists of:
- index templated promotion and supertype functions (`capabilities::Promotable::promote`, `capabilities::Subtyped::into_supertype`)
- a value that specifies the maximum index of specialization for the templated conversion functions (`DatatypePromotionSpecializationOverride::max_specialization_ix`, `DatatypeSupertypeSpecializationOverride::max_specialization_ix`)
- and some small changes in the hierarchy calculation

I am not sure why the diff is so horrible btw :(

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/LiteralDatatype.hpp
+++ b/src/rdf4cpp/rdf/datatypes/LiteralDatatype.hpp
@@ -101,21 +101,23 @@ concept ComparableLiteralDatatype = LiteralDatatype<LiteralDatatypeImpl> && requ
                                                                             };
 
 template<typename LiteralDatatypeImpl>
-concept PromotableLiteralDatatype = LiteralDatatype<LiteralDatatypeImpl> && requires(typename LiteralDatatypeImpl::cpp_type const &value, typename LiteralDatatypeImpl::promoted_cpp_type const &promoted_value) {
-                                                                                requires LiteralDatatype<typename LiteralDatatypeImpl::promoted>;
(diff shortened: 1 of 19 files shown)


liss-h:
~~Ok CI compilers too old again, will fix tomorrow.~~ 👍

bigerl (approved):
The changes look good and to know the types in different distances seems the only thing that helps here. On the downside, it increases the complexity of the type system and makes manual overrides necessary. To me, the price seems worth it.

Let's wait another month and see if more issues arise. If not, we should start writing down a full documentation that covers:
- general concept like promotion, substitution
- techniques like rank calculation
- what is implemented where and how things are glued together
- how to implement a new type
- how to implement new capabilities
- ...

PR #129, opened 2023-01-25
Fix: ambiguous Literal ctors

liss-h:
Fixes the constructor ambiguity in `Literal`.
For the sake of uniformeness also added make functions to `Node`, `BlankNode` and `IRI`.

### DIFF
--- a/src/rdf4cpp/rdf/BlankNode.cpp
+++ b/src/rdf4cpp/rdf/BlankNode.cpp
@@ -1,12 +1,20 @@
 #include "BlankNode.hpp"
 
 namespace rdf4cpp::rdf {
-BlankNode::BlankNode() noexcept : Node{} {}
+BlankNode::BlankNode() noexcept : Node{NodeBackendHandle{{}, storage::node::identifier::RDFNodeType::BNode, {}}} {}
 BlankNode::BlankNode(std::string_view identifier, Node::NodeStorage &node_storage)
     : Node(NodeBackendHandle{node_storage.find_or_make_id(storage::node::view::BNodeBackendView{.identifier = identifier}),
                              storage::node::identifier::RDFNodeType::BNode,
                              node_storage.id()}) {}
 BlankNode::BlankNode(Node::NodeBackendHandle handle) noexcept : Node(handle) {}
+
+BlankNode BlankNode::make_null() noexcept {
+    return BlankNode{};
+}
+
+BlankNode BlankNode::make(std::string_view identifier, Node::NodeStorage &node_storage) {
+    return BlankNode{identifier, node_storage};
+}
 
 std::string_view BlankNode::identifier() const noexcept { return handle_.bnode_backend().identifier; }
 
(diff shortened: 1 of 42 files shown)


src/rdf4cpp/rdf/BlankNode.cpp
```diff
@@ -1,13 +1,21 @@
 #include "BlankNode.hpp"
 
 namespace rdf4cpp::rdf {
-BlankNode::BlankNode() noexcept : Node{} {}
+BlankNode::BlankNode() noexcept : Node{NodeBackendHandle{{}, storage::node::identifier::RDFNodeType::BNode, {}}} {}
```
liss-h, line 4: Very long diff mostly not that interesting. Just wanted to show you this explicitly so you don't miss it.
I think this was buggy, at least when I compare it to `Literal` and `IRI`.

src/rdf4cpp/rdf/query/Variable.hpp
```diff
@@ -15,6 +15,9 @@ class Variable : public Node {
     explicit Variable(std::string_view name, bool anonymous = false,
```
liss-h, line 15: While fixing some other things I came across this curious interface in `query::Variable`.
Isn't an anonymous variable supposed to _not_ have a name? But maybe I got that wrong.
I replicated this behaviour in the additional functions but found it weird.

Also, to me this ctor also seems like it could be removed in favor of the functions I added because 
```
auto v = query::Variable::make_anonymous("a"); // expresses intent
query::Variable v{"a", true}; // not very expressive, need to look at declaration of ctor
``` 
sidenote: while writing these two lines I originally got the meaning of the boolean wrong (i.e. true being named), even though I just looked at it a few minutes ago.

bigerl: Anonymous variables are strange design decision in SPARQL. They look like blank nodes but are actually variables that cannot be projected. As blank nodes, they start with `_:` followed by a blank node identifier. 

So the "name" is actually a blank node identifier

src/rdf4cpp/rdf/BlankNode.cpp
bigerl: looks good. Yes it was a bug before.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -7,20 +7,96 @@
...
+}
+
+Literal Literal::make_typed(std::string_view lexical_form, IRI const &datatype, Node::NodeStorage &node_storage) {
+    using namespace datatypes::registry;
+
+    DatatypeIDView const datatype_identifier{datatype};
+
+    if (datatype_identifier == datatypes::rdf::LangString::datatype_id) {
+        // see: https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal
+        throw std::invalid_argument{"cannot construct rdf:langString without a language tag, please call one of the other constructors"};
```
bigerl: "... one of the other factory functions."

bigerl (changes requested):
Looks good. I am still a bit sad about the loss of Literal ctors. But it's obvious that it is much less ambiguous like that. 

There is just this one phrase in the an exception message to be fixed.

PR #130, opened 2023-01-25
Feature: Blank Node Management

### LINKED ISSUE #105: Feature: Blank Node Management
Blank nodes are no stable references in RDF, see https://www.w3.org/TR/rdf11-concepts/#section-blank-nodes

rdf4cpp is not trying to prevent blank node ID collisions. Scopes allow the the user to correctly use blank nodes within files and graph stores/datasets. Comparison between blank nodes from different scopes is technically possible within rdf4cpp but not allowed by RDF. 

- [ ] By default, we should not keep the identifier of a blank node loaded from file but assign new ones
- [ ] There should be a generator that you can ask for a fresh, so far unused identifier
  - [ ] replacable
  - Variants:
    - [ ] Random IDs (could be UUIDs)
    - [ ] increasing IDs
      - [ ] should be as short as allowed in turtle
      - [ ] optional user-defined pre- or suffix
- [ ] There should be a BlankNodeMapper that keeps track of what blank node identifiers from a file are represented by what blank node identifiers in a NodeStorage. (Scopes)
  - [ ] This needs exemplary use-cases to get an idea for a good interface.
  - [x] Check how other libraries handle that
  - [ ] Needs to be separated into front and backend. Backend without virtual methods (see tentris)
  - [ ] Should be able to support both RDF Dataset semantics merge and union (see https://www.w3.org/TR/rdf11-datasets/#default-graph-as-union-or-as-merge)
- [ ] There should be subscopes. Subscopes know all blank nodes from their parent scope. 
  - [ ] Subscopes are initialized with all blank nodes from their parent scope. If a blank node is added to a parent scope, it is forwarded to the subscope.
- [ ] A scope is not bound to a NodeStorage. Each method has a node_storage argument with a default value
- [ ] Add support for [Skolem IRIs](https://www.w3.org/TR/rdf11-concepts/#section-blank-nodes)

*Note:* SPARQL variables are not handled as blank nodes but as anonymous variables. Thus, name clashes between variables and blank nodes are no concern.

liss-h:
Intended as a solution to #105.
The implementation is not finished as of right now so I can get some early feedback before it becomes annoying
to change things.

Btw I had absolutely no idea where to put those types so they are in `rdf4cpp::rdf::util` for now 😅.
Have a look at the tests to see how they can be used.

Copy pasting the issue checklist here for convinience:
- [x] By default, we should not keep the identifier of a blank node loaded from file but assign new ones
- [x] There should be a generator that you can ask for a fresh, so far unused identifier
  - [x] replacable
  - Variants:
    - [x] Random IDs (could be UUIDs)
    - [x] increasing IDs
      - [x] should be as short as allowed in turtle
      - [x] optional user-defined pre- or suffix
- [x] There should be a BlankNodeMapper that keeps track of what blank node identifiers from a file are represented by what blank node identifiers in a NodeStorage. (Scopes)
  - [x] This needs exemplary use-cases to get an idea for a good interface.
  - [x] Check how other libraries handle that
  - [x] Needs to be separated into front and backend. Backend without virtual methods (see tentris)
  - [x] Should be able to support both RDF Dataset semantics merge and union (see https://www.w3.org/TR/rdf11-datasets/#default-graph-as-union-or-as-merge)
~~- [x] There should be subscopes. Subscopes know all blank nodes from their parent scope.~~
  ~~- [x] Subscopes are initialized with all blank nodes from their parent scope. If a blank node is added to a parent scope, it is forwarded to the subscope.~~
- [x] A scope is not bound to a NodeStorage. Each method has a node_storage argument with a default value
- [x] Add support for [Skolem IRIs](https://www.w3.org/TR/rdf11-concepts/#section-blank-nodes)


liss-h:
Open Questions:
1. `BlankNodeIdManager` needs to be able to create BNodes and skolem IRIs. Should it have a template param that determines which it is or should it always be able to create both? 
    - the first one may be problematic with the pimpl nature of the parser, but should be doable
    - the second one has the (small) problem of what to do if you ask for the same label twice but with IRI and BNode
2. Do we need to be able to reverse the translation? `BlankNodeIdManager` then needs an additional map for the other direction.
3. Do we even want a `default_instance()` function? After the meeting today I am not that sure its useful.
4. Do we need scoping of the labels is useful if you want to use the same manager for multiple files? For example when two different files mention the same bnode label but actually mean different things. Or is just using a different manager fine? Alternatively if we don't care about reversibility I could also add a `clear_mapping_state()` function that clears the map but keeps the generator state so you can reuse the same manager but get different ids.

AtesComp:
Some input:
1. No opinion.
2. I don't think reverse engineering is required. However, many implementation have at least a method to identify a Blank Node's internal representation and use it as an IRI.  Several use a generated UUID for that internal use as they need to be universally unique...which leads into 4.
3. I'm not sure of the `default_instance()` function, but if that means a singleton common manager, then that seems to be a good move so that just asking for a BNode without needing to create, hold, and manage a bespoke manager is preferred.  This will likely be the default use.
4. By definition, BNodes are always local and the labels are treated as variables in any meaningful query operation.  Implementations even apply to the graph so that different SELECT queries do not share the same generated BNode. From https://www.w3.org/TR/sparql11-query/#QSynBlankNodes, 4.1.4, states "The same blank node label cannot be used in two different basic graph patterns in the same query." So, while the label may be associated with a BNode during a specific use when generating it, say in an INSERT, the label is meaningless outside of that specific operation.  Then, different files using the same BNode label are absolutely different.  Even in the same file, they should mean different BNodes when in different graphs. Then, retaining the label initially used to create a BNode has little use other than historical.  Maybe someone could use those labels, but they should just be temporary variable names for a given operation.

liss-h:
Thanks for your input :) I changed the interface based on your comments and how this would affect the type's usage in the parser.

src/rdf4cpp/rdf/parser/ParsingState.hpp
```diff
@@ -0,0 +1,76 @@
...
+    node_storage_type node_storage = storage::node::NodeStorage::default_instance();
+
+    /**
+     * The scope that generates blank nodes.
+     * @note default value is a new, empty scope that puts generated bnodes into the default node storage.
+     * @note if the parser has ParsingFlag::KeepBlankNodeIds set this field is ignored
+     * @note the generator has its own node storage,
+     *      a mismatch between this->node_storage and this->blank_node_generator's node storage may not be desired
+     */
+    blank_node_generator_type blank_node_generator = util::BlankNodeIdGenerator::default_instance().scope(this->node_storage);
```
liss-h: Note: internal node storage of `this->blank_node_generator` and `this->node_storage` can be different.
I personally think this is probably undesirable, but there is not really a way (I could think of) that avoids this, 
as:

- `BlankNodeIdScope` needs to know a node storage since it has to put nodes into it and be able to find the same nodes again later when asking for the same label.
- having the parsers node storage be determined by the `BlankNodeIdScope` seems weird and counter intuitive
- this type being aggregate is useful for initializing because of named member init syntax

liss-h:
An unrelated thing: I was thinking about adding flags for the syntax used. 
But I noticed is that `serd` ignores the `SerdSyntax::SERD_NQUADS` flag when using `serd_reader_read_chunk`; It just always reads triples. 

So theoretically the parser can now properly handle different graphs, but serd cannot (with `serd_reader_read_chunk`). I guess thats a task for future improvements then.

AtesComp:
I'm wondering about this comment:

> BlankNodeIdScope needs to know a node storage since it has to put nodes
into it and be able to find the same nodes again later when asking for the
same label.

Due to pervasive reuse of blank node labels that are not the same actual
node, there will be many blank nodes that may register the same label, such
as b0, b1, etc.

I see finding these nodes by label to be highly problematic unless you have
some augmented way to "find" them.  Usually, finding blank nodes I ln a
query is by context to other IRI and literal references...or by some
established reference to the nodes assigned UUID, etc.

Is there some solution to this I'm not seeing?

bigerl:
Thx for the input. We've considered the feedback. Updated requirements for the blank node support in rdf4cpp are in #105.

src/rdf4cpp/rdf/parser/ParsingState.hpp
liss-h: solved by design change

src/rdf4cpp/rdf/util/SharedPtr.hpp
```diff
@@ -0,0 +1,524 @@
...
+    }
+};
+
+/**
+ * Shared pointer implementation, that exposes ptr equality and hash on WeakPtrs.
+ * This implementation is inspired by Rust's std::sync::Arc and boost's boost::shared_ptr.
+ * @tparam T type of object to manage
+ */
+template<typename T>
+struct SharedPtr {
```
liss-h: I honestly don't really want to have a SharedPtr implementation in here but for _some_ reason `std::weak_ptr` and `boost::weak_ptr` both do not expose pointer equality even though that is a perfectly valid operation
on the control block (because weak pointer keep that alive and there is a 1-1 relationship between managed objects and control blocks).

This functionality is needed for fast (and user friendly) comparisions of BlankNodes in merge and union semantics.

This is needed here: https://github.com/rdf4cpp/rdf4cpp/blob/1a7b18ddd4f9c132534d341de7abc300828200cb/src/rdf4cpp/rdf/BlankNode.cpp#L37-L76

liss-h:
Some comments from todays meeting: (you can also add your notes here @bigerl (or just add a review))
- [ ] ~~change `SharedPtr` to be `SharedOffsetPtr`~~
- [x] proper hash combine in BNodeBackendView
- [x] Implement a persistent version for tentris
- [x] NodeScope (and maybe other frontend classes) needs a public way to be constructable for side-loading

src/rdf4cpp/rdf/util/SharedPtr.hpp
bigerl: Makes sense to me. As discussed, a through review might need some second opinion from https://codereview.stackexchange.com/ after it was ported to offset pointers. :)

src/rdf4cpp/rdf/bnode_management/INodeFactory.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include <rdf4cpp/rdf/bnode_management/IIdGenerator.hpp>
+#include <rdf4cpp/rdf/storage/node/NodeStorage.hpp>
+
+namespace rdf4cpp::rdf::util {
+
+struct NodeScope;
+
+/**
+ * A factory for nodes. Any kind of node. That given a generator and scope
```
bigerl:
```suggestion
 * A factory for nodes; so not just BlankNodes. Given a generator and scope
```

src/rdf4cpp/rdf/bnode_management/INodeScope.hpp
```diff
@@ -0,0 +1,46 @@
...
+     */
+    [[nodiscard]] virtual std::optional<std::string_view> find_label(storage::node::identifier::NodeBackendHandle handle) const noexcept = 0;
+
+    /**
+     * Insert a mapping label <-> handle into this scope
+     *
+     * @param label the label of the mapping
+     * @param handle the handle of the mapping
+     */
+    virtual void label_node(std::string_view label, storage::node::identifier::NodeBackendHandle handle) = 0;
```
bigerl: We should probably define here what happens if the label or the node is already mapped. Throw?

bigerl: state in docs that handle must be a valid BNode handle otherwise UB

src/rdf4cpp/rdf/bnode_management/INodeScope.hpp
```diff
@@ -0,0 +1,46 @@
...
+     */
+    [[nodiscard]] virtual storage::node::identifier::NodeBackendHandle find_node(std::string_view label) const noexcept = 0;
+
+    /**
+     * Try to find the label corresponding to the given node handle-
+     *
+     * @param handle the node handle to search for
+     * @return either the label for the given node, or std::nullopt if this scope does not know this handle
+     */
+    [[nodiscard]] virtual std::optional<std::string_view> find_label(storage::node::identifier::NodeBackendHandle handle) const noexcept = 0;
```
bigerl: state in docs that handle must be a valid BNode handle otherwise UB

src/rdf4cpp/rdf/bnode_management/INodeScopeManager.hpp
```diff
@@ -0,0 +1,17 @@
+#ifndef RDF4CPP_INODESCOPEMANAGER_HPP
+#define RDF4CPP_INODESCOPEMANAGER_HPP
+
+#include <string_view>
+
+namespace rdf4cpp::rdf::util {
+
+struct NodeScope;
+
+struct INodeScopeManager {
```
bigerl: add doc

src/rdf4cpp/rdf/bnode_management/NodeGenerator.hpp
```diff
@@ -0,0 +1,140 @@
...
+#include <string_view>
+
+namespace rdf4cpp::rdf::util {
+
+struct NodeScope;
+
+/**
+ * A standalone generator to generate nodes using a given backend.
+ * Nodes generated by this generator are not remembered in any way
+ * and new generate_node/generate_id calls generate new nodes on every invocation
```
bigerl: Should be guaranteed as long as IIdGenerator guarantees it.

src/rdf4cpp/rdf/bnode_management/NodeScope.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    auto old_rc = slot.refcount.load(std::memory_order_relaxed);
+    while (true) {
+        if (old_rc == 0) {
+            return std::nullopt;
+        }
+
+        if (slot.refcount.compare_exchange_weak(old_rc, old_rc + 1, std::memory_order_acquire, std::memory_order_relaxed)) {
+            assert(slot.backend != nullptr);
+            assert(slot.generation == this->generation_);
```
bigerl: If during the loop (e.g. parallel to the old_rc == 0 comparison) the old NodeScope in this bucket is destructed and a new one is constructed (with generation+=1) this assertion will fail. This loop might never observe the old_rc == 0 that happens shortly.

src/rdf4cpp/rdf/bnode_management/NodeScope.hpp
```diff
@@ -0,0 +1,253 @@
...
+ * Instances of this type cannot be constructed directly, use NodeGenerator::scope instead.
+ */
+struct NodeScope {
+    using NodeStorage = storage::node::NodeStorage;
+
+private:
+    friend struct WeakNodeScope;
+    friend struct std::hash<NodeScope>;
+
+    static inline std::array<node_scope_detail::ControlBlock, 1023> node_scope_instances_{};
```
bigerl: This limit is most probably too low. It should be at least in the millions better billions. Who wouldn't want a dataset with a graph for every person on the planet :)

examples/blank_node_management.cpp
```diff
@@ -0,0 +1,95 @@
...
+        print(remembered_b1);
+
+        Node remembered_b1_2 = scope.get_or_generate_node("some-bnode-label", generator);
+        assert(remembered_b1 == remembered_b1_2);
+
+        Node remembered_b2 = scope.get_or_generate_node("other-bnode-label", generator);
+        assert(remembered_b1 != remembered_b2);
+        print(remembered_b2);
+
+        /*{
```
bigerl: why are subscopes commented out here?

src/rdf4cpp/rdf/parser/ParsingFlags.hpp
```diff
@@ -9,6 +9,7 @@ namespace rdf4cpp::rdf::parser {
 enum struct ParsingFlag : uint8_t {
     Strict = 1 << 0,
     NoParsePrefix = 1 << 1,
+    KeepBlankNodeIds = 1 << 2,
```
bigerl, line 17: Based on https://www.w3.org/TR/rdf11-datasets/#default-graph-as-union-or-as-merge we should be able to switch between three BNode behaviors: KeepBlankNodeIds, MergeBNodeSemnatics, UnionBNodeSemantics

bigerl (changes requested):
This is only a first partial review. I'll come back to this.

tests/util/tests_BlankNodeIdManagement.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+using namespace rdf4cpp::rdf::util;
+using namespace rdf4cpp::rdf;
+
+TEST_SUITE("blank node id management") {
+    TEST_CASE("blank node id generation") {
+        SUBCASE("from entropy") {
+            auto gen = NodeGenerator::new_instance();
+
+            auto const id1 = gen.generate_id();
```
bigerl, line 16: deduplicate code with a lambda for subcases from entropy and from seed

tests/util/tests_BlankNodeIdManagement.cpp
```diff
@@ -0,0 +1,143 @@
...
+            CHECK(b1 == b3);
+
+            CHECK(scope.try_get_node("hello").null());
+
+            auto fresh = scope.generate_node();
+            CHECK(fresh != b1);
+            CHECK(fresh != b2);
+
+            /*Node sibling_1;
+            Node sibling_2;
```
bigerl: Did we drop the sibling/subscope thing or why is this deactivated?

tests/util/BlankNodeIdManagement_metall_phase1.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+        BlankNode bnode = generator.generate_node().as_blank_node();
+        std::cout << bnode.backend_handle().raw() << " " << bnode.identifier() << std::endl;
+        assert(bnode.identifier() == "0");
+
+
+        auto scope_impl_ptr = manager.construct<PersistableScope>(scope_name)(manager.get_allocator());
+        NodeScope scope = NodeScope::new_instance<PersistableScopeFrontent>(scope_impl_ptr);
+
+        BlankNode bnode2 = scope.get_or_generate_node("spherical cow", generator).as_blank_node();
```
bigerl: A spherical cow typically means that we oversimplify sth here. I don't see what :)

src/rdf4cpp/rdf/bnode_management/reference_backends/scope/ReferenceNodeScope.hpp
```diff
@@ -0,0 +1,68 @@
...
+        }
+    };
+
+private:
+    std::shared_mutex mutable mutex; // protects label_to_storage and storage_to_label
+
+    storage::util::tsl::sparse_map<PinnedString,
+                                   storage::node::identifier::NodeBackendHandle,
+                                   storage::util::robin_hood::hash<std::string_view>,
+                                   std::equal_to<>> label_to_storage;
```
bigerl: label_to_nodehandle

src/rdf4cpp/rdf/bnode_management/reference_backends/scope/ReferenceNodeScope.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+    storage::util::tsl::sparse_map<storage::node::identifier::NodeBackendHandle,
+                                   std::string_view,
+                                   storage::util::robin_hood::hash<storage::node::identifier::NodeBackendHandle>> storage_to_label;
```
bigerl: nodehandle_to_label

src/rdf4cpp/rdf/bnode_management/NodeScope.hpp
bigerl: What was the reason again why metall shared pointers wouldn't work?

bigerl (changes requested):
some more comments

src/rdf4cpp/rdf/bnode_management/INodeScope.hpp
liss-h: Technically the result of this function is not undefined if you didn't cause UB before by labeling an invalid node

examples/blank_node_management.cpp
liss-h: fixed now

tests/util/tests_BlankNodeIdManagement.cpp
liss-h: Yes, only 1 is supported now using the scopemanager

tests/util/tests_BlankNodeIdManagement.cpp
liss-h: code is different

src/rdf4cpp/rdf/bnode_management/NodeScope.hpp
liss-h: basically polymorphism

shared_ptr<Derived> -> shared_ptr<Base> causes problems with freeing the resulting shared_ptr<Base> using an arbitrary allocator

liss-h: For reference: We talked about this and allowing an arbitrary number of node scopes is not quite so easy.
Will revisit in the future.

src/rdf4cpp/rdf/parser/ParsingFlags.hpp
liss-h: Union semantics achieved by using SingleNodeScopeManager for the parser

src/rdf4cpp/rdf/bnode_management/NodeScope.cpp
liss-h: Good point, need to address that

liss-h:
TODO: 
- [x] put into proper namespace (bnode_mng, rename folder)
- [x] fix example

src/rdf4cpp/rdf/bnode_management/NodeScope.hpp
bigerl: makes sense

tests/util/tests_BlankNodeIdManagement.cpp
bigerl: right, and that makes me wonder. But I'll raise that in a separate comment

tests/util/tests_BlankNodeIdManagement.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+            auto const id1 = gen.generate_id();
+            auto const id2 = gen.generate_id();
+            auto const id3 = gen.generate_id();
+
+            CHECK(id1 != id2);
+            CHECK(id2 != id3);
+            CHECK(id1 != id3);
+
+            auto gen2 = NodeGenerator::new_instance();
```
bigerl, line 24: After your reply on my comment above I double checked it and found that gen2 is not used. Shoudn't id{4-6} be generated with gen2?

Then the code would actually be the same (except how the generator is instantiated, of course).

liss-h: Good catch. But still just almost equal, the last line is also different

bigerl (approved):

PR #137, opened 2023-02-02
Fix: Ownership and thread safety rework of node storage

### LINKED ISSUE #90: Review Ownership Model of NodeStorage and Backends
To make rdf4cpp completely thread safe node storage handles are necessary

Reminder, something like this:
```cpp
struct NodeStorageHandle {
     NodeStorageId index;
     uint32_t generation;
};
```
See also PR Review of #77 for details. To implement those however, a review and overhaul of the ownership model are needed.

liss-h:
Currently Missing:
- [x] Test cases that actually try to race
- [x] some documentation here and there


src/rdf4cpp/rdf/storage/node/NodeStorage.cpp
```diff
@@ -1,142 +1,281 @@
...
+
+    auto it = std::find_if(NodeStorage::node_context_instances.begin(), NodeStorage::node_context_instances.end(), [&](auto const &slot) {
+        return slot.backend == backend_instance;
+    });
+
+    if (it == NodeStorage::node_context_instances.end()) {
+        throw std::runtime_error{"Backend instance not found"};
+    }
+
+    // TODO: just remove this function? It is basically just always UB when the reference count
```
liss-h: Opinion?

bigerl: The idea behind unregister_backend was to have a way to prevent a call to the <del>constructor</del> destructor of a persistent INodeStorage implementation. There might also be other ways.

liss-h: Not 100% sure what you mean by that. Do you mean persistent as in Tentris' node storage or something different?
And how would it prevent that ctor call?

bigerl: Sorry, I meant destructor. Yes, like in Tentris.

liss-h: How about something like rust's [ManuallyDrop](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html) for Tentris. Basically a type that prevents the destructor of an object from running automatically.

In c++ it can be implemented using a `union` because `union` doesn't call constructors or destructors automatically. Haven't thought of how this could be integrated into tentris, just an idea.

bigerl: Reviewing the current design of tentris' persistent node storage, I would say we already have sth. like that with the polymorphic implementation of INodeStorage and it's non-polymorphic implementation that is available via a pointer. 

So no, I don't think we need it anymore

liss-h:
I had an idea for a perf improvement so I am implementing a cache for the backend pointer right now (this saves an atomic load on node storage access). But should be a small change overall so don't wait for that to start the review.

src/rdf4cpp/rdf/storage/node/NodeStorage.cpp
```diff
@@ -1,187 +1,384 @@
...
 
-    node_context_instances[backend_instance->manager_id.value] = backend_instance;
-    return NodeStorage(backend_instance);
-}
-void NodeStorage::unregister_backend(INodeStorageBackend *backend_instance) {
-    if (backend_instance == nullptr)
-        throw std::runtime_error("Backend instance must not be null.");
-    node_context_instances[backend_instance->manager_id.value] = nullptr;
+            return NodeStorage{identifier::NodeStorageID{static_cast<uint16_t>(ix)}, backend_instance};
+        } else if (old_value == backend_instance) {
```
bigerl: `[[unlikely]]`

src/rdf4cpp/rdf/namespaces/util/NamespaceInstance.hpp
```diff
@@ -12,18 +12,20 @@ template<typename NamespaceClass>
 struct NamespaceInstance {
     static NamespaceClass &instance(storage::node::NodeStorage &node_storage = storage::node::NodeStorage::default_instance()) {
         static storage::util::tsl::sparse_map<uint16_t, NamespaceClass> instances;
```
bigerl, line 16: Given that NodeStorage is thread-safe now, it would probably make sense to make Namespaces threadsafe as well.

src/rdf4cpp/rdf/storage/tuple/IDatasetBackend.hpp
```diff
@@ -17,14 +17,14 @@ class IDatasetBackend {
     using QuadPattern = rdf4cpp::rdf::query::QuadPattern;
 
 protected:
-    mutable node::NodeStorage node_storage_;
+    mutable node::WeakNodeStorage node_storage_;
```
bigerl: The original idea behind Datasets having a NodeStorage was that the dataset keeps a NodeStorage alive.

src/rdf4cpp/rdf/Dataset.cpp
```diff
@@ -10,25 +10,30 @@ Dataset::Dataset(Dataset::NodeStorage node_storage)
     : dataset_storage(Dataset::DatasetStorage::new_instance<storage::tuple::DefaultDatasetBackend>(std::move(node_storage))) {}
 
 void Dataset::add(const Quad &quad) {
-    dataset_storage.add(quad.to_node_storage(backend().node_storage()));
+    auto ns = backend().node_storage().upgrade();
+    dataset_storage.add(quad.to_node_storage(ns));
```
bigerl: I guess this would get a bit more expensive that way.

src/rdf4cpp/rdf/Namespace.hpp
```diff
@@ -31,7 +32,7 @@ class Namespace {
     /**
      * NodeStorage from which IRI objects are created.
      */
-    mutable NodeStorage node_storage_;
+    WeakNodeStorage node_storage_;
```
bigerl, line 35: Here I would say that a WeakNodeStorage is the right choice. Otherwise it get's ugly with the default instances.

bigerl (changes requested):
Great work from what I can tell. Code is much clearer now. The concurrency control code matches the description. Argumentation why we won't have critical race condition or deadlocks is clear and seems sound. Please take this review with a grain of salt as I am not an expert in concurrency stuff. 

Some minor points are listed below.

src/rdf4cpp/rdf/storage/tuple/IDatasetBackend.hpp
liss-h: I guess that makes sense. The reason why i changed it to weak was that there is a default instance for the dataset storage which will keep effectively the default node storage alive forever. If that is expected then I will change it back.

liss-h: I would however suggest to change the initialization behaviour of the DatasetStorage default instance to only get initialized when first accessed and not on program start

src/rdf4cpp/rdf/Dataset.cpp
liss-h: reverted

src/rdf4cpp/rdf/storage/tuple/IDatasetBackend.hpp
liss-h: is reverted

bigerl (approved):

PR #138, opened 2023-02-02
Canonicalize lang tag instead of doing case insentive comparison

### LINKED ISSUE #135: Lang Tag canonicalizations
langTags are currently not canonicalized into storage, instead they get special treatment for comparisions.
This results in identical langStrings with different casing in langTags being assigned different backend ids. Which is
not optimal for performance in comparisions.

liss-h:
Closes #135


bigerl (approved):
Looks fine and makes comparing Lang-tagged strings so much cheaper.

PR #139, opened 2023-02-02
Feature: add checked conversion functions to node

### LINKED ISSUE #125: Improvement: Node to subtype casting ergonomics
Casting `Node` to `Literal` is currently tedious, as the node type needs to be checked manually before doing the actual casting operation. A more convinient interface would be something like the following.

```c++
struct Node { 
    ...

    /**
      * Casts this Node into a Literal if it is one, otherwise returns the null literal.
      */ 
    Literal as_literal_checked() const noexcept;

    /**
     * Casts this Node into an IRI if it is one, otherwise returns the null iri.
     */
    IRI as_iri_checked() const noexcept;

    BNode as_bnode_checked() const noexcept;
    query::Variable as_variable_checked() const noexcept;
    ...
};
```

This addition would e.g. change this:

```c++
// given inputs
Node n1{...};
Node n2{...};

if (!n1.is_literal() || !n2.is_literal()) {
   return Literal{};
}

Literal l1 = static_cast<Literal>(n1);
Literal l2 = static_cast<Literal>(n2);

return l1 + l2;
```

```c++
// given inputs
Node n1{...};
Node n2{...};

// this works because the transformative functions on e.g. literal check for null anyways.
// by transformative functions is mean functions that take one or multiple literals and turn it into another literal
return n1.as_literal_checked() + n2.as_literal_checked();
```

liss-h:
Closes #125

### DIFF
--- a/src/rdf4cpp/rdf/writer/NQuadsWriter.cpp
+++ b/src/rdf4cpp/rdf/writer/NQuadsWriter.cpp
@@ -14,7 +14,7 @@ NQuadsWriter::operator std::string() const {
 }
 std::ostream &operator<<(std::ostream &os, const NQuadsWriter &writer) {
     for (const Quad &quad : writer.dataset_) {
-        if (not(quad.graph().is_iri() and (static_cast<IRI>(quad.graph())) == IRI::default_graph(writer.dataset_.backend().node_storage())))
+        if (!quad.graph().null() && quad.graph().as_iri() != IRI::default_graph(writer.dataset_.backend().node_storage()))
             os << NNodeWriter(quad.graph()) << " ";
         os << NNodeWriter(quad.subject()) << " "
            << NNodeWriter(quad.predicate()) << " "
--- a/examples/rdf_nodes.cpp
+++ b/examples/rdf_nodes.cpp
@@ -68,7 +68,7 @@ int main() {
     }
 
     if (nodes[1].is_variable()) {
-        auto v = static_cast<query::Variable>(nodes[1]);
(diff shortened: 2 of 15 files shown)


src/rdf4cpp/rdf/writer/NQuadsWriter.cpp
```diff
@@ -14,7 +14,7 @@ NQuadsWriter::operator std::string() const {
 }
 std::ostream &operator<<(std::ostream &os, const NQuadsWriter &writer) {
     for (const Quad &quad : writer.dataset_) {
-        if (not(quad.graph().is_iri() and (static_cast<IRI>(quad.graph())) == IRI::default_graph(writer.dataset_.backend().node_storage())))
+        if (!quad.graph().null() && quad.graph().as_iri() != IRI::default_graph(writer.dataset_.backend().node_storage()))
```
bigerl, line 17: NOT(a AND b) = NOT a OR NOT b -> `||`

bigerl, line 17:
```suggestion
        if (quad.graph().as_iri() != IRI::default_graph(writer.dataset_.backend().node_storage()))
```

Should be sufficient, right?

liss-h: but `null-iri != default_graph_iri` and then we would try to print a null-iri

bigerl: you are right

bigerl (approved):
looks good

PR #143, opened 2023-02-03
Fix: literal misc functions defects

### LINKED ISSUE #142: Literal::lang_matches behaviour
Currently there is `Literal::lang_matches(std::string_view lang_range)` which checks if the language tag of the given literal matches `lang_range`. I.e. the following C++ code:

```c++
auto lit = Literal::make_lang_tagged("Liss", "en");
return lit.lang_matches("en");
```
Is equivalent to this SPARQL code (notice the additional inserted `lang()` call)
```sparql
langMatches(lang("Liss"@en) , "en")
```

<br>----------------------------------

If `Literal::lang_matches` were to behave like the standard wants it should be more like the following:
```c++
auto lit = Literal::make_simple("en");
return lit.lang_matches("en");
```

To get the old behaviour back you would then need to write (after another function: `Literal::as_language_tag` is added):
```c++
auto lit = Literal::make_lang_tagged("Liss", "en");
return lit.as_language_tag().lang_matches("en");
```

<br>----------------------------------

Solutions:
Personally I prefer the current behaviour so I would keep it as well, only problem is naming:

Best I could come up with, still ugly:
```c++
struct Literal {
    /**
     * returns true if this->language_tag() matches lang_range
     */
    util::TriBool language_tag_matches_range(std::string_view lang_range) const noexcept;
    Literal as_language_tag_matches_range(std::string_view lang_range, NodeStorage &) const noexcept;
    Literal as_language_tag_matches_range(Literal const &lang_range, NodeStorage &) const noexcept; // lang_range is xsd:string

    /**
     * returns true if this->language_tag() == other.language_tag()
     * (also rename Literal::datatype_matches to Literal::datatype_eq)
     */
    util::TriBool language_tag_eq(Literal const &other) const noexcept;

    /**
     * effectively returns Literal::make_simple(this->language_tag());
     * for good measure because as_lexical_form() also exists 
     * (and can be implemented more efficiently if we have lang tag inlining in the future)
     */
    Literal as_language_tag(NodeStorage &) const noexcept;
};

util::TriBool lang_matches(std::string_view lang_tag, std::string_view lang_range);
Literal lang_matches(Literal const &lang_tag, Literal const &lang_range); // needs 2 xsd:strings
```

liss-h:
Closes #142.

I also discovered some additional defects that were either not optimal in performance or missing a transformation function.

And the tests somehow got wrongly formatted (i.e. something like `5_literal.function()` is invalid because of c++ parsing stuff).

### DIFF
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -354,6 +354,10 @@ bool Literal::is_iri() const noexcept { return false; }
 bool Literal::is_numeric() const noexcept {
     using namespace datatypes::registry;
 
+    if (this->null()) {
+        return false;
+    }
+
     return this->handle_.node_id().literal_type().is_numeric()
            || DatatypeRegistry::get_numerical_ops(this->datatype_id()) != nullptr;
 }
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -72,14 +72,17 @@ Literal Literal::make_typed_unchecked(std::any const &value, datatypes::registry
 }
 
 Literal Literal::make_string_like_copy_lang_tag(std::string_view str, Literal const &lang_tag_src, Node::NodeStorage &node_storage) noexcept {
-    auto const lt_dt = lang_tag_src.datatype_id();
-
-    if (lt_dt == datatypes::rdf::LangString::datatype_id) {
+    if (lang_tag_src.datatype_eq<datatypes::rdf::LangString>()) {
         return Literal::make_lang_tagged_unchecked(str, lang_tag_src.language_tag(), node_storage);
     }
 
-    assert(lt_dt == datatypes::xsd::String::datatype_id);
+    assert(lang_tag_src.datatype_eq<datatypes::xsd::String>());
     return Literal::make_simple_unchecked(str, node_storage);
+}
+
+bool Literal::dynamic_datatype_eq_impl(std::string_view datatype) const noexcept {
+    assert(!this->is_fixed());
+    return this->datatype().identifier() == datatype;
 }
 
 Literal Literal::make_simple(std::string_view lexical_form, Node::NodeStorage &node_storage) {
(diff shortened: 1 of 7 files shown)


src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -740,12 +855,12 @@ bool Literal::is_fixed_not_numeric() const noexcept {
 }
 
 bool Literal::is_string_like() const noexcept {
-    if (this->null()) {
+    auto const type = this->handle_.node_id().literal_type();
+    if (!type.is_fixed()) {
```
bigerl, line 863: `Literal::is_fixed()` has an assertion `!this->null()`. Should we define a deterministic behavior for the `is_*` methods? e.g., they always return false.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1260,6 +1408,33 @@ Literal Literal::substr(Literal const &start, Literal const &len, Node::NodeStor
...
+bool lang_matches(std::string_view const lang_tag, std::string_view const lang_range) noexcept {
+    if (lang_range.empty()) {
+        return lang_tag.empty();
+    }
+
+    if (lang_range == "*") {
+        return !lang_tag.empty();
+    }
+
+    auto const lang_ci = util::CiStringView{lang_tag.data(), lang_tag.size()};
```
bigerl, line 1424: Add a comment that it is case-insensitive

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1260,6 +1408,33 @@ Literal Literal::substr(Literal const &start, Literal const &len, Node::NodeStor
...
+    auto const lang_range_ci = util::CiStringView{lang_range.data(), lang_range.size()};
+
+    return lang_ci.starts_with(lang_range_ci) && (lang_ci.size() == lang_range_ci.size() || lang_ci[lang_range_ci.size()] == '-');
```
bigerl, line 1428: possible buffer over-read
```
lang_ci[lang_range_ci.size()] == '-'
```

tests/nodes/tests_Literal.cpp
```diff
@@ -501,48 +501,48 @@ TEST_CASE("Literal - misc functions") {
         CHECK(99_xsd_int .round() == 99_xsd_integer);
         CHECK(1.2_xsd_double .round() == 1.0_xsd_double);
         CHECK(1.5_xsd_double .round() == 2.0_xsd_double);
-        CHECK("hello"_xsd_string.round().null());
+        CHECK("hello"_xsd_string .round().null());
```
bigerl: What is the plan here? Only viable method working with clang-format seems to be putting parenthesis around. Googled for it but didn't find anything related. 

Could you please fill a bug report for clang-format?

bigerl (changes requested):
Looks good and sensible. Some small issues listed below.

src/rdf4cpp/rdf/Literal.cpp
liss-h: My general idea was:
- public transformation functions (i.e. `Literal -> Literal functions`) check for null and return back a null literal (for chaining)
- public non-transformation functions (i.e. `Literal -> other type`) usually do not check for null, but if the null-literal has a meaningful value for that function it will behave correctly (e.g. for `null-lit <=> non-null-lit`)
- private functions in general (except where necessary) do not check for null
- i guess public `is_*` functions should check for null to make them harder to misuse

Its probably not entirely consistent right now, so if you find anything more that is inconsistent with these rules feel free to mention it here. Alternatively if you don't agree with them, suggest changes.

- [x] fix `is_numeric()` to check for null
- [x] Another thing would be that it is probably a good idea to write those rules down somewhere.

src/rdf4cpp/rdf/Literal.cpp
liss-h: No, at least not if I am missing something (which is definitely possible), because of short-circuiting we know the following during evaluation of `lang_ci[lang_range_ci.size()] == '-'`:

1. `lang_ci.starts_with(lang_range_ci)` otherwise the second part of the `&&` would not evaluate
    - therefore `lang_range_ci.size() <= lang_ci.size()`
2. `lang_ci.size() != lang_range_ci.size()` because otherwise the second part of the `||` would not evaluate
    - therefore `lang_range_ci.size() < lang_ci.size()` (strictly less than)

therefore `lang_range_ci[lang_range_ci.size()]` is strictly in bounds


https://godbolt.org/z/szrhEh6v7

bigerl: You are right. clever :)

tests/nodes/tests_Literal.cpp
liss-h: Filed a bug: https://github.com/llvm/llvm-project/issues/60576

I also found out that this is only invalid formatting for numeric user-defined-literals but not string user-defined-literals. I guess I'll wrap everything in parentheses so that it will not break by accident in the future

liss-h: And inserted parentheses

PR #145, opened 2023-02-07
Feature: Add Sanitizers to CI

liss-h:


liss-h:
Not 100% certain that they are enabled now but compiling takes _ages_ now so I would guess they are.

bigerl (changes requested):
Based on the build-times I would enable the sanitizers only on clang-14 static builds. To pass in CXXFLAGs via env is fine here.

.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
```diff
@@ -67,12 +67,12 @@ jobs:
...
         run: |
           cmake -G Ninja -B build_dir -DBUILD_TESTING=ON -DBUILD_EXAMPLES=ON \
-          ${{ fromJSON('["", "-DBUILD_SHARED_LIBS=ON -DCMAKE_POSITION_INDEPENDENT_CODE=ON"]')[inputs.build-shared-libs] }} \
-          ${{ fromJSON('["", "-DCMAKE_CXX_FLAGS=--coverage"]')[inputs.with-coverage] }}  
+          ${{ fromJSON('["", "-DBUILD_SHARED_LIBS=ON -DCMAKE_POSITION_INDEPENDENT_CODE=ON"]')[inputs.build-shared-libs] }}
           cmake --build build_dir
         env:
           CC: ${{ steps.install_cc.outputs.cc }}
           CXX: ${{ steps.install_cc.outputs.cxx }}
+          CXXFLAGS: -fsanitize=address -fsanitize=undefined ${{ fromJSON('["", "--coverage"]')[inputs.with-coverage] }}
```
bigerl: [clang suggests](https://clang.llvm.org/docs/AddressSanitizer.html#id3) to use <del>clang++ for linking and</del> (I think that is anyways by invoking the linker through the compiler) to keep function frame pointers

liss-h: Added `-fno-omit-frame-pointer -Wno-unused-command-line-argument -fuse-ld=lld`

`-Wno-unused-command-line-argument` is just that clang does complain on every cpp file getting _compiled_ that it doesn't use the linker, which is expected anyways at compile stage.

liss-h:
Ok after rerunning the CI once it appears to work now, no idea why it didn't work the first time

bigerl (changes requested):
Looks good now. Please also add an info "with address and UB sanitizers" to clang-14 static's job name

bigerl (approved):

PR #146, opened 2023-02-09
Feature: per type node storages

### LINKED ISSUE #114: Improvement: native-type backend Storage
Currently, all datatype values are either stored encoded as canonical strings in the backend or inlined into the NodeBackendHandle. 
While working with inlined values is almost for free, working with values in the backend is expensive. They need to be parsed from string to their native type every time they are used. 

The proposed solution is to customize the backend storage and it's access in such a way that it stores the values instead of the string representation for certain types in the backend. This could be a feature of a type definition, e.g. like inlineable.

liss-h:
Solves #114


src/rdf4cpp/rdf/storage/node/reference_node_storage/ReferenceNodeStorageBackend.cpp
```diff
@@ -1,144 +1,259 @@
...
+        identifier::NodeID const next_id = next_id_func();
+        auto const [it, inserted] = storage.id2data.emplace(next_id, std::make_unique<typename Storage::Backend>(view));
+        assert(inserted);
+        storage.data2id.emplace(it->second.get(), next_id);
+
+        return next_id;
     }
 }
 
 identifier::NodeID ReferenceNodeStorageBackend::find_or_make_id(view::LiteralBackendView const &view) noexcept {
```
bigerl, line 155: This implementation requires from the Literal implementation to know what types are supported by the backend and prepare them accordingly. Some assertions would be good that help debugging if we ever add things that do not adhere to those rules. Also, the rule should be explicitly stated somewhere with a big caution sign.

src/rdf4cpp/rdf/storage/node/reference_node_storage/ReferenceNodeStorageBackend.cpp
```diff
@@ -1,144 +1,259 @@
...
+            [this](view::LexicalFormLiteralBackendView const &lexical) noexcept {
+                auto const next_id_func = [this, datatype = lexical.datatype_id]() noexcept {
+                    return identifier::NodeID{next_literal_id++, identifier::iri_node_id_to_literal_type(datatype)};
+                };
+
+                return lookup_or_insert_impl<true>(lexical, fallback_literal_storage_, next_id_func);
+            },
+            [this](view::ValueLiteralBackendView const &any) noexcept {
+                auto const next_id_func = [this, datatype = any.datatype]() noexcept {
+                    return identifier::NodeID{next_literal_id++, datatype};
```
bigerl, line 166: We need a next_literal_id per LiteralType::underlying_type. So for 2^6 types. For the types that consume more than 42-bits we should probably define a bit-depth threshold up to what they are inlined and reserve the remaining IDs for other values. 

Also, we should throw an exception or terminate if ids are exceeded.  

Addressing this might be something to create an issue and adress in a subsequent PR. Makes reviewing it much easier.

tests/datatype/tests_String.cpp
```diff
@@ -24,11 +24,6 @@ TEST_CASE("Datatype String") {
     auto type_iri = IRI(datatypes::xsd::String::identifier);
     CHECK(type_iri.identifier() == correct_type_iri_cstr);
```
bigerl, line 26: We need some tests for the native-type storage part. My suggestion would be to make things protected instead of private and inherit from it for the tests to gain access to internals.

bigerl (changes requested):
Very nice. Some things need further work. Please review the comments below.

bigerl:
- [x] final todo: update documentation

src/rdf4cpp/rdf/storage/node/reference_node_storage/ReferenceNodeStorageBackend.cpp
liss-h: added a note in the readme

src/rdf4cpp/rdf/storage/node/reference_node_storage/ReferenceNodeStorageBackend.cpp
liss-h: Already prepared the PR #148 by branching off this branch. You can review it when this is merged

tests/datatype/tests_String.cpp
liss-h: I went for a different, more indirect approach. Instead of exposing the internals I am now checking Literal construction and value extractions for consistency which is effectively the same thing, because an improperly implemented backend would sometimes return the wrong type of view or the wrong type of value.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -471,7 +471,8 @@ std::any Literal::value() const noexcept {
 
                 return std::any{};
             },
-            [](storage::node::view::ValueLiteralBackendView const &any) noexcept {
+            [&datatype](storage::node::view::ValueLiteralBackendView const &any) noexcept {
```
bigerl, line 474: Add an maybe_unused to datatype? Is that even possible here? 🤔

bigerl (approved):

src/rdf4cpp/rdf/Literal.cpp
liss-h: No not possible but `(void)datatype;` should do it, added that

PR #147, opened 2023-02-13
Feature/single iterable parse rdf

### LINKED ISSUE #81: Single Iterable for Parsing RDF file
with #76 a iterable parser for RDF is introduced that is very efficient and flexible but needs several lines in the code to be used. Additionally, it would be nice to have an iterable that can be used in the default case where one wants to extract triples or quads from a file:

```cpp
for (auto &entry : RDFFileParser{"myrdffile.ttl"})
    if (entry.has_value())
        std::cout << entry.value() << '\n';
    else
        std::cerr << entry.error() << '\n';
```

An (unfinished) draft could look like this:

```cpp
class RDFFileParser {
    std::string file_path_;

public:
    explicit RDFFileParser(std::string file_path, ParsingFlags flags = ParsingFlags::none(), storage::node::NodeStorage node_storage = storage::node::NodeStorage::default_instance())
        : file_path_(std::move(file_path)) {}

    class iterator {
        std::unique_ptr<std::ifstream> ifstream_;
        std::unique_ptr<IStreamQuadIterator> iter_;

    public:
        using value_type = IStreamQuadIterator::value_type;
        using reference = IStreamQuadIterator::reference;
        using pointer = IStreamQuadIterator::pointer;
        using difference_type = IStreamQuadIterator::difference_type;
        using iterator_category = IStreamQuadIterator::iterator_category;
        using istream_type = IStreamQuadIterator::istream_type;

        iterator() = default;

        explicit iterator(std::string const &file_path)
            : ifstream_([&file_path]() -> std::unique_ptr<std::ifstream> {
                  std::ifstream ifs{file_path};
                  if (ifs.is_open())
                      return std::make_unique<std::ifstream>(std::move(ifs));
                  else
                      return {};
              }()),
              iter_([this]() -> std::unique_ptr<IStreamQuadIterator> {
                  if (this->iter_)
                      return std::make_unique<IStreamQuadIterator>(*this->ifstream_);
                  else
                      return {};
              }()) {}

    protected:
        [[nodiscard]] bool is_at_end() const noexcept {
            return bool(ifstream_) or iter_->is_at_end();
        }

    public:
        reference operator*() const noexcept { return (*iter_).operator*(); }
        pointer operator->() const noexcept { return (*iter_).operator->(); }
        iterator &operator++() {
            ++(*iter_);
            return *this;
        }

        bool operator==(bool other) const noexcept {
            return not is_at_end() == other;
        }
        bool operator==(iterator const &other) const noexcept { return (*iter_) == (*other.iter_); }
        bool operator!=(iterator const &other) const noexcept { return (*iter_) != (*other.iter_); }
    };

    [[nodiscard]] iterator begin() const {
        return iterator{file_path_};
    }

    [[nodiscard]] bool begin() const {
        return false;
    }
};
```

mcb5637:
implementing #81


src/rdf4cpp/rdf/parser/RDFFileParser.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+            reference operator*() const noexcept;
+            pointer operator->() const noexcept;
+            Iterator& operator++();
+            [[nodiscard]] bool operator==(const Iterator& other)const noexcept;
+            // != gets generated by compiler
+        };
+
+        [[nodiscard]] Iterator begin();
+        std::default_sentinel_t end();
```
bigerl: I didn't know they added that to C++20. I always used a bool instead. The bool has one advantage that you can check with `if (bool(iter))` weather an iterator is not at the end yet. So, you don't need an end iterator to compare with. 
Having said that, I think we don't need the bool conversion here and then the `std::default_sentinel_t` is the much cleaner standard way of doing it.

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
```diff
@@ -0,0 +1,68 @@
...
+     * }
+     * @endcode
+     */
+    class RDFFileParser {
+        std::string filePath_;
+        ParsingFlags flags_;
+        storage::node::NodeStorage node_storage_;
+
+    public:
+        explicit RDFFileParser(const std::string& filePath, ParsingFlags flags = ParsingFlags::none(),
```
bigerl: please use snake_case for instance names (no matter if variable, parameter or static value.

tests/CMakeLists.txt
```diff
@@ -296,4 +304,5 @@ set_property(TARGET tests_Namespace_life_cycle PROPERTY CXX_STANDARD 20)
 add_test(NAME tests_Namespace_life_cycle COMMAND tests_Namespace_life_cycle)
 
 # copy files for testing to the binary folder
-# file(COPY foldername DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
\ No newline at end of file
+# file(COPY foldername DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+file(COPY ./parser/tests_RDFFileParser_simple.ttl DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
bigerl:
```suggestion
file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/parser/tests_RDFFileParser_simple.ttl" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
```

In CMake, `${ ... }` should be used within a string. 

The second thing is probably more a matter of taste: please use an absolute path here. Double check that the suggestion above really works (i.e. delete the cmake folder and run cmake again).

tests/parser/tests_RDFFileParser_simple.ttl
```diff
@@ -0,0 +1 @@
+<http://data.semanticweb.org/workshop/admire/2012/paper/12> <http://purl.org/dc/elements/1.1/subject> "search" .
```
bigerl: Please add two more triples

bigerl (changes requested):
Thank you for your PR. Despite the many comments, it looks already pretty good. I just found minor things. 
 
Further todos: 
- [x] Add an example
- [x] Format the files with changes with clang-format. There is an clang-format config file at the root folder of the repo. You can also set CLion up such that it uses clang-format with that file whenever you ask it to format a file.

src/rdf4cpp/rdf/parser/RDFFileParser.cpp
```diff
@@ -0,0 +1,48 @@
+#include "RDFFileParser.hpp"
+
+rdf4cpp::rdf::parser::RDFFileParser::RDFFileParser(const std::string &file_path, rdf4cpp::rdf::parser::ParsingFlags flags,
```
liss-h: Please wrap the definitions in `namespace rdf4cpp::rdf::parser { }`. Without it the signal-to-noise ratio is quite high.

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
```diff
@@ -0,0 +1,68 @@
...
+    explicit RDFFileParser(const std::string &file_path, ParsingFlags flags = ParsingFlags::none(),
+                           storage::node::NodeStorage node_storage = storage::node::NodeStorage::default_instance());
+    explicit RDFFileParser(std::string &&file_path, ParsingFlags flags = ParsingFlags::none(),
+                           storage::node::NodeStorage node_storage = storage::node::NodeStorage::default_instance());
+
+    class Iterator {
+        friend class RDFFileParser;
+        friend bool operator==(const RDFFileParser::Iterator &iter, std::default_sentinel_t) noexcept;
+        std::ifstream stream_;
+        IStreamQuadIterator iter_;
```
liss-h: This here is actually UB; very hard to trigger UB. I couldn't get it to trigger in GCC at all for some reason, but using clang with `-O0 -fno-elide-constructors -fsanitize=address` the following code snippet crashes:

```c++
rdf4cpp::rdf::parser::RDFFileParser parser{"./tests_RDFFileParser_simple.ttl"};
auto it1 = parser.begin();
++it1;
```

**TLDR:**
The core problem is that moving a self-referential struct (which iterator is) is in almost all cases UB (except if you specifically account for that).

**Long version:**
The reason is that the iterator is moved once (because I forced the compiler to actually do it, normally this would be elided). Moving the iterator changes the address of the iteratator object, and therefore the address of `stream_` as well. But `IStreamQuadIterator` internally holds a pointer to the given stream, which of course still points to the old address of `stream_`, i.e. it points into garbage memory because `stream_` was moved elsewhere (into `it1` from the return slot of `begin()`).

Basically moving the iterator (and then using it) is UB at the moment, even if its hard to trigger UB because the optimizer is obscuring it. Thats why the snippet from the issue had `unique_ptrs`, because they heap allocate and thus effectively "pin" the object in place, the only thing you are moving then when moving the iterator is the pointer to the object, which is of course fine.

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
```diff
@@ -0,0 +1,68 @@
...
+        using istream_type = IStreamQuadIterator::istream_type;
+
+        reference operator*() const noexcept;
+        pointer operator->() const noexcept;
+        Iterator &operator++();
+        [[nodiscard]] bool operator==(const Iterator &other) const noexcept;
+        // != gets generated by compiler
+    };
+
+    [[nodiscard]] Iterator begin();
```
liss-h: `begin()` could be `const`
and `end()` could be `const noexcept`

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
```diff
@@ -0,0 +1,68 @@
+#ifndef RDF4CPP_RDFFILEPARSER_HPP
+#define RDF4CPP_RDFFILEPARSER_HPP
+
+#include "IStreamQuadIterator.hpp"
```
liss-h: For the sake of consistency with the rest of rdf4cpp I would include this as 
`#include <rdf4cpp/rdf/parser/IStreamQuadIterator.hpp>`

This of course works, its just very minor a consistency thingy

src/rdf4cpp/rdf/parser/RDFFileParser.cpp
```diff
@@ -0,0 +1,48 @@
+#include "RDFFileParser.hpp"
```
liss-h: see other comment

liss-h (changes requested):
Thanks for your PR :)
I mostly found some very minor things.
But there is one major problem that definitely needs to be addressed

liss-h:
Oh and btw, this is unrelated to your work but could you add a defaulted move assignment operator to `IStreamQuadIterator`? Would be nice to have I just forgot to provide that one

src/rdf4cpp/rdf/parser/RDFFileParser.cpp
liss-h: I see you wrapped it in `namespace` but did not shorten the identifiers. If you didn't know yet (nothing to worry about) if you wrap it in namespace (like you now did) you can shorten for example this:
```c++
rdf4cpp::rdf::parser::RDFFileParser::RDFFileParser(const std::string &file_path, rdf4cpp::rdf::parser::ParsingFlags flags,
                                                   rdf4cpp::rdf::storage::node::NodeStorage node_storage) {
    ...                                                  
}
```

like so:
```c++

RDFFileParser::RDFFileParser(const std::string &file_path, ParsingFlags flags, 
                             storage::node::NodeStorage node_storage) {
    ...
}

...

}  // namespace rdf4cpp::rdf::parser
``` 

which is much more readable if there are many definitions. Thats why I originally suggested the change

liss-h:
Oh and general PR note: I think everybody of us is totally fine with it if you just mark really simple issues as resolved yourself after you fixed them (for example the issue regarding marking `begin()` as `const`). Only if you want to of course, but I usually use that like "checking off entries in a todo list" :)

liss-h (approved):

PR #152, opened 2023-02-20
encode_for_uri

mcb5637:
implementing ENCODE_FOR_URI of #99


src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1,7 +1,9 @@
 #include "Literal.hpp"
 
-#include <sstream>
 #include <random>
+#include <sstream>
+
+#include <utf8.h>
```
liss-h: This include is unecessary as `Literal.cpp` already includes `<rdf4cpp/rdf/util/Utf8.hpp>` (which includes this file). That file was needed to work around some kind of problem I don't quite remember

liss-h (changes requested):
Thanks for the PR; looks fine to me :) . One minor issue with the includes

liss-h:
good catch with the utf8 exception :), didn't think of that

Issue #150, opened 2023-02-16
Bug: lower and uppercase unicode support


liss-h:
Currently the uppercase and lowercase functions of Literal only support ASCII and probably mangle the string if it is unicode.

mcb5637:
implemented via Boost::Locale
crrently does not convert `ß` tp `SS`, everything else seems to work as expected

PR #153, opened 2023-02-22
use boost to convert case of string literals

mcb5637:
fix for #150

### DIFF
--- a/.github/workflows/reusable_fetch_content.yml
+++ b/.github/workflows/reusable_fetch_content.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Install systemlib boost for FetchContent
         run: |
           # FetchContent does not handle transitive dependencies and doesn't run conan.
-          sudo apt-get install libboost-math-dev
+          sudo apt-get install libboost-math-dev libboost-locale-dev
       - name: Check out sources
         uses: actions/checkout@v3
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,6 +200,7 @@ target_link_libraries(rdf4cpp
         PUBLIC
         nonstd::expected-lite
         Boost::headers
+        Boost::locale
         PRIVATE
         fmt::fmt-header-only
         re2::re2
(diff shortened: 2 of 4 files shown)


bigerl:
The fetch-content workflow is failing with 
> Target "rdf4cpp" links to target "Boost::locale" but the target was not
  found. 

The respective package must be installed here: https://github.com/rdf4cpp/rdf4cpp/blob/develop/.github/workflows/reusable_fetch_content.yml#L43

For the CMake FetchContent retrieval we assume that the boost packages are installed on the system already.

bigerl (changes requested):
Thank you for the PR. This review covers only the CI part. 

The fetch-content workflow is failing with 
> Target "rdf4cpp" links to target "Boost::locale" but the target was not
  found. 

The respective package must be installed here: https://github.com/rdf4cpp/rdf4cpp/blob/develop/.github/workflows/reusable_fetch_content.yml#L43

For the CMake FetchContent retrieval we assume that the boost packages are installed on the system already.

liss-h (changes requested):
Hi :) impl looks fine, like mentioned in the meeting last week, my only concern is that the boost conversion function uses a locale, which, at least to me, does not make any sense. So I am currently not really convinced that this does the correct thing in all cases.

Reading the documentation for `boost::locale::to_upper` does not inspire that much confidence in me either. But there is probably nothing (at least nothing reasonable) that would convince me fully that this does what it is supposed to.

So:
- [x] So please add a comment in the functions that points to the part in the documentation that convinced you this does the correct thing
- [x] Also please add some more test cases that cover some common non-ascii characters in german or other languages
- [x] Additionally, please add a test case that ensures commas (in numbers) do not get converted. Realistically it would make absolutely no sense if this happens, just want to be sure.

bigerl (approved):

bigerl:
(Approval is for the changes regarding the CI)

liss-h (approved):
Additions look good. One last thing: please add a comment under #150 explaining the progress and the thing(s) not working

bigerl:
This came up today in my GitHub's explore repositories: https://github.com/uni-algo/uni-algo

Might be worth another look. 

Data for cases states to be up-to-date, i.e. 15.0.0: https://github.com/uni-algo/uni-algo/blob/main/include/uni_algo/impl/data/data_case.h

bigerl:
#160 merged, closing this one

PR #154, opened 2023-02-24
Feature: Port to Conan 2.0

bigerl:
I see right now no way how one could easily use conan 2.0 within CLion. As I see it, either CLion would need better support for CMakePreset creating processes (like conan 2.0) or we would need an updated cmake integration for conan. 

I'll keep this one draft and open for when conan evolved.


liss-h:
Superceeded by #221

PR #156, opened 2023-02-28
hash funtions

### LINKED ISSUE #98: Feature: 17.4.6 Hash functions
https://www.w3.org/TR/sparql11-query/#func-hash

- [x] 17.4.6.1 MD5
- [x] 17.4.6.2 SHA1
- [x] 17.4.6.3 SHA256
- [x] 17.4.6.4 SHA384
- [x] 17.4.6.5 SHA512

mcb5637:
implementing #98 via OpenSSL

### DIFF
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -12,6 +12,8 @@
 #include <rdf4cpp/rdf/util/CaseInsensitiveCharTraits.hpp>
 #include <rdf4cpp/rdf/util/Utf8.hpp>
 
+#include <openssl/evp.h>
+
 namespace rdf4cpp::rdf {
 
 Literal::Literal(Node::NodeBackendHandle handle) noexcept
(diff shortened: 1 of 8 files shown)


bigerl:
I've cleaned the GitHub Actions cache. The separation between different configurations was beached somehow. Should be fine again, now. 
I also revered the changes to the CMake files regarding OpenSSL. 

One build problem with gcov remains. Please investigate that locally in your local environment. Right now, gcov is run but not uploaded. So, if there is no simple solution, just deactivate gcov in the CI.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1556,6 +1558,48 @@ Literal Literal::substr(Literal const &start, Literal const &len, Node::NodeStor
     return this->substr(start_ix, len_ix, node_storage);
 }
 
+Literal Literal::hash_with(const char *alg, NodeStorage &node_storage) const {
+    if (!this->is_string_like())
```
liss-h: This passes for `rdf:langString` and `xsd:string` but as far as I can tell these functions are only supposed to work for `xsd:string`

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -920,6 +920,43 @@ class Literal : public Node {
                                  Literal const &len = Literal::make_typed_from_value<datatypes::xsd::Double>(std::numeric_limits<datatypes::xsd::Double::cpp_type>::infinity()),
                                  NodeStorage &node_storage = NodeStorage::default_instance()) const noexcept;
 
+private:
+    /**
+     * hash via OpenSSL
+     * @param alg OpenSSL hash algorithm identifier
```
liss-h: A link to the options for `alg` would be nice here

liss-h (changes requested):
Thank you; again only small details to fix

liss-h (approved):

PR #159, opened 2023-03-06
compound assignment operators

### LINKED ISSUE #67: Feature Request: Literal op+assign operators
Literal would be more convenient to use if it had operators +=,-=, *=, /=

mcb5637:
closes #67

### DIFF
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -974,10 +974,14 @@ Literal Literal::operator+(Literal const &other) const noexcept {
     return this->add(other);
 }
 
-Literal &Literal::operator+=(const Literal &other) noexcept {
-    auto result = *this + other;
+Literal &Literal::add_assign(const Literal &other, NodeStorage &node_storage) noexcept {
+    auto result = this->add(other, node_storage);
     this->handle_ = result.handle_;
     return *this;
+}
+
+Literal &Literal::operator+=(const Literal &other) noexcept {
+    return this->add_assign(other);
 }
 
 Literal Literal::sub(Literal const &other, Node::NodeStorage &node_storage) const noexcept {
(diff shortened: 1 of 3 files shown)


src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -966,47 +966,76 @@ bool Literal::is_string_like() const noexcept {
...
+    },
+                                    other, node_storage);
 }
 
 Literal Literal::operator+(Literal const &other) const noexcept {
     return this->add(other);
 }
 
+Literal &Literal::operator+=(const Literal &other) noexcept {
+    auto result = *this + other;
```
liss-h: We also need `Literal &Literal::{add,sub,...}_assign(Literal const &other, NodeStorage &node_storage)` (like `Literal::{add,sub,...}`) so one can directly specify the node storage where the new literal is supposed to end up.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -966,47 +966,76 @@ bool Literal::is_string_like() const noexcept {
...
+    this->handle_ = result.handle_;
+    return *this;
+}
+
 Literal Literal::sub(Literal const &other, Node::NodeStorage &node_storage) const noexcept {
     return this->numeric_binop_impl([](auto const &num_ops) noexcept {
         return num_ops.sub_fptr;
-    }, other, node_storage);
+    },
+                                    other, node_storage);
```
liss-h: something strange happened to the formatting here (and some other places) 😅

liss-h (changes requested):
Thx, just one feature missing

src/rdf4cpp/rdf/Literal.cpp
mcb5637: that seems to be what clang format wants it to look like

liss-h: Looks like a bug to me, I would understand this:
```c++
return this->numeric_binop_impl([](auto const &num_ops) noexcept {
				    return num_ops.sub_fptr;
    			        },
    			        other, node_storage);

```

but the thing it chose just looks broken.

Could you please try formatting it like I did and see if clang-format still wants to break it?
And regardless of what it does could you please file a bug report against clang-format? Just to make them aware of it. I mean it could be on purpose but I kinda doubt that

mcb5637: seems like this is already reported: https://github.com/llvm/llvm-project/issues/58146
the suggested workaround formats it like this:
```cpp
    return this->numeric_binop_impl(
            [](auto const &num_ops) noexcept {
                return num_ops.add_fptr;
            },
            other, node_storage);
```
should i change them to the above or keep the current
```cpp
    // clang-format off
    return this->numeric_binop_impl([](auto const &num_ops) noexcept {
        return num_ops.add_fptr;
    }, other, node_storage);
    // clang-format on``` ?

liss-h: Please change it to the suggested workaround

liss-h (approved):

liss-h:
The coverage test is still failing but I don't think this has anything to do with your PR.

PR #160, opened 2023-03-08
alternative case conversion

### LINKED ISSUE #150: Bug: lower and uppercase unicode support

mcb5637:
closing #150 
using https://github.com/uni-algo/uni-algo instead of boost

### DIFF
--- a/.github/workflows/reusable_packaging_and_deployment.yml
+++ b/.github/workflows/reusable_packaging_and_deployment.yml
@@ -44,6 +44,7 @@ jobs:
           conan user
           conan profile new --detect default
           conan profile update settings.compiler.libcxx=libstdc++11 default
+          conan remote add dice-group https://conan.dice-research.org/artifactory/api/conan/tentris # TODO: remove when uni-algo/0.7.1 is available on Conan Center
 
       - name: Cache conan data
         id: cache-conan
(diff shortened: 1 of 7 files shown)


bigerl:
The build problem is when executing the test step of the conan package. That runs and builds what is in the test_package folder. 

There, the compiled files of uni-algo / una cannot be found because it is retrieved via FetchContent in the main CMakeLists.txt and not packaged by conan. 

You can reproduce this locally with:
```
conan create . -pr:b=default -pr:h=default --build missing
```

make sure you have a conan version <2.0:
```
pip3 install "conan<2.0"
```

To fix this, I would suggest you use either the single include https://github.com/uni-algo/uni-algo-single-include and make it appear in the private-include folder during cmake configure time or even better, the conan package https://github.com/uni-algo/uni-algo 

For the second option, please use for now the 0.7.0 version but make a PR to conan center with a version bump to 0.7.1. Use this fork https://github.com/dice-group/conan-center-index-fork/tree/master (do not work on master but create a branch) and read into the contribution guide lines. You have to sign something and post to some issue to be unlocked for their CI. Consider it a long-time project as it may take 1-3 weeks until the package is actually published from now.

bigerl:
I just saw that Conan uses the header-only version. Please double check first if the header-only version works. There seems to be a limitation with the header-only version regarding locale: https://github.com/uni-algo/uni-algo-single-include

I think we should update the uni-algo Conan package then to make it configurable if it should build a library or use the header-only version.

bigerl:
See https://github.com/rdf4cpp/rdf4cpp/issues/163

This can be merged before non-header-only uni-algo v0.7.1 is published on Conan Central.

liss-h (approved):
Looks good

PR #162, opened 2023-03-13
rdf:langString language tag inlining

### LINKED ISSUE #112: Future improvement: rdf:langString (lang tag) inlining
It could be possible to inline the language tag for some `rdf:langString`s by sacrificing some bits in
the LiteralID to encode a fixed set of lang tags and using only the remaining bits to refer to the backend.
This only works for sufficiently small backend ids.

mcb5637:
closes #112


src/rdf4cpp/rdf/datatypes/rdf/LangString.hpp
```diff
@@ -1,12 +1,66 @@
...
 namespace rdf4cpp::rdf::datatypes::registry {
 
+namespace lang_tags {
+enum class InlinedTags : uint8_t {
+    en,
+    de,
+    fr,
+    ch,
+};
+constexpr const size_t inlined_size = 2;
```
liss-h: `constexpr` implies `const`.
Additionally these should be `inline constexpr`, almost sure that these definitions are currently invalid (even if compilers accept them).

src/rdf4cpp/rdf/datatypes/rdf/LangString.hpp
```diff
@@ -20,14 +74,29 @@ struct DatatypeMapping<rdf_lang_string> {
 };
 
 template<>
-inline capabilities::Default<rdf_lang_string>::cpp_type capabilities::Default<rdf_lang_string>::from_string(std::string_view) {
+inline capabilities::Default<rdf_lang_string>::cpp_type capabilities::Default<rdf_lang_string>::from_string([[maybe_unused]] std::string_view) {
```
liss-h: unnecessary `[[maybe_unused]]`, unnamed parameters are always `[[maybe_unused]]` by default.
same applies for the following functions

src/rdf4cpp/rdf/datatypes/rdf/LangString.hpp
```diff
@@ -1,12 +1,66 @@
...
+    de,
+    fr,
+    ch,
+};
+constexpr const size_t inlined_size = 2;
+constexpr const uint64_t shift = storage::node::identifier::LiteralID::width - inlined_size;
+constexpr const uint64_t mask_inlined = ((1l << inlined_size) - 1) << shift;
+constexpr const uint64_t mask_id = (1l << shift) - 1;
+
+constexpr const char *from_inlined(InlinedTags t) noexcept {
```
liss-h: please rename either this function or the other overload, having them named the same is unecessarly confusing

src/rdf4cpp/rdf/datatypes/rdf/LangString.hpp
```diff
@@ -1,12 +1,66 @@
...
+            return "de";
+        case InlinedTags::fr:
+            return "fr";
+        case InlinedTags::ch:
+            return "ch";
+        default:
+            return "";
+    }
+}
+constexpr std::optional<InlinedTags> try_into_inlined(std::string_view t) noexcept {
```
liss-h: same here

src/rdf4cpp/rdf/datatypes/rdf/LangString.hpp
```diff
@@ -1,12 +1,66 @@
...
+    if (t == "de")
+        return InlinedTags::de;
+    if (t == "fr")
+        return InlinedTags::fr;
+    if (t == "ch")
+        return InlinedTags::ch;
+    return std::nullopt;
+}
+
+constexpr std::optional<uint64_t> try_into_inlined(uint64_t id, InlinedTags tag) noexcept {
```
liss-h: It would be helpful to be more specific with the types here. Using `LiteralID` here would be better, but apparently I also didn't do that for `Inlineable::(from|to)_inlined`. I don't quite remember why I didn't but can you try changing it?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -310,6 +347,10 @@ Literal Literal::as_simplified_lexical_form(NodeStorage &node_storage) const noe
 
 std::string_view Literal::language_tag() const noexcept {
     if (this->datatype_eq<datatypes::rdf::LangString>()) {
+        if (this->is_inlined()) {
+            auto id = rdf4cpp::rdf::datatypes::registry::lang_tags::from_inlined(this->handle_.node_id().literal_id().value);
```
liss-h: Not really important more fyi if you didn't know yet. C++ now (since 17) has [structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding), which can destructure some things into their elements. For example this line could be rewritten as

```c++
auto [tag, _] = rdf4cpp::rdf::datatypes::registry::lang_tags::from_inlined(this->handle_.node_id().literal_id().value);
```

liss-h (changes requested):
Thanks, baseline implementation looks like I imagined it would look like :). As always I have some things to nitpick.

The next step I guess would now be to implement a facility to specify the inlineable language tags (and their bit width) at runtime like @bigerl suggested.

Although I don't particularly like this idea, it probably makes the most sense to store this information in the registry. In some way it's kind of a weird place to put it but on the other hand I currently can't come up with a better alternative. Open to suggestions from your side, obviously.

Another question open for discussion would be what happens if you change it at runtime while you also have some literals floating around. I would say UB, as this is only really useful for very specialized use cases anyways so I think we can expect people to read the documentation.

src/rdf4cpp/rdf/datatypes/rdf/LangString.hpp
```diff
@@ -1,12 +1,125 @@
...
+#include <string_view>
+#include <vector>
+
 #include <rdf4cpp/rdf/datatypes/registry/DatatypeMapping.hpp>
 #include <rdf4cpp/rdf/datatypes/registry/LiteralDatatypeImpl.hpp>
 #include <rdf4cpp/rdf/storage/util/robin-hood-hashing/robin_hood_hash.hpp>
 
 namespace rdf4cpp::rdf::datatypes::registry {
 
+namespace lang_tags {
```
liss-h: Could you move this stuff to to the registry, i.e. `rdf4cpp::rdf::datatypes::registry::DatatypeRegistry` I feel like it's odd to have that floating around here.

src/rdf4cpp/rdf/datatypes/rdf/LangString.hpp
```diff
@@ -1,12 +1,125 @@
...
+/**
+ * language tags to try to inline.
+ * the vector can be modified at startup.
+ * modifying the vector after any Literal was created is undefined behavior.
+ * this includes Literals already created in a persistent node storage.
+ */
+#ifdef MODIFYABLE_LANG_TAG_INLINES
+inline const std::vector<std::string>
```
liss-h: After you move this to the registry, please also add a nice public facing interface that hides how this is implemented. For example something like

- `DatatypeRegistry::set_inlined_language_tags(std::initializer_list<string_something>)`
- `DatatypeRegistry::add_inlined_language_tag(string_something)`

the exact details are not that relevant, but you probably get the idea.
And a **BIG** warning of the UB cases above those functions would be good

src/rdf4cpp/rdf/datatypes/rdf/LangString.hpp
```diff
@@ -1,12 +1,125 @@
...
+        tags_to_inline{
+                "de",
+                "en",
+                "fr",
+                "ch",
+        };
+#else
+constexpr const std::array<std::string_view, 4> tags_to_inline{
```
liss-h: `constexpr` implies `const`

src/rdf4cpp/rdf/datatypes/xsd/integers/non_negative/PositiveInteger.cpp
```diff
@@ -48,18 +48,21 @@ nonstd::expected<capabilities::Default<xsd_positive_integer>::cpp_type, DynamicE
...
 template<>
-std::optional<uint64_t> capabilities::Inlineable<xsd_positive_integer>::try_into_inlined(cpp_type const &value) noexcept {
+std::optional<storage::node::identifier::LiteralID> capabilities::Inlineable<xsd_positive_integer>::try_into_inlined(cpp_type const &value) noexcept {
     auto const to_pack_value = value - 1;
     if (to_pack_value >= (uint64_t{1} << storage::node::identifier::LiteralID::width)) {
         return std::nullopt;
     }
 
-    return util::try_pack_integral<uint64_t, storage::node::identifier::LiteralID::width>(static_cast<uint64_t>(to_pack_value));
+    auto p = util::try_pack_integral<uint64_t, storage::node::identifier::LiteralID::width>(static_cast<uint64_t>(to_pack_value));
```
liss-h: - [x] Task for @Clueliss check if that can be done directly without optional fiddling

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -253,11 +276,15 @@ IRI Literal::datatype() const noexcept {
 
 util::CowString Literal::lexical_form() const noexcept {
     if (this->is_inlined()) {
+        if (this->datatype_eq<datatypes::rdf::LangString>()) {
+            return this->lang_tagged_get_de_inlined().lexical_form();
```
liss-h, line 296: Example of thing I talk about in the top level comment

src/rdf4cpp/rdf/datatypes/xsd/Double.cpp
```diff
@@ -46,23 +46,23 @@ nonstd::expected<capabilities::Numeric<xsd_double>::ceil_result_cpp_type, Dynami
 }
 
 template<>
-std::optional<uint64_t> capabilities::Inlineable<xsd_double>::try_into_inlined(cpp_type const &value) noexcept {
+std::optional<storage::node::identifier::LiteralID> capabilities::Inlineable<xsd_double>::try_into_inlined(cpp_type const &value) noexcept {
```
liss-h, line 49: Thx, for fixing this stuff

liss-h (changes requested):
So, a little interface cleanup left.

More minor probably:
I'm a little conflicted on the recursive calls on the deinlined Literal that happen in some places, I'm like 95% sure it probably doesn't matter at all but maybe it does? @bigerl please give an opinion

bigerl:
Please investigate and an summarize how http://www.w3.org/TR/xpath-functions/#func-compare works in SPARQL equal (around [here](https://www.w3.org/TR/sparql11-query/#op_lt)) and comparison if used. An important aspects seems to be https://www.w3.org/TR/xpath-functions/#dt-codepoint-collation

mcb5637:
I assume the last comment should be here instead #166

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -322,6 +323,150 @@ class DatatypeRegistry {
...
+        using LangTagID = uint64_t;
+
+        static constexpr size_t bits_needed_for(uint64_t i) {
+            return std::bit_width(i);
+        }
+
+        /**
+         * number of bits needed for the current tags to inline
+         */
+        static inline size_t inlined_size() noexcept {
```
liss-h: There are some functions here that I don't think should be publicly accessible, as they are kind of implementation details. This is one of them, but there are a few more below here. (everything related to the bit fiddling).

Also: it probably makes sense to mark all these `constexpr`, compiler probably treats them as such anyways but doesn't hurt. (also no need for `inline` then)

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -322,6 +323,150 @@ class DatatypeRegistry {
      * or there is no viable conversion, else the found conversion
      */
     [[nodiscard]] static std::optional<DatatypeConverter> get_common_type_conversion(DatatypeIDView lhs_type_id, DatatypeIDView rhs_type_id) noexcept;
+
+public:
+    class LangTagInlines {
```
liss-h, line 328: Like the idea with the nested class :)

liss-h (changes requested):
Just some minor things that don't really belong in the public interface.

And btw, fixed the packing so the workaround for the optional is not necessary anymore.

src/rdf4cpp/rdf/Literal.cpp
bigerl: look safe to me. It's not super maintainable because when messing up later only an stack overflow at runtime tell you that you created an infinite recursion. I would leave it like that unless anybody has a good idea at hand how to do it better.

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.hpp
mcb5637: there is currently a test case for `inlined_size()`, everything else is private now

liss-h (approved):
Looks good

liss-h:
Merging the other PRs created a conflict, please resolve @mcb5637

PR #166, opened 2023-03-27
unicode fixes

### LINKED ISSUE #161: Bug: Unicode support for more string functions
SPARQL functions that expect unicode support, but currently only support ASCII

- [x] literal ctor should normalize strings to NFC
- [x] then modify any comparisons to not do a normalization themselves

- [x] SUBSTR
- [x] STRSTARTS
- [x] STRENDS
- [x] CONTAINS
- [x] STRBEFORE
- [x] STRAFTER
- [x] make_lang_tagged (convert lang tag to lowercase)
- [x] make_typed_from_value (does currently not convert lang tag to lowercase, add)
- [x] operator == (will be fixed by ctor normalization)

to implememt, use https://github.com/uni-algo/uni-algo
also these functions need to be adapted to the new libraray:

- [x] STRLEN 
- [x] ENCODE_FOR_URI

after that, the old utf8 lib can be removed

mcb5637:
resolve #161

### DIFF
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -30,9 +30,10 @@ Literal Literal::make_null() noexcept {
 }
 
 Literal Literal::make_simple_unchecked(std::string_view lexical_form, Node::NodeStorage &node_storage) noexcept {
+    auto lf = una::norm::to_nfc_utf8(lexical_form);
     return Literal{NodeBackendHandle{node_storage.find_or_make_id(storage::node::view::LexicalFormLiteralBackendView{
                                              .datatype_id = storage::node::identifier::NodeID::xsd_string_iri.first,
-                                             .lexical_form = lexical_form,
+                                             .lexical_form = lf,
                                              .language_tag = ""}),
                                      storage::node::identifier::RDFNodeType::Literal,
                                      node_storage.id()}};
@@ -55,9 +56,10 @@ Literal Literal::make_noninlined_special_unchecked(std::any &&value, storage::no
 }
 
 Literal Literal::make_lang_tagged_unchecked(std::string_view lexical_form, std::string_view lang, Node::NodeStorage &node_storage) noexcept {
+    auto lf = una::norm::to_nfc_utf8(lexical_form);
     return Literal{NodeBackendHandle{node_storage.find_or_make_id(storage::node::view::LexicalFormLiteralBackendView{
                                              .datatype_id = storage::node::identifier::NodeID::rdf_langstring_iri.first,
-                                             .lexical_form = lexical_form,
+                                             .lexical_form = lf,
                                              .language_tag = lang}),
                                      storage::node::identifier::RDFNodeType::Literal,
                                      node_storage.id()}};
@@ -94,12 +96,13 @@ Literal Literal::make_typed_unchecked(std::any &&value, datatypes::registry::Dat
 }
 
 Literal Literal::make_string_like_copy_lang_tag(std::string_view str, Literal const &lang_tag_src, Node::NodeStorage &node_storage) noexcept {
+    auto s = una::norm::to_nfc_utf8(str);
     if (lang_tag_src.datatype_eq<datatypes::rdf::LangString>()) {
-        return Literal::make_lang_tagged_unchecked(str, lang_tag_src.language_tag(), node_storage);
+        return Literal::make_lang_tagged_unchecked(s, lang_tag_src.language_tag(), node_storage);
     }
 
     assert(lang_tag_src.datatype_eq<datatypes::xsd::String>());
-    return Literal::make_simple_unchecked(str, node_storage);
+    return Literal::make_simple_unchecked(s, node_storage);
 }
 
 bool Literal::dynamic_datatype_eq_impl(std::string_view datatype) const noexcept {
@@ -1321,9 +1324,8 @@ util::TriBool Literal::contains(std::string_view const needle) const noexcept {
 
     auto const s = this->lexical_form();
     auto norm_needle = needle | una::views::utf8 | una::views::norm::nfc | una::ranges::to_utf8<std::string>();
-    auto norm_this = s.view() | una::views::utf8 | una::views::norm::nfc | una::ranges::to_utf8<std::string>();
 
-    auto r = una::casesens::search_utf8(norm_this, norm_needle);
+    auto r = una::casesens::search_utf8(s.view(), norm_needle);
     return static_cast<bool>(r);
 }
 
@@ -1360,14 +1362,13 @@ Literal Literal::substr_before(std::string_view const needle, Node::NodeStorage 
 
     auto const s = this->lexical_form();
     auto const norm_needle = una::norm::to_nfc_utf8(needle);
-    auto const norm_this = una::norm::to_nfc_utf8(s.view());
 
-    const auto r = una::casesens::search_utf8(norm_this, norm_needle);
+    const auto r = una::casesens::search_utf8(s.view(), norm_needle);
 
     if (!r)
         return Literal::make_simple_unchecked("", node_storage);
 
-    auto substr = static_cast<std::string_view>(norm_this).substr(0, r.pos());  // search_utf8 returns byte position, not unicode character position
+    auto substr = static_cast<std::string_view>(s.view()).substr(0, r.pos());  // search_utf8 returns byte position, not unicode character position
     return Literal::make_string_like_copy_lang_tag(substr, *this, node_storage);
 }
 
@@ -1390,14 +1391,13 @@ Literal Literal::substr_after(std::string_view const needle, Node::NodeStorage &
 
     auto const s = this->lexical_form();
     auto const norm_needle = una::norm::to_nfc_utf8(needle);
-    auto const norm_this = una::norm::to_nfc_utf8(s.view());
 
-    const auto r = una::casesens::search_utf8(norm_this, norm_needle);
+    const auto r = una::casesens::search_utf8(s.view(), norm_needle);
 
     if (!r)
         return Literal::make_simple_unchecked("", node_storage);
 
-    auto substr = static_cast<std::string_view>(norm_this).substr(r.pos() + norm_needle.size());  // search_utf8 returns byte position, not unicode character position
+    auto substr = static_cast<std::string_view>(s.view()).substr(r.pos() + norm_needle.size());  // search_utf8 returns byte position, not unicode character position
     return Literal::make_string_like_copy_lang_tag(substr, *this, node_storage);
 }
 
@@ -1417,10 +1417,9 @@ util::TriBool Literal::str_starts_with(std::string_view const needle) const noex
     auto const s = this->lexical_form();
 
     auto norm_needle = needle | una::ranges::views::utf8 | una::views::norm::nfc;
-    auto norm_this = s.view() | una::ranges::views::utf8 | una::views::norm::nfc;
     auto len = std::ranges::distance(norm_needle);
     // TODO use std::ranges::starts_with as soon as c++ 23 arrives
-    return std::ranges::equal(norm_needle, norm_this | una::views::take(len));
+    return std::ranges::equal(norm_needle, s.view() | una::ranges::views::utf8 | una::views::take(len));
 }
 
 Literal Literal::as_str_starts_with(std::string_view const needle, Node::NodeStorage &node_storage) const noexcept {
@@ -1456,7 +1455,7 @@ util::TriBool Literal::str_ends_with(std::string_view const needle) const noexce
 
     auto const s = this->lexical_form();
     auto norm_needle = needle | una::views::utf8 | una::views::norm::nfc;
-    auto norm_this = s.view() | una::views::utf8 | una::views::norm::nfc;
+    auto norm_this = s.view() | una::views::utf8;
     auto const len_needle = std::ranges::distance(norm_needle);
     auto const len_this = std::ranges::distance(norm_this);
 
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -101,48 +101,52 @@ Literal Literal::make_string_like_copy_lang_tag(std::string_view str, Literal co
     return Literal::make_simple_unchecked(str, node_storage);
 }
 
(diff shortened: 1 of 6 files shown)


bigerl:
We could also store string unicode NFC in the first place. Please check the RDF specs if anything speaks against it. Would help with deduplication.

It would also make sense to normalize IRIs. It would involve unicode as well, I guess. But that would be a separate issue.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -54,9 +56,10 @@ Literal Literal::make_noninlined_special_unchecked(std::any &&value, storage::no
 }
 
 Literal Literal::make_lang_tagged_unchecked(std::string_view lexical_form, std::string_view lang, Node::NodeStorage &node_storage) noexcept {
+    auto lf = una::norm::to_nfc_utf8(lexical_form);
```
liss-h: You are normalizing twice in some cases, e.g. when calling `make_lang_tagged`. Don't normalize here, this function is not meant to do anything to the input data except putting it into storage.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -29,9 +30,10 @@ Literal Literal::make_null() noexcept {
 }
 
 Literal Literal::make_simple_unchecked(std::string_view lexical_form, Node::NodeStorage &node_storage) noexcept {
+    auto lf = una::norm::to_nfc_utf8(lexical_form);
```
liss-h: Same here, do not normalize in `*_unchecked` functions.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1319,13 +1317,25 @@ Literal Literal::regex_replace(Literal const &pattern, Literal const &replacemen
     return this->regex_replace(*repl, node_storage);
 }
 
-util::TriBool Literal::contains(std::string_view const needle) const noexcept {
+util::TriBool Literal::contains_normalized(std::string_view const needle) const noexcept {
```
liss-h: This is still an open question, just highlighting this for later
- [ ] figure out if we actually want to normalize the inputs

liss-h (changes requested):
We are still kind of waiting for a resolution for the normalization discussion, but I'll leave some small comments for now

And while I'm at it I'll just leave a list of the options we discussed here

Equality interpretations:
1. equality (and other related stuff) happens after both operands are normalized
2. equality (and other related stuff) happens without modifying the operands in any way

Related: do we want to normalize before putting strings to storage?
1. Yes
2. No, only check if the unicode is valid, normalized or not (=> also do not normalize on equality operator and similar)

Arguments for 1: (in no particular order)
- this is the way the unicode people intended it (which begs the question why there are multiple ways to represent the same character at all)
- theoretically some dedup (minor at best)
- unnormalized and normalized chars represent the same thing in the real world
- rdf recommends it
- consistency with the way we implement numeric normalization

Arguments for 2: (in no particular order)
- normalization changes user data (precedence that this doesn't matter is set by numeric types)
- performance
- code reduction (minor)
- precendence in other programming languages (and I would say SPARQL is kind of something simmilar)
- rdf _only_ recommends it which means we can just not do it

The wikidata query service is inconsistent (whichever triplestore that is) some queries normalize the strings others don't, weird.

Hopefully I got everything and didn't misrepresent anything

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -106,20 +106,32 @@ bool Literal::dynamic_datatype_eq_impl(std::string_view datatype) const noexcept
     return this->datatype().identifier() == datatype;
 }
 
-Literal Literal::make_simple(std::string_view lexical_form, Node::NodeStorage &node_storage) {
-    return Literal::make_simple_unchecked(lexical_form, node_storage);
+Literal Literal::make_simple(std::string_view lexical_form, Node::NodeStorage &node_storage, bool normalize) {
```
liss-h: Please don't use a boolean flag, I know alex and nikos like to do that way too much 😉 but its not very pleasant to read at the call site, because looking at the code you just read:
```c++
auto l = Literal::make_simple("spherical cow", some_node_storage, true);
```
and then you have to go to the declaration and check what that `true` even means. (IDE highlighting can help but thats really unreliable).

Would prefer if this would read:
```c++
auto l = Literal::make_simple("ferris", some_node_storage, FindGoodNameForEnum::Normalize); //option a
auto l_alt = Literal::make_simple_normalize("moo", some_node_storage); // option b
```

Maybe it also makes sense to reverse the parameters but I'm not sure which use case (custom node storage, normalization) is more likely, so I don't really have a strong opinion on that.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -106,20 +106,32 @@ bool Literal::dynamic_datatype_eq_impl(std::string_view datatype) const noexcept
...
+    if (normalize) {
+        const auto ln = una::norm::to_nfc_utf8(lexical_form);
+        return Literal::make_simple_unchecked(ln, node_storage);
+    } else {
+        if (una::norm::is_nfc_utf8(lexical_form))
```
liss-h: This is only supposed to check for _any_ valid utf8. It should not only accept nfc.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -106,20 +106,32 @@ bool Literal::dynamic_datatype_eq_impl(std::string_view datatype) const noexcept
...
+            return Literal::make_simple_unchecked(lexical_form, node_storage);
+        else
+            return Literal{};
```
liss-h: throw exception instead to make it consistent with behaviour of other malformed datatype inputs. (e.g. `xsd:integer`)

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -106,20 +106,32 @@ bool Literal::dynamic_datatype_eq_impl(std::string_view datatype) const noexcept
...
 }
 
-Literal Literal::make_lang_tagged(std::string_view lexical_form, std::string_view lang_tag, Node::NodeStorage &node_storage) {
-    std::string lowercase_lang_tag;
-    std::transform(lang_tag.begin(), lang_tag.end(), std::back_inserter(lowercase_lang_tag), [](char const ch) {
-        return static_cast<char>(std::tolower(ch));
-    });
-
-    return Literal::make_lang_tagged_unchecked(lexical_form, lowercase_lang_tag, node_storage);
+Literal Literal::make_lang_tagged(std::string_view lexical_form, std::string_view lang_tag, Node::NodeStorage &node_storage, bool normalize) {
```
liss-h: same thing with the boolean as above

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -385,17 +397,19 @@ Literal::operator std::string() const noexcept {
                     break;
                 }
                 [[likely]] default : {
-                    out << character;
-                    break;
-                }
+                        out << character;
```
liss-h: formatting got messed up

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -385,17 +397,19 @@ Literal::operator std::string() const noexcept {
...
+                    }
             }
         }
         out << "\"";
     };
 
     std::ostringstream oss;
 
-    if (this->is_inlined()) {
+    if (this->null()) {
```
liss-h, line 414: good idea

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1570,11 +1600,12 @@ Literal Literal::substr(size_t start, size_t len, Node::NodeStorage &node_storag
 
     auto const s = this->lexical_form();
 
-    if (start > s.size()) {
+    if (start > this->strlen()) {
```
liss-h: I don't think this check makes sense anymore with unicode support, as it requires iteration of the whole string. (hope the `una` views just return an empty view if the start index is too large)

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1658,6 +1689,10 @@ Literal Literal::sha512(NodeStorage &node_storage) const {
     return this->hash_with("SHA2-512", node_storage);
 }
 
+std::string Literal::normalize_unicode(std::string_view utf8) {
```
liss-h: I would say this should be a free function, as it doesn't really have anything to do with `Literal`

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -231,20 +238,22 @@ class Literal : public Node {
...
      * @param compatible_value instance for which the literal is created
      * @param node_storage NodeStorage used
+     * @param normalize normalize lexical_form to UTF-8 NFC instead of failing (only for xsd::string/rdf::langString)
      * @return literal instance representing compatible_value
      */
     template<datatypes::LiteralDatatype T>
     [[nodiscard]] static Literal make_typed_from_value(typename T::cpp_type const &compatible_value,
-                                                       NodeStorage &node_storage = NodeStorage::default_instance()) noexcept {
+                                                       NodeStorage &node_storage = NodeStorage::default_instance(),
+                                                       bool normalize = false) noexcept {
```
liss-h: same thing with the bool here

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1388,13 +1408,13 @@ Literal Literal::substr_after(std::string_view const needle, Node::NodeStorage &
...
 
-    if (pos == std::string_view::npos) {
+    const auto r = una::casesens::search_utf8(s.view(), needle);
+
+    if (!r)
         return Literal::make_simple_unchecked("", node_storage);
-    }
 
-    auto const substr = s.view().substr(pos + needle.size());
+    auto substr = static_cast<std::string_view>(s.view()).substr(r.pos() + needle.size());  // search_utf8 returns byte position, not unicode character position
```
liss-h: why the `static_cast` to `std::string_view`?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1570,11 +1600,12 @@ Literal Literal::substr(size_t start, size_t len, Node::NodeStorage &node_storag
 
     auto const s = this->lexical_form();
 
-    if (start > s.size()) {
+    if (start > this->strlen()) {
         return Literal::make_string_like_copy_lang_tag("", *this, node_storage);
     }
 
-    auto const substr = s.view().substr(start, len);
+    auto substr = s.view() | una::ranges::views::utf8 | una::views::drop(start) | una::views::take(len) | una::ranges::to_utf8<std::string>();
```
liss-h, line 1608: Can this somehow be converted to a `string_view` without an allocation? In theory it should work, not sure if the interface of `una` allows that tho.

liss-h (changes requested):
Mostly interface stuff, some minor implementation things.

src/rdf4cpp/rdf/Literal.cpp
mcb5637: to use `std::string_view::substr` instead of `std::string::substr`, which does not allocate any memory

liss-h: `s.view()` returns a `std::string`?

src/rdf4cpp/rdf/Literal.cpp
mcb5637: removed it, and added a test case just in case it does not

src/rdf4cpp/rdf/Literal.cpp
mcb5637: oh, you are correct
that s.view() was a std::string in one of the previous versions and i forgot to remove the cast

src/rdf4cpp/rdf/Literal.cpp
mcb5637: no

liss-h: Sad. Its fine then

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -165,33 +165,53 @@ class Literal : public Node {
...
      */
     [[nodiscard]] static Literal make_simple(std::string_view lexical_form, NodeStorage &node_storage = NodeStorage::default_instance());
 
+    /**
+     * Constructs a simple Literal from a lexical form. Datatype is `xsd:string`.
+     * normalizes lexical_form to UTF-8 NFC.
+     * @param lexical_form the lexical form
+     * @param node_storage optional custom node_storage used to store the literal
+     */
+    [[nodiscard]] static Literal make_simple_normalize(std::string_view lexical_form, NodeStorage &node_storage = NodeStorage::default_instance());
```
liss-h, line 178: What happens if you provide invalid Unicode here? Is it an exception or are the invalid chars replaced by the unicode invalid character char-thing?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -101,36 +101,52 @@ Literal Literal::make_string_like_copy_lang_tag(std::string_view str, Literal co
     return Literal::make_simple_unchecked(str, node_storage);
 }
 
-bool Literal::dynamic_datatype_eq_impl(std::string_view datatype) const noexcept {
-    assert(!this->is_fixed());
-    return this->datatype().identifier() == datatype;
-}
+    bool Literal::dynamic_datatype_eq_impl(std::string_view datatype) const noexcept {
```
liss-h: Formatting still broken

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1658,6 +1690,10 @@ Literal Literal::sha512(NodeStorage &node_storage) const {
     return this->hash_with("SHA2-512", node_storage);
 }
 
+    std::string normalize_unicode(std::string_view utf8) {
```
liss-h: another formatting bug

liss-h (changes requested):
Formatting still borked 😅

src/rdf4cpp/rdf/Literal.hpp
mcb5637: replaced by replacement character codepoints (thats what the unicode standard says should happen) (and yes, there are more of them, but they basically all have the same purpose)

liss-h: makes sense, thx for clarifying

liss-h (approved):

PR #173, opened 2023-04-11
more file formats for IStreamQuadIterator

### LINKED ISSUE #164: Allow parsing Trig and NQuads
Currently, the parser is only able to parse Turtle (and its subset NTriple). The backend library, serd, is also able to Trig and NQuads. There should be an option to support those formats.

mcb5637:
resolve #164

### DIFF
--- a/private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.hpp
+++ b/private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.hpp
@@ -57,6 +57,18 @@ private:
     static SerdStatus on_prefix(void *voided_self, SerdNode const *name, SerdNode const *uri) noexcept;
     static SerdStatus on_stmt(void *voided_self, SerdStatementFlags, SerdNode const *graph, SerdNode const *subj, SerdNode const *pred, SerdNode const *obj, SerdNode const *obj_datatype, SerdNode const *obj_lang) noexcept;
 
+    static constexpr SerdSyntax extract_syntax_from_flags(ParsingFlags flags) noexcept {
+        if (flags.contains(ParsingFlag::Turtle))
+            return SerdSyntax::SERD_TURTLE;
+        if (flags.contains(ParsingFlag::NTriples))
+            return SerdSyntax::SERD_NTRIPLES;
+        if (flags.contains(ParsingFlag::NQuads))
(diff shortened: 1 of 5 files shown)


src/rdf4cpp/rdf/parser/ParsingFlags.hpp
```diff
@@ -9,6 +9,10 @@ namespace rdf4cpp::rdf::parser {
 enum struct ParsingFlag : uint8_t {
     Strict = 1 << 0,
     NoParsePrefix = 1 << 1,
+    Turtle = 1 << 2,
```
liss-h: The flags could be encoded into the 2 bits after the `NoParsePrefix` bit. Behaviour would be essentially identical to how it is right now except that it saves some space (maybe relevant, idk yet) and it makes more sense semantically. Just encode turtle as `0b00` which automatically makes it the default value.

private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.hpp
```diff
@@ -57,6 +57,18 @@ struct IStreamQuadIterator::Impl {
     static SerdStatus on_prefix(void *voided_self, SerdNode const *name, SerdNode const *uri) noexcept;
     static SerdStatus on_stmt(void *voided_self, SerdStatementFlags, SerdNode const *graph, SerdNode const *subj, SerdNode const *pred, SerdNode const *obj, SerdNode const *obj_datatype, SerdNode const *obj_lang) noexcept;
 
+    static constexpr SerdSyntax extract_syntax_from_flags(ParsingFlags flags) noexcept {
+        if (flags.contains(ParsingFlag::Turtle))
```
liss-h: If you implement the thing from below, just take care to check the things in the correct order. Or alternatively make this into a switch on the flag's underlying repr with the non-syntax bits masked out (not that important tbh, you choose which of the options you like better).

liss-h (changes requested):
Looks straightforward, just two comments related to the flags.
Btw, did you find out what the bug related to the quads actually was or was it just fixed by merging from serd's main repo?

mcb5637:
was a bug in serd, link to the issue: https://github.com/drobilla/serd/issues/42

liss-h (approved):

PR #175, opened 2023-04-24
java style BigDecimal

### LINKED ISSUE #172: Bug: xsd:decimal has invalid internal representation
Currently `xsd:decimal` is internally represented as `boost::multiprecision::dec_float` (a decimal arbitrary precision decimal **floating** point value).

This is invalid because it results in precision errors in math operations. Consider the following code:
```c++
auto sum = 1_xsd_int + 2_xsd_int + 3_xsd_int;
auto avg = sum / 3_xsd_int;

CHECK(avg == "2.0"_xsd_decimal); // fails with expansion: 2.0 == 2.0
```

The fix is to implement `xsd:decimal`'s internal represenstation like java's `BigDecimal`, i.e. an arbitrary precision integer unscaled value (`boost::multiprecision::cpp_int`) with a `uint32_t` scale. Such that `value = unscaled_value * 10^(-scale)`.

This is also what the spec implies: https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#dt-decimal.

For some information on the implementation details (e.g. how to choose the scale) the java `BigDecimal` docs should be helpful: https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html.

mcb5637:
resolves #172

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/xsd/Decimal.cpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/Decimal.cpp
@@ -25,25 +25,12 @@ capabilities::Default<xsd_decimal>::cpp_type capabilities::Default<xsd_decimal>:
 
 template<>
 std::string capabilities::Default<xsd_decimal>::to_canonical_string(cpp_type const &value) noexcept {
-    auto s = value.str(std::numeric_limits<cpp_type>::digits10, std::ios_base::fixed | std::ios_base::showpoint);
-    auto const non_zero_pos = s.find_last_not_of('0');
-
-    // cannot be npos because, showpoint is set and '.' != '0'
-    assert(non_zero_pos != std::string::npos);
-
-    // dot was found char implies there is a char after the dot (because precision == digits10 > 0) and char after dot must be zero
-    assert(s[non_zero_pos] != '.' || (s.size() >= non_zero_pos + 2 && s[non_zero_pos + 1] == '0'));
-
-    // +1 for pos -> size conversion
-    // maybe +1 to include one zero after dot
-    s.resize(non_zero_pos + 1 + static_cast<std::string::size_type>(s[non_zero_pos] == '.'));
-
-    return s;
+    return static_cast<std::string>(value);
 }
 
 template<>
 std::string capabilities::Default<xsd_decimal>::to_simplified_string(cpp_type const &value) noexcept {
-    return value.str();
+    return static_cast<std::string>(value);
 }
 
 template<>
@@ -51,12 +38,11 @@ nonstd::expected<capabilities::Numeric<xsd_decimal>::add_result_cpp_type, Dynami
     // https://www.w3.org/TR/xpath-functions/#op.numeric
     // decimal needs overflow protection
 
-    auto res = lhs + rhs;
-    if (isinf(res)) [[unlikely]] {
+    try {
+        return lhs + rhs;
+    } catch (const std::overflow_error &) {
         return nonstd::make_unexpected(DynamicError::OverOrUnderFlow);
     }
-
-    return res;
 }
 
 template<>
@@ -64,12 +50,11 @@ nonstd::expected<capabilities::Numeric<xsd_decimal>::sub_result_cpp_type, Dynami
     // https://www.w3.org/TR/xpath-functions/#op.numeric
     // decimal needs overflow protection
 
-    auto res = lhs - rhs;
-    if (isinf(res)) [[unlikely]] {
+    try {
+        return lhs - rhs;
+    } catch (const std::overflow_error &) {
         return nonstd::make_unexpected(DynamicError::OverOrUnderFlow);
     }
-
-    return res;
 }
 
 template<>
@@ -81,12 +66,11 @@ nonstd::expected<capabilities::Numeric<xsd_decimal>::div_result_cpp_type, Dynami
         return nonstd::make_unexpected(DynamicError::DivideByZero);
     }
 
-    auto res = lhs / rhs;
-    if (isinf(res)) [[unlikely]] {
+    try {
+        return lhs / rhs;
+    } catch (const std::overflow_error &) {
         return nonstd::make_unexpected(DynamicError::OverOrUnderFlow);
     }
-
-    return res;
 }
 
 template<>
@@ -94,48 +78,41 @@ nonstd::expected<capabilities::Numeric<xsd_decimal>::mul_result_cpp_type, Dynami
     // https://www.w3.org/TR/xpath-functions/#op.numeric
     // decimal needs overflow protection
 
-    auto res = lhs * rhs;
-    if (isinf(res)) [[unlikely]] {
+    try {
+        return lhs * rhs;
+    } catch (const std::overflow_error &) {
         return nonstd::make_unexpected(DynamicError::OverOrUnderFlow);
     }
-
-    return res;
 }
 
 template<>
 nonstd::expected<capabilities::Numeric<xsd_decimal>::abs_result_cpp_type, DynamicError> capabilities::Numeric<xsd_decimal>::abs(cpp_type const &operand) noexcept {
-    return boost::multiprecision::abs(operand);
+    return operand.abs();
 }
 
 template<>
 nonstd::expected<capabilities::Numeric<xsd_decimal>::round_result_cpp_type, DynamicError> capabilities::Numeric<xsd_decimal>::round(cpp_type const &operand) noexcept {
-    return boost::multiprecision::round(operand);
+    return operand.round(rdf4cpp::rdf::util::RoundingMode::Round);
 }
 
 template<>
 nonstd::expected<capabilities::Numeric<xsd_decimal>::floor_result_cpp_type, DynamicError> capabilities::Numeric<xsd_decimal>::floor(cpp_type const &operand) noexcept {
-    return boost::multiprecision::floor(operand);
+    return operand.round(rdf4cpp::rdf::util::RoundingMode::Floor);
 }
 
 template<>
 nonstd::expected<capabilities::Numeric<xsd_decimal>::ceil_result_cpp_type, DynamicError> capabilities::Numeric<xsd_decimal>::ceil(cpp_type const &operand) noexcept {
-    return boost::multiprecision::ceil(operand);
+    return operand.round(rdf4cpp::rdf::util::RoundingMode::Ceil);
 }
 
 template<>
 bool capabilities::Logical<xsd_decimal>::effective_boolean_value(cpp_type const &value) noexcept {
-    return !isnan(value) && value != 0.0;
+    return value != 0;
 }
 
 template<>
 std::partial_ordering capabilities::Comparable<xsd_decimal>::compare(cpp_type const &lhs, cpp_type const &rhs) noexcept {
-    if (lhs < rhs) {
-        return std::partial_ordering::less;
-    } else if (rhs < lhs) {
-        return std::partial_ordering::greater;
-    } else {
-        return std::partial_ordering::equivalent;
-    }
+    return lhs <=> rhs;
 }
 
 template struct LiteralDatatypeImpl<xsd_decimal,
--- a/src/rdf4cpp/rdf/datatypes/xsd/Decimal.cpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/Decimal.cpp
@@ -37,12 +37,11 @@ template<>
 nonstd::expected<capabilities::Numeric<xsd_decimal>::add_result_cpp_type, DynamicError> capabilities::Numeric<xsd_decimal>::add(cpp_type const &lhs, cpp_type const &rhs) noexcept {
     // https://www.w3.org/TR/xpath-functions/#op.numeric
     // decimal needs overflow protection
-
-    try {
-        return lhs + rhs;
-    } catch (const std::overflow_error &) {
+    auto r = lhs.add_checked(rhs);
+    if (r.has_value())
+        return r.value();
+    else
         return nonstd::make_unexpected(DynamicError::OverOrUnderFlow);
-    }
 }
 
 template<>
@@ -50,11 +49,11 @@ nonstd::expected<capabilities::Numeric<xsd_decimal>::sub_result_cpp_type, Dynami
     // https://www.w3.org/TR/xpath-functions/#op.numeric
     // decimal needs overflow protection
 
-    try {
-        return lhs - rhs;
-    } catch (const std::overflow_error &) {
+    auto r = lhs.sub_checked(rhs);
+    if (r.has_value())
+        return r.value();
+    else
         return nonstd::make_unexpected(DynamicError::OverOrUnderFlow);
-    }
 }
 
 template<>
@@ -78,11 +77,11 @@ nonstd::expected<capabilities::Numeric<xsd_decimal>::mul_result_cpp_type, Dynami
     // https://www.w3.org/TR/xpath-functions/#op.numeric
     // decimal needs overflow protection
 
-    try {
-        return lhs * rhs;
-    } catch (const std::overflow_error &) {
+    auto r = lhs.mul_checked(rhs);
+    if (r.has_value())
+        return r.value();
+    else
         return nonstd::make_unexpected(DynamicError::OverOrUnderFlow);
-    }
 }
 
 template<>
--- a/src/rdf4cpp/rdf/datatypes/xsd/Decimal.cpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/Decimal.cpp
@@ -65,11 +65,11 @@ nonstd::expected<capabilities::Numeric<xsd_decimal>::div_result_cpp_type, Dynami
         return nonstd::make_unexpected(DynamicError::DivideByZero);
     }
 
-    try {
-        return lhs / rhs;
-    } catch (const std::overflow_error &) {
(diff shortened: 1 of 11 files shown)


src/rdf4cpp/rdf/datatypes/xsd/Decimal.cpp
```diff
@@ -25,51 +25,36 @@ capabilities::Default<xsd_decimal>::cpp_type capabilities::Default<xsd_decimal>:
...
 
 template<>
 nonstd::expected<capabilities::Numeric<xsd_decimal>::add_result_cpp_type, DynamicError> capabilities::Numeric<xsd_decimal>::add(cpp_type const &lhs, cpp_type const &rhs) noexcept {
     // https://www.w3.org/TR/xpath-functions/#op.numeric
     // decimal needs overflow protection
 
-    auto res = lhs + rhs;
-    if (isinf(res)) [[unlikely]] {
+    try {
+        return lhs + rhs;
```
liss-h: Please add `BigDecimal::checked_(add|sub|div|...)` instead of wrapping this into a try catch.
A more general thing and related, I don't think any math ops should be able to throw. Do something different by default, probably UB.

So basically for each op there should be the default one (that is the overloaded operator) which does the fast thing (UB) and `checked_op` which returns `std::optional` or `expected`.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,662 @@
...
+template<typename T>
+concept BigDecimalBaseType = (std::unsigned_integral<T> && sizeof(T) >= sizeof(int32_t)) || std::same_as<T, boost::multiprecision::cpp_int>;
+
+template<BigDecimalBaseType UnscaledValue_t = boost::multiprecision::cpp_int, BigDecimalBaseType Exponent_t = uint32_t>
+class BigDecimal {
+    UnscaledValue_t unscaled_value;
+    Exponent_t exponent;
+    Sign sign = Sign::Positive;
+
+    static_assert(!std::is_integral_v<boost::multiprecision::cpp_int>);
```
liss-h: This can't be false since you are not allowed to specialize `is_integral`. Is this an artifact of an earlier version?

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,662 @@
...
+     * pow(base, ex) * v
+     */
+    static constexpr UnscaledValue_t shift_pow(UnscaledValue_t v, Exponent_t ex) {
+        for (Exponent_t i = 0; i < ex; ++i)
+            v = save_mul(v, UnscaledValue_t{base}, "shift_pow precision overflow");
+        return v;
+    }
+
+    template<class T>
+    static constexpr auto save_pow(const T &a, unsigned int b, const char *exc) {
```
liss-h: The `save_*` functions all have a typo, it should be `safe_*`.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,662 @@
...
+    /**
+     * conversion to a double
+     * @return
+     */
+    constexpr explicit operator double() const noexcept {
+        double v = static_cast<double>(unscaled_value) * std::pow(static_cast<double>(base), -static_cast<double>(exponent));
+        v = sign == Sign::Negative ? -v : v;
+        if (!std::isnan(v) && !std::isinf(v))
+            return v;
+        // even Javas BigDecimal has no better solution
```
liss-h, line 725: Lol. There is maybe some fancy thing you can do here, but this would be something for the people who are actually able to implement a proper floating point from/to string conversion algorithm. Which is not me, and we don't use that function anyways so: its fine to leave it like this.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,662 @@
...
+#include <functional>
+#include <sstream>
+#include <string>
+#include <string_view>
+
+#include <boost/multiprecision/cpp_int.hpp>
+
+namespace rdf4cpp::rdf::util {
+enum class RoundingMode {
+    ThrowInstead,
```
liss-h: Just remove this option, it prevents functions from being `noexcept`.
Instead of having this kind of an option, please add a differently named function that returns `expected`

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,662 @@
...
+            return boost::multiprecision::pow(a, b);
+        }
+    }
+
+    template<class T>
+    static constexpr auto save_add(const T &a, const T &b, const char *exc) {
+        if constexpr (std::is_integral_v<T>) {
+            T result;
+            if (__builtin_add_overflow(a, b, &result))
+                throw std::overflow_error{exc};
```
liss-h: In the general theme of "removing exceptions" this should return `std::optional<>` and be `noexcept`. Same for functions below.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,662 @@
...
+    /**
+     * converts a double to a BigDecimal.
+     * this conversion might not be exact, due to the built in limitations of doubles.
+     * if you have the possibility, use one of the other constructors.
+     * @param value
+     * @throw std::overflow_error on exceeding the types numeric limits
+     */
+    explicit BigDecimal(double value) : unscaled_value(0), exponent(0) {
+        if (std::isinf(value) || std::isnan(value))
+            throw std::invalid_argument{"value is NaN or infinity"};
```
liss-h, line 212: The exception here is fine btw.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,662 @@
...
+                return BigDecimal{v, e, sign};
+            case RoundingMode::Ceil:
+                return BigDecimal{v + 1, e, sign};
+            case RoundingMode::Round:
+                if (rem >= 5)
+                    return BigDecimal{v + 1, e, sign};
+                else
+                    return BigDecimal{v, e, sign};
+            default:
+                throw std::logic_error{"invalid rounding mode"};
```
liss-h:
```c++
assert(false);
__builtin_unreachable();
```

and make function `noexcept`

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,662 @@
...
+
+template<class UnscaledValue_t, class Exponent_t>
+std::string to_string(const BigDecimal<UnscaledValue_t, Exponent_t> &r) noexcept {
+    return static_cast<std::string>(r);
+}
+
+template<class UnscaledValue_t, class Exponent_t>
+BigDecimal<UnscaledValue_t, Exponent_t> pow(const BigDecimal<UnscaledValue_t, Exponent_t> &r, int n) {
+    return r.pow(n);
+}
```
liss-h, line 819: `abs(BigDecimal)` and  `round` are missing

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,662 @@
...
+        if (exponent == 0)
+            return unscaled_value;
+        return unscaled_value / shift_pow(1, exponent);
+    }
+
+    /**
+     * conversion to a string
+     * @return
+     */
+    explicit operator std::string() const noexcept {
```
liss-h: Note to future self: look at this later

liss-h (changes requested):
Hi, so this is a more general design review. I didn't look into the algorithms at all yet.

The TLDR is: please remove (almost) everything exceptiony (throwing in ctors is fine) and spam a bunch of `noexcept`s everywhere.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,859 @@
...
+};
+
+template<typename T>
+concept BigDecimalBaseType = (std::unsigned_integral<T> && sizeof(T) >= sizeof(int32_t)) || std::same_as<T, boost::multiprecision::cpp_int>;
+
+template<BigDecimalBaseType UnscaledValue_t = boost::multiprecision::cpp_int, BigDecimalBaseType Exponent_t = uint32_t>
+class BigDecimal {
+    UnscaledValue_t unscaled_value;
+    Exponent_t exponent;
+    Sign sign = Sign::Positive;
```
liss-h: Is there a particular reason that you are encoding the sign seperately? `boost::mp::cpp_int` is already signed.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,859 @@
...
+    return static_cast<Sign>(static_cast<T>(!static_cast<T>(a)));
+}
+
+enum class DecimalError {
+    Overflow,
+    NotDefined,  // aka NotANumber
+};
+
+template<typename T>
+concept BigDecimalBaseType = (std::unsigned_integral<T> && sizeof(T) >= sizeof(int32_t)) || std::same_as<T, boost::multiprecision::cpp_int>;
```
liss-h: I would argue this concept is kind of unecessary. Right now it only serves to arbitrarily limit the types one can use.

A more useful set of constraints (two if base and unscaled value have different requirements) would be ones that actually check for the required operations. Otherwise I would just remove them.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,859 @@
...
+     * this conversion might not be exact, due to the built in limitations of doubles.
+     * if you have the possibility, use one of the other constructors.
+     * @param value
+     * @throw std::overflow_error on exceeding the types numeric limits
+     */
+    explicit BigDecimal(double value) : unscaled_value(0), exponent(0) {
+        if (std::isinf(value) || std::isnan(value))
+            throw std::invalid_argument{"value is NaN or infinity"};
+        // this might fail on anything that is not x86-32/64
+        uint64_t v = *reinterpret_cast<uint64_t *>(&value);
```
liss-h: So I had a look at the rules around this again. And you were right, this is technically UB because it violates the strict aliasing rule. Now, I couldn't get `clang` or `gcc` to behave weirdly or even trigger UBSan, but it is what it is.

Luckily we have C++20 and therefore `std::bit_cast` so you can use that instead, it was created for exactly these kinds of things.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,859 @@
...
+    constexpr explicit BigDecimal(float value) : BigDecimal(static_cast<double>(value)) {}
+
+    /**
+     * converts a double to a BigDecimal.
+     * this conversion might not be exact, due to the built in limitations of doubles.
+     * if you have the possibility, use one of the other constructors.
+     * @param value
+     * @throw std::overflow_error on exceeding the types numeric limits
+     */
+    explicit BigDecimal(double value) : unscaled_value(0), exponent(0) {
```
liss-h, line 209: You can add `static_assert(std::endian::native == std::endian::little, "some error text")` here this ensures we are at least on little endian, not quite that we are on x86 but good enough.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,859 @@
...
+        if (std::isinf(value) || std::isnan(value))
+            throw std::invalid_argument{"value is NaN or infinity"};
+        // this might fail on anything that is not x86-32/64
+        uint64_t v = *reinterpret_cast<uint64_t *>(&value);
+        sign = (v >> 63) == 0 ? Sign::Positive : Sign::Negative;
+        auto ex = static_cast<int>((v >> 52) & 0x7ffL);
+        uint64_t significand = ex == 0
+                                       ? (v & ((1L << 52) - 1)) << 1
+                                       : (v & ((1L << 52) - 1)) | (1L << 52);
+        ex -= 1075;
```
liss-h, line 226: An ascii art drawing of the bit repr of a `double` would be helpful here to understand this.
Something like this: https://github.com/rdf4cpp/rdf4cpp/blob/c1036f15aa515e70cc2657c4dae8d7cb3f3d0689/src/rdf4cpp/rdf/datatypes/xsd/Base64Binary.cpp#L33-L49

Or alternatively a link to a website showing the layout.

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,859 @@
...
+                if (ex == std::numeric_limits<Exponent_t>::max()) {
+                    if (mul_checked<m>(rem, UnscaledValue_t{base}, rem))
+                        return true;
+                    result = handle_rounding(res, ex, rem / div, s, mode);
+                    return false;
+                }
+            }
+            ++ex;
+            if (mul_checked<m>(res, UnscaledValue_t{base}, res))
+                return true;  // TODO check if we can round here instead
```
liss-h: Forgot one todo here

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,859 @@
...
+                else
+                    return BigDecimal{v, e, sign};
+            default:
+                assert(false);
+                __builtin_unreachable();
+        }
+    }
+
+    template<OverflowMode m>
+    constexpr bool div(const BigDecimal &other, Exponent_t max_scale_increase, RoundingMode mode, BigDecimal &result) const noexcept {
```
liss-h, line 331: If these algorithms are based on the java impl, please add a link to the reference implementation to the functions

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,859 @@
...
+        str << s;
+        return str;
+    }
+
+    /**
+     * combined hash of a BigDecimals components hashes.
+     * @return
+     */
+    [[nodiscard]] size_t hash() const {
+        return (sign == Sign::Negative ? 1 : 0) ^ std::hash<UnscaledValue_t>{}(unscaled_value) ^ std::hash<Exponent_t>{}(exponent);
```
liss-h: We have `robin_hood` in `storage/util/robin-hood-hashing` so you can use:

```c++
return robin_hood::hash<std::array<size_t, 3>>{}(std::array<size_t, 3>{
    robin_hood::hash<Sign>{}(sign),
    robin_hood::hash<UnscaledValue_t>{}(unscaled_value),
    robin_hood::hash<Exponent_t>{}(exponent)});
```

Probably gives better results than xoring (haven't personally tested that theory but we do that in a bunch of places).

src/rdf4cpp/rdf/util/BigDecimal.hpp
```diff
@@ -0,0 +1,859 @@
...
+            if (!hasDot && ex == 0) {
+                if (s.view().empty())
+                    s << '0';
+                s << '.';
+                hasDot = true;
+            } else {
+                --ex;
+            }
+            auto c = static_cast<uint32_t>(v % base);
+            if (!(!hasDot && c == 0 && s.view().empty()))  // skip trailing 0s
```
liss-h: could use de-morgan simplification

liss-h (changes requested):
Mostly future maintanance related things

src/rdf4cpp/rdf/util/BigDecimal.hpp
mcb5637: to use unsigned integers in case of not using cpp_int

liss-h: Why can't you just use signed ones?

mcb5637: To avoid the asymetry between positive and negative numbers.

src/rdf4cpp/rdf/util/BigDecimal.hpp
mcb5637: The intention behind this was to define a minimum size for the components. Relaxed it to anything that looks numeric and is not a float/double.

src/rdf4cpp/rdf/util/BigDecimal.hpp
liss-h: Is avoiding that required for anything?

mcb5637: It makes arithmetic easier.

bigerl: It might be a more effort but for the sake of compactness we should use a signed type for `unscaled_value`. Instead of using the sign you can check for (v > 0). To squeeze out the full performance potential, you could have a template function `Sign get_sign(UnscaledValue_t& const)` that defaults to `{ return x > 0; }` but can be overwritten to use a type specific function, i.e. `{ if x.positive() return Sign::Positive; else return Sign::Negative; }`

mcb5637: Removed the Sign enum completely and changed the UnscaledValue_t to signed.
Drawback: Subtraction is now its own function and no longer just an add with unary minus applied to one of the operands. And unary minus/abs can now theoretically overflow. (Both because `-std::numeric_limits<int>::min()` is undefined)

liss-h (approved):
Thanks 🙂

PR #177, opened 2023-05-17
Fix: namespace in version header

liss-h:
Remember that thing in in the tentris PR where I needed to always write `::rdf4cpp` (absolute namspace), that was caused by this namespace being wrong.


nkaralis:
Nice catch. Feel free to merge it.

PR #179, opened 2023-05-22
xsd Date and Time types

### LINKED ISSUE #53: Date Type Implemenation
- [x] date
- [x] dateTime
  - [x] dateTimeStamp
- [x] gDay
- [x] gMonth
- [x] gMonthDay
- [x] gYear
- [x] gYearMonth
- [x] time

Should be implemented using ~~https://github.com/HowardHinnant/date~~ `<chrono>` from `libstdc++13`.

mcb5637:
close #53 
date/time arithmetic will be another PR


mcb5637:
@Clueliss please rewiew parsing and to_string of date/time types

src/rdf4cpp/rdf/datatypes/xsd/time/DateTimeStamp.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+extern template struct LiteralDatatypeImpl<xsd_dateTimeStamp,
+                                           capabilities::Comparable,
+                                           capabilities::FixedId>;
+
+}  // namespace rdf4cpp::rdf::datatypes::registry
+
+namespace rdf4cpp::rdf::datatypes::xsd {
+
+struct DateTimeStamp : registry::LiteralDatatypeImpl<registry::xsd_dateTimeStamp,
```
liss-h, line 58: Note for later: check if it makes sense to make this a subtype of `xsd:dateTime`. Regardless of whether xpath says so or not.

I would imagine being able to cast between them is helpful since they are so similar.

src/rdf4cpp/rdf/datatypes/xsd/time/Timezone.hpp
```diff
@@ -0,0 +1,211 @@
...
+        for (size_t i = milli_s.length(); i < 3; ++i) {
+            ms *= 10;
+        }
+        s = s.substr(0, p);
+    }
+    std::chrono::seconds sec{util::from_chars<unsigned int>(s)};
+    return sec + ms;
+}
+
+class TimeComparer {
```
liss-h: as it is right now, I think this could be a namespace instead.

src/rdf4cpp/rdf/datatypes/xsd/time/Timezone.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+    inline static auto deconstruct(Timepoint a) {
+        auto days = std::chrono::time_point_cast<std::chrono::days>(a);
+        return std::make_pair(std::chrono::year_month_day{days}, std::chrono::hh_mm_ss{a - days});
+    }
+
+    inline static std::partial_ordering piecewise_compare(Timepoint a, Timepoint b) {
+        auto av = deconstruct(a);
+        auto bv = deconstruct(b);
+        std::partial_ordering cmp = av.first.year() <=> bv.first.year();
```
liss-h: Does the auto synthesised spaceship not do the right thing?

src/rdf4cpp/rdf/datatypes/xsd/time/DateTimeStamp.cpp
```diff
@@ -0,0 +1,44 @@
+#include <rdf4cpp/rdf/datatypes/xsd/time/DateTimeStamp.hpp>
+
+#include <rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp>
+
+namespace rdf4cpp::rdf::datatypes::registry {
+
+template<>
+capabilities::Default<xsd_dateTimeStamp>::cpp_type capabilities::Default<xsd_dateTimeStamp>::from_string(std::string_view s) {
+    auto year = parse_date_time_fragment<std::chrono::year, int, '-'>(s);
```
liss-h, line 10: I would say the parsing in general looks very reasonable and I would have probably done it this way as well.

Maybe for the tests try add some weird manually constructed edge cases. I wouldn't expect anything to come up but never hurts to at least try to break it.

Another Idea would be to grep all the triples with date times out of swdf or something, try to parse it using `IStreamQuadIterator` and see if something comes up.

liss-h (changes requested):
On first look the parsing looks fine to me and I would have probably done it very similarly 👍

Have some unrelated comments, but nothing very serious as of right now.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1734,6 +1734,154 @@ Literal Literal::sha512(NodeStorage &node_storage) const {
...
 
+Literal Literal::now(NodeStorage &node_storage) {
+    auto n = std::chrono::floor<std::chrono::milliseconds>(std::chrono::system_clock::now());
+    datatypes::registry::Timezone tz{};
+    datatypes::registry::TimePoint t = tz.to_local(n);
+    datatypes::registry::OptionalTimezone opt = std::nullopt;
+    return make_typed_from_value<datatypes::xsd::DateTime>(std::make_pair(t, opt), node_storage);
+}
+
+Literal Literal::year(NodeStorage &node_storage) const {
```
liss-h: These functions should all be called `as_*`, i.e. `as_year`. And there should be an additional set of functions called the way these are called which return an int (uint?) directly.

Also: noexcept

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -411,6 +411,105 @@ class Literal : public Node {
...
     }
 
+    /**
+     * Tries to cast this literal to a literal of the given type and return the result without creating a Literal.
+     *
+     * @return conversion result
+     */
+    template<datatypes::LiteralDatatype T>
+    requires (!std::same_as<T, datatypes::xsd::String>)
+    std::optional<typename T::cpp_type> cast_to_value() const noexcept {
```
liss-h, line 432: ~~Why does this exist?~~

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -997,6 +1096,61 @@ class Literal : public Node {
      */
     [[nodiscard]] Literal sha512(NodeStorage &node_storage = NodeStorage::default_instance()) const;
 
+    /**
+     * returns the current time.
+     * Note: will need to be buffered for each query, because each query has only one now.
```
liss-h, line 1112: Not sure what this comment means.

liss-h (changes requested):
Partial review: frontend interface

src/rdf4cpp/rdf/Literal.hpp
bigerl: I think what's meant here is that: 
```SPARQL
SELECT ?x ?y
WHERE {
  # Bind ?x and ?y to the current time
  BIND(NOW() AS ?x)
  BIND(NOW() AS ?y)
}
```

should return the same value for ?x and ?y. If we would call now() twice ?x and ?y will be different.

liss-h: Ok, but thats kinda obvious, right? Apart from that it's not even `rdf4cpp`'s responsibilty to handle that, so I find it confusing that it is even mentioned here.

bigerl: It's obvious if you think about it. But it's also that kind of thing one might overlook and that causes unnecessary bugs. So, I see no harm in leaving the comment there but would also add the example from my comment above to make it more self-contained.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1984,6 +1984,152 @@ Literal Literal::sha512(NodeStorage &node_storage) const {
...
+    rdf::util::OptionalTimezone opt = std::nullopt;
+    return make_typed_from_value<datatypes::xsd::DateTime>(std::make_pair(t, opt), node_storage);
+}
+
+std::optional<std::chrono::year> Literal::year() const noexcept {
+    if (!datatype_eq<datatypes::xsd::DateTime>() && !datatype_eq<datatypes::xsd::DateTimeStamp>() && !datatype_eq<datatypes::xsd::Date>()
+            && !datatype_eq<datatypes::xsd::GYearMonth>() && !datatype_eq<datatypes::xsd::GYear>())
+        return std::nullopt;
+    auto casted = this->cast_to_value<datatypes::xsd::DateTime>();
+    if (!casted.has_value())
```
liss-h, line 2000: Can this happen?

src/rdf4cpp/rdf/Literal.hpp
liss-h: Ok found out why it exists.

What I don't like about it that is effectively the exact same function as `Literal::cast`.
I know you did it to avoid the overhead of creating a literal so my suggestion would be to resolve this the following way:

```c++
std::any Literal::cast_to_value(IRI const  &) {
    // implementation goes here
}

template<LiteralDatatype T>
std::optional<typename T::cpp_type> Literal::cast_to_value(NodeStorage &node_storage = ..) {
    if (auto v = cast_to_value(IRI{T::datatype_id, node_storage}); v.has_value()) {
        return std::any_cast<typename T::cpp_type>(v);
    }
    
    return std::nullopt;
}

Literal Literal::cast(IRI const &datatype) {
    auto v = cast_to_value(datatype);
    return Literal::make_typed_unchecked(...)
}

template<LiteralDatatype T>
Literal Literal::cast() {
    auto v = cast_to_value<T>();
    ....
}
```

Hope I didn't overlook something that makes this not work.
I know this is also not fully optimal performance wise. But I think the improved maintainability is more important here.

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -422,6 +422,107 @@ class Literal : public Node {
...
+
+        if (this_dtid == target_dtid) {
+            return this->value<T>();
+        }
+
+        if (this_dtid == String::datatype_id) {
+            // string -> any
+            try {
+                return T::from_string(this->lexical_form());
+            } catch (std::runtime_error const &) {
```
liss-h: Use `catch (...)` for robustness

src/rdf4cpp/rdf/datatypes/registry/util/DateTimeUtils.hpp
```diff
@@ -0,0 +1,170 @@
...
+            if (p14 != m14)
+                return std::partial_ordering::unordered;
+            return p14;
+        } else {
+            return a_tp <=> b_tp;
+        }
+    }
+}
+template<std::unsigned_integral T>
+constexpr T numberOfBits(T x) noexcept {
```
liss-h: snake_case

liss-h (changes requested):
Ok, I only have some minor things. Otherwise this is good to go from my side.

I haven't extensively looked at the parsing and serialization again, but if there are weird edge cases we will probably find them when this is deployed, I don't think just staring at the code and spec will go anywhere (at least not in a time efficient manner).

src/rdf4cpp/rdf/Literal.cpp
mcb5637: Only if the datatype registry changes.

src/rdf4cpp/rdf/Literal.hpp
mcb5637: I think that would not work, because casting to a string is a edge case i purposefully avoid for cast_to_value.
(due to the borrowed/owned semantic and xsd::string::cpp_type not having an option for that)

liss-h: Forgot about that one, bummer

src/rdf4cpp/rdf/Literal.hpp
liss-h: Could you also change that for `cast`? It has the same kinda catch block

liss-h (approved):

bigerl (approved):

PR #183, opened 2023-06-01
Fix: CI Debug build and flag optimization

liss-h:
The CI was not building the tests in ReleaseWithDebugInfo instead of Debug which will disable all `assert`s. This is not very useful for testing purposes. Also added an option that makes `ASan` output more readable, and enabled parallel build.

For some unknown-to-me reason I am not allowed to use yamls `>` to put the `CXXFLAGS` into a multiline block, cmake just can't cope for some reason. (Even though the `>` should simply replace all newlines in the block with single spaces.)


nkaralis (commented):
It looks good and clear to me. I am not sure why the newline does not work. Maybe, we should wait for @bigerl to have a look as well before merging, since he has set up these actions.

bigerl (approved):

PR #184, opened 2023-06-07
Fix: prevent default node storage instance overwrite

liss-h:
Previously a sequence of
```c++
NodeStorage::set_default_instance(...some-instance);
auto &dinst = NodeStorage::default_instance();
```
would overwrite the original node storage the user had set because the internal init-once-flag was never set by `set_default_instance()`, and therefore `default_instance()` would itself initialize the default instance to an instance of the default backend.

This PR does not synchronize multiple calls to `set_default_instance()` or `default_instance()` in general cases, so multiple threads trying to initialize the `default node storage` will not work reliably. But that is probably either something for later or never since at least to me it is not entirely clear why multiple threads would try to compete around setting the default instance.

### DIFF
--- a/src/rdf4cpp/rdf/storage/node/NodeStorage.cpp
+++ b/src/rdf4cpp/rdf/storage/node/NodeStorage.cpp
@@ -55,6 +55,7 @@ NodeStorage &NodeStorage::default_instance() {
 }
 
 void NodeStorage::set_default_instance(NodeStorage const &node_context) {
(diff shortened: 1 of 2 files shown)


bigerl (approved):
Good point and straight forward fix. 

I also don't see a use-case for concurrent set_default_instance calls. I've added https://github.com/rdf4cpp/rdf4cpp/issues/187 to document that

PR #186, opened 2023-06-12
fixed decimal simplified string conversion

### LINKED ISSUE #185: Casting `xsd:decimal` to `xsd:string` produces wrong results since update to new decimal
Casting an integral valued `xsd:decimal` to `xsd:string` is equivalent to first casting the `xsd:decimal` to `xsd:integer` and then to `xsd:string`. I.e. there should be no decimal point present in the representation (as per https://www.w3.org/TR/xpath-functions/#casting-to-string).

This is not the case anymore because `datatypes::xsd::Decimal::to_simplified_string` uses the normal conversion function.

The error was introduced [here](https://github.com/rdf4cpp/rdf4cpp/commit/ade3c951cfbdc215a8393e3d439c6e612f88b4f0#diff-32bf1ad020bb5e6df10af1001856d0fd42e3198ad5806e4ce217b1e8b1830bacR215-R216) because the test case was changed which got overlooked in the corresponding review.

mcb5637:
close #185

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/xsd/Decimal.cpp
+++ b/src/rdf4cpp/rdf/datatypes/xsd/Decimal.cpp
@@ -33,7 +33,7 @@ std::string capabilities::Default<xsd_decimal>::to_simplified_string(cpp_type co
     cpp_type v = value;
     v.normalize();
     if (v.get_exponent() == 0) {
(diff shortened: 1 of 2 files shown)


liss-h (approved):
Looks good

PR #188, opened 2023-06-16
Feature+Fix: adding `Node::try_get_in_node_storage` and fixing `Node::to_node_storage`

liss-h:
This PR adds `Node::try_get_in_node_storage` a similar function `Node::to_node_storage` except that it will fail if the node does not already exist in the specified target node storage. Basically used non-mutably translating between node storages.

Additionally it fixes the implementation of `Node::to_node_storage` which did not account for all cases previously.

I will now list all cases that need to be handled here, to make the review easier:
1. normal inlined lit (doesn't matter what the target node storage is, will stay inlined)
2. inlined `rdf:langString` -> able to inline in new storage
3. inlined `rdf:langString` -> not able to inline in new storage
4. specialized storage -> specialized storage
5. specialized storage -> non-specialized storage
6. non-specialized storage -> non-specialized storage
7. non-specialized storage -> specialized storage

I've added best effort tests, but I simply cannot test the transfer between different node storage implementations without writing another implementation.

Note: yes I know the implementation of `Literal::to_node_storage` and `Literal::try_get_in_node_storage` are very similiar but I couldn't find a good way to extract the common bits.

### DIFF
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -28,7 +28,7 @@ Literal Literal::make_null() noexcept {
     return Literal{};
 }
 
-Literal Literal::make_simple_unchecked(std::string_view lexical_form, Node::NodeStorage &node_storage) noexcept {
+Literal Literal::make_simple_unchecked(std::string_view lexical_form, NodeStorage &node_storage) noexcept {
     return Literal{NodeBackendHandle{node_storage.find_or_make_id(storage::node::view::LexicalFormLiteralBackendView{
                                              .datatype_id = storage::node::identifier::NodeID::xsd_string_iri.first,
                                              .lexical_form = lexical_form,
@@ -37,7 +37,7 @@ Literal Literal::make_simple_unchecked(std::string_view lexical_form, Node::Node
                                      node_storage.id()}};
 }
 
-Literal Literal::make_noninlined_typed_unchecked(std::string_view lexical_form, IRI const &datatype, Node::NodeStorage &node_storage) noexcept {
+Literal Literal::make_noninlined_typed_unchecked(std::string_view lexical_form, IRI const &datatype, NodeStorage &node_storage) noexcept {
     return Literal{NodeBackendHandle{node_storage.find_or_make_id(storage::node::view::LexicalFormLiteralBackendView{
                                              .datatype_id = datatype.to_node_storage(node_storage).backend_handle().node_id(),
                                              .lexical_form = lexical_form,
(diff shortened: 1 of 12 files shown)


src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -203,19 +203,46 @@ Literal Literal::generate_random_double(Node::NodeStorage &node_storage) {
     return Literal::make_typed_from_value<datatypes::xsd::Double>(state.dist(state.rng), node_storage);
 }
 
-Literal Literal::to_node_storage(Node::NodeStorage &node_storage) const noexcept {
+Literal Literal::to_node_storage(NodeStorage &node_storage) const noexcept {
```
bigerl, line 206: TODO: review this in detail

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -226,24 +253,127 @@ Literal Literal::to_node_storage(Node::NodeStorage &node_storage) const noexcept
                 return node_storage.find_or_make_id(literal_view);
             },
             [&node_storage, &literal_view](storage::node::view::ValueLiteralBackendView const &any) noexcept {
-                // no need to send over IRI, as it (a fixed datatype) must already be present
+                // no need to send over IRI, as this having a specialized storage requires
```
```suggestion
                // no need to send over datatype IRI, as this having a specialized storage requires
```

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -226,24 +253,127 @@ Literal Literal::to_node_storage(Node::NodeStorage &node_storage) const noexcept
                 return node_storage.find_or_make_id(literal_view);
             },
             [&node_storage, &literal_view](storage::node::view::ValueLiteralBackendView const &any) noexcept {
```
```suggestion
            [&node_storage, &literal_view](storage::node::view::ValueLiteralBackendView const &literal_backend) noexcept {
```

Or maybe something else. I find `any` confusing.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -226,24 +253,127 @@ Literal Literal::to_node_storage(Node::NodeStorage &node_storage) const noexcept
...
+                                                                                                 handle_.node_storage_id()});
+                lexical.datatype_id = node_storage.find_id(dtype_iri_view);
+                if (lexical.datatype_id == NodeID{}) {
+                    // datatype IRI not present, therefore literal cannot be present
+                    return NodeID{};
+                }
+
+                return node_storage.find_id(literal_view);
+            },
+            [&node_storage, &literal_view](storage::node::view::ValueLiteralBackendView const &any) noexcept {
```
bigerl: same here. I do not particularly like the variable name `any`

tests/nodes/tests_Literal.cpp
```diff
@@ -795,4 +795,182 @@ TEST_CASE("UUID") {
...
+
+            lit2 = lit.try_get_in_node_storage(ns2);
+            CHECK(!lit2.null());
+            CHECK(!lit2.is_inlined());
+            CHECK(lit.value<datatypes::xsd::String>() == lit2.value<datatypes::xsd::String>());
+            CHECK(lit.backend_handle().node_id().literal_type() == lit2.backend_handle().node_id().literal_type());
+            CHECK(lit.backend_handle().node_storage_id() != lit2.backend_handle().node_storage_id());
+        }
+    }
+}
```
bigerl: duplication?

bigerl (changes requested):
Very nice! Just cosmetic things in the code and there seems to be a duplication in the tests.

tests/nodes/tests_Literal.cpp
liss-h: No, just forgot to rename the test after copy pasting. This one tests `try_get_in_node_storage` the one above tests `to_node_storage`

src/rdf4cpp/rdf/Literal.cpp
bigerl: that todo is mainly what I just did 😅

PR #189, opened 2023-06-22
update CI and fix warnings

liss-h:
Also snuck some CI unrelated changes in there, mainly to silence useless warnings.
On the other hand I added `-Werror` on workflow runs with the most recent compilers.


.github/workflows/reusable_packaging_and_deployment.yml
```diff
@@ -40,10 +52,16 @@ jobs:
           CC: ${{ steps.install_cc.outputs.cc }}
           CXX: ${{ steps.install_cc.outputs.cxx }}
         run: |
-          pip3 install "conan==1.59" ${{ fromJSON('["", "gcovr"]')[inputs.with-coverage] }}
+          pip3 install "conan==1.59" yq
```
bigerl: I would suggest `conan==1.60.1`. With 1.60 support for gcc13 was added (see https://docs.conan.io/1/changelog.html#id3)

This would make yq obsolete

.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
```diff
@@ -51,10 +64,15 @@ jobs:
           CC: ${{ steps.install_cc.outputs.cc }}
           CXX: ${{ steps.install_cc.outputs.cxx }}
         run: |
-          pip3 install "conan==1.59" ${{ fromJSON('["", "gcovr"]')[inputs.with-coverage] }}
-          conan user
+          pip3 install "conan==1.59" yq ${{ fromJSON('["", "gcovr"]')[inputs.with-coverage] }}
```
bigerl: see other comment regarding Conan 1.60.1

conanfile.py
```diff
@@ -19,29 +19,32 @@ class Recipe(ConanFile):
 
     # Binary configuration
     settings = "os", "compiler", "build_type", "arch"
-    options = {"shared": [True, False], "fPIC": [True, False]}
-    default_options = {"shared": False, "fPIC": True}
+    options = {"shared": [True, False], "fPIC": [True, False], "with_test_deps": [True, False]}
+    default_options = {"shared": False, "fPIC": True, "with_test_deps": False}
     exports = "LICENSE",
     exports_sources = "src/*", "private/*", "CMakeLists.txt", "cmake/*"
-    requires = (("fmt/9.0.0", "private"),  # format must only be used within cpp files
```
bigerl, line 26: where did fmt go?

conanfile.py
```diff
@@ -19,29 +19,32 @@ class Recipe(ConanFile):
...
+        self.requires("uni-algo/0.7.1")
+
+        if self.options.with_test_deps:
+            self.requires("doctest/2.4.11")
+
+
     def set_version(self):
         if not hasattr(self, 'version') or self.version is None:
             cmake_file = load(self, os.path.join(self.recipe_folder, "CMakeLists.txt"))
-            self.version = re.search(r"project\([^)]*VERSION\s+(\d+\.\d+.\d+)[^)]*\)", cmake_file).group(1)
```
bigerl: 1. why did you remove this line? 
2. As you removed this line, why did you not remove set_version entirely?

src/rdf4cpp/rdf/ClosedNamespace.hpp
```diff
@@ -23,7 +22,7 @@ class ClosedNamespace : public Namespace {
     ClosedNamespace(std::string_view namespace_iri, Suffixes all_suffixes, NodeStorage &node_storage)
         : Namespace(namespace_iri, node_storage) {
         for (auto const &suffix : all_suffixes)
-            this->cache_.template emplace(suffix, IRI{namespace_iri_ + std::string{suffix}, node_storage}.backend_handle());
+            this->cache_.emplace(suffix, IRI{namespace_iri_ + std::string{suffix}, node_storage}.backend_handle());
```
bigerl, line 25: Is this change because of a warning. The template seemed fine to me.

.github/workflows/reusable_packaging_and_deployment.yml
```diff
@@ -40,10 +52,16 @@ jobs:
...
           conan user
           conan profile new --detect default
+          conan profile update env.CC="${CC}" default
```
bigerl: This is actually something I would suggest to do in cmake. See https://github.com/dice-group/tentris-cwr-pg/blob/f0cfa64a9bdb128b0dab3a98c2b164176ff9c78e/cmake/conan_cmake.cmake#L23

.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
```diff
@@ -51,10 +64,15 @@ jobs:
...
           conan profile new --detect default
+          conan profile update env.CC="${CC}" default
```
bigerl: see other comment to env variable is cmake

bigerl (changes requested):
Looks good and would work. However, I would do some things a little bit different and have some questions here and there. See below.

.github/workflows/reusable_packaging_and_deployment.yml
liss-h: Ok, will change didn't notice conan 1.60 was a thing

conanfile.py
liss-h: I intentionally removed with the rationale being that I'm trying to shave compile time wherever possible in our projects and `fmt` was very rarely used. Additionally all cases where it was could trivially be replaced by things in `std` without much downside. So it was kind of unecessary to have this as an additional dependency.

But of course if you feel that we absolutely need it I can revert those changes.

conanfile.py
liss-h: That must have been an accident, at least I don't remember doing that intentionally.

src/rdf4cpp/rdf/ClosedNamespace.hpp
liss-h: yes, this was fine just saw that by chance. Having `.template` is only necessary if you have 2 or more template arguments after the function name and this has none.

.github/workflows/reusable_packaging_and_deployment.yml
liss-h: `CONAN_DISABLE_STRICT_MODE=1` (https://github.com/dice-group/tentris-cwr-pg/blob/f0cfa64a9bdb128b0dab3a98c2b164176ff9c78e/cmake/conan_cmake.cmake#L23) this sounds like not so much of a good idea at least from what I could gather from the conan documentation

bigerl: Right, but the part `CXX=${CMAKE_CXX_COMPILER};CC=${CMAKE_C_COMPILER}` does the trick done here via the profile. This also helps that the CMake project simply builds and ensures that CMake and Conan use the same settings. Disabling strict mode was necessary at some point because of some buggy library but should not be necessary anymore.

conanfile.py
bigerl: OK, I am convinced.

.github/workflows/reusable_packaging_and_deployment.yml
liss-h: Ok got it, makes sense

liss-h: Looks like its working

bigerl (approved):

PR #191, opened 2023-07-19
Fix: literal serialization perf and some node storage related cleanup

liss-h:
TODO:
- [x] build tests with AVX2 and without


src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -16,6 +18,51 @@
...
+            if (_mm256_movemask_epi8(eq) > 0) {
+                return true;
+            }
+        }
+    }
+
+    auto const rest = lexical_form.size() % 32;
+    return lexical_form_needs_escape_non_simd(lexical_form.substr(lexical_form.size() - rest));
+}
+} // rdf4cpp::rdf
```
bigerl: I would in general suggest to use a SIMD library. Either [highway](https://github.com/google/highway) or [xsimd](https://github.com/google/highway). Implementation looks good though 👍🏻

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -16,6 +18,51 @@
...
+namespace rdf4cpp::rdf {
+bool Literal::lexical_form_needs_escape(std::string_view const lexical_form) noexcept {
+    // https://www.w3.org/TR/n-triples/#grammar-production-STRING_LITERAL_QUOTE
+    std::array<__m256i, 4> const masks{_mm256_set1_epi8('"'),
+                                       _mm256_set1_epi8('\\'),
+                                       _mm256_set1_epi8('\n'),
+                                       _mm256_set1_epi8('\r')};
+
+    for (size_t bix = 0; bix < lexical_form.size() / 32; ++bix) {
+        __m256i const chars = _mm256_loadu_si256(reinterpret_cast<__m256i const *>(lexical_form.data() + (bix * 32)));
```
bigerl, line 44: I looked a bit but didn't really find anything. Do you know if `_m256i` requires a particular alignment?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -212,8 +272,9 @@ Literal Literal::to_node_storage(NodeStorage &node_storage) const noexcept {
 
     if (this->is_inlined()) {
         if (this->datatype_eq<datatypes::rdf::LangString>()) {
-            auto const data = this->value<datatypes::rdf::LangString>();
```
bigerl, line 215: Is the idea behind this change to avoid the checks from `value()` (because things are clear)?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -542,81 +609,126 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
...
 
 Literal::operator std::string() const noexcept {
-    // TODO: escape non-standard chars correctly
-    auto const quoted_lexical_into_stream = [](std::ostream &out, std::string_view const lexical) noexcept {
-        // TODO: escape everything that needs to be escaped in N-Tripels/N-Quads
+    if (this->null()) {
+        return "null";
+    }
+
+    auto const escaped_size = [](size_t size) noexcept {
```
```suggestion
    auto const estimated_escaped_size = [](size_t size) noexcept {
```

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -542,81 +609,126 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
...
+        auto const value = this->lang_tagged_get_de_inlined().backend_handle().literal_backend().get_lexical();
+        buf.reserve(escaped_size(value.lexical_form.size()) + value.language_tag.size() + 1);
+
+        if (value.needs_escape) [[unlikely]] {
+            append_quoted_lexical(buf, value.lexical_form);
+        } else {
+            buf.push_back('"');
+            buf.append(value.lexical_form);
+            buf.push_back('"');
         }
```
bigerl: reserve buffer in the (value.needs_escape) branches to be more precise in memory usage.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -542,81 +609,126 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
...
 
-        quoted_lexical_into_stream(oss, this->lexical_form());
+        buf.push_back('@');
+        buf.append(value.language_tag);
+    } else if (this->is_inlined()) {
+        assert(!this->datatype_eq<datatypes::rdf::LangString>());
+        // Notes:
+        // 1. inlined values are assumed to not require escaping
```
bigerl, line 671: This must be documented clearly.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -542,81 +609,126 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
...
-        auto const &dtype_iri = NodeStorage::find_iri_backend_view(storage::node::identifier::datatype_iri_handle_for_fixed_lit_handle(handle_));
+        auto const *entry = datatypes::registry::DatatypeRegistry::get_entry(this->datatype_id());
+        assert(entry != nullptr);
+        assert(entry->inlining_ops.has_value());
 
-        oss << "^^" << dtype_iri.n_string();
-    } else if (this->datatype_eq<datatypes::rdf::LangString>()) {
-        auto const value = this->value<datatypes::rdf::LangString>();
+        auto const inlined_value = this->backend_handle().node_id().literal_id();
+        auto const lexical_form = entry->to_canonical_string_fptr(entry->inlining_ops->from_inlined_fptr(inlined_value));
```
bigerl, line 679: This is very hard to read but yes, we don't need the checks of value (again).

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -542,81 +609,126 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
...
+                    buf.reserve(escaped_size(lexical_backend.lexical_form.size()) + dtype_iri.identifier.size() + 4);
 
-                    quoted_lexical_into_stream(oss, to_string(value_backend.value));
-                    oss << "^^" << dtype_iri.n_string();
+                    if (lexical_backend.needs_escape) [[unlikely]] {
+                        append_quoted_lexical(buf, lexical_backend.lexical_form);
+                    } else {
+                        buf.push_back('"');
+                        buf.append(lexical_backend.lexical_form);
+                        buf.push_back('"');
```
bigerl: move reserve into the if

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1762,7 +1881,7 @@ Literal Literal::encode_for_uri(std::string_view string, NodeStorage &node_stora
             }
         }
     }
-    return make_simple_unchecked(stream.view(), node_storage);
+    return make_simple_unchecked(stream.view(), false, node_storage); // TODO check
```
bigerl: Should be fine because only `A–Za–z0–9` and `- . _ ~` are allowed.

src/rdf4cpp/rdf/namespaces/util/NamespaceInstance.hpp
```diff
@@ -13,7 +14,7 @@ template<typename NamespaceClass>
 struct NamespaceInstance {
     static NamespaceClass &instance(storage::node::NodeStorage &node_storage = storage::node::NodeStorage::default_instance()) {
         static std::mutex instances_mutex;
-        static storage::util::tsl::sparse_map<uint16_t, NamespaceClass> instances;
+        static dice::sparse_map::sparse_map<uint16_t, NamespaceClass> instances;
```
bigerl: use wyhash here aswell

src/rdf4cpp/rdf/storage/node/identifier/NodeBackendHandle.cpp
```diff
@@ -2,52 +2,23 @@
...
-    constexpr NodeBackendHandleImpl(NodeID node_id, RDFNodeType node_type, NodeStorageID node_storage_id, bool inlined, uint8_t tagging_bits) noexcept
-        : fields_{.node_id_ = node_id, .type_ = node_type, .storage_id_ = node_storage_id.value, .inlined = inlined, .free_tagging_bits = tagging_bits} {}
-
-    [[nodiscard]] constexpr NodeStorageID storage_id() const noexcept { return NodeStorageID{fields_.storage_id_}; }
-    [[nodiscard]] constexpr NodeID node_id() const noexcept { return fields_.node_id_; }
+    NodeID node_id_;
+    RDFNodeType type_ : 2;
+    NodeStorageID::underlying_type storage_id_ : NodeStorageID::width;
+    uint8_t inlined : 1;
+    uint8_t free_tagging_bits : 3;
```
bigerl, line 14: much better 👍🏻

src/rdf4cpp/rdf/storage/node/reference_node_storage/NodeTypeStorage.hpp
```diff
@@ -49,7 +49,7 @@ struct NodeTypeStorage {
 
     struct DefaultBackendTypeHash {
```
bigerl, line 48: Is there an alternative? If not, is SelectBackendTypeHash still needed?

src/rdf4cpp/rdf/storage/node/reference_node_storage/ReferenceNodeStorageBackend.cpp
```diff
@@ -238,6 +238,8 @@ identifier::NodeID ReferenceNodeStorageBackend::find_id(view::LiteralBackendView
             [this](view::ValueLiteralBackendView const &any) {
                 return visit_specialized(specialized_literal_storage_, any.datatype, [this, &any](auto const &storage) {
                     assert(this->has_specialized_storage_for(any.datatype));
+                    (void) this;
```
bigerl, line 241: what's that good for? To suppress warnings?

bigerl (changes requested):
This looks very good and includes a bunch of sensible cleanups along the road. The main thing I'd suggest would be to use a library for the SIMD stuff. The rest is minor things, comments and remarks from my side. 
Please also double check @nkaralis' branch [here](https://github.com/rdf4cpp/rdf4cpp/tree/feature/improve-literal-serialization) that all optimizations from there are covered here as well. The only thing we will probably need to discuss later is if we migrate from sparse_map to unordered_dense. But that's out of scope for this PR.

src/rdf4cpp/rdf/Literal.cpp
liss-h: Normally yes, but `_mm256_loadu_si256` is an unaligned load.

"Moves integer values from unaligned memory location to a destination vector." - Intel

src/rdf4cpp/rdf/Literal.cpp
liss-h: Yes, and additionally the `.value()` doesn't have the `needs_escape` flag which I need.

src/rdf4cpp/rdf/storage/node/reference_node_storage/ReferenceNodeStorageBackend.cpp
liss-h: Yes, otherwise the compiler warns in release builds that the `this` capture is not used because the `assert` doesn't exist in those. There is also no way to conditionally capture. (Well at least no pretty way)

src/rdf4cpp/rdf/storage/node/identifier/NodeBackendHandle.cpp
liss-h: 🙂

src/rdf4cpp/rdf/namespaces/util/NamespaceInstance.hpp
liss-h: good catch

src/rdf4cpp/rdf/Literal.cpp
liss-h: Added comments to the corresponding READMEs

src/rdf4cpp/rdf/storage/node/reference_node_storage/NodeTypeStorage.hpp
liss-h: Yes, if `Backend == SpecializedLiteralBackend<T>` the hasher and equality function are provided by that. Because it needs to be able to unwrap the `std::any` value from a view to be able to compare it to its internal value.

src/rdf4cpp/rdf/Literal.cpp
liss-h: I thought this is simple enough to just do by hand

liss-h:
The additional things in Nikos branch are:
1. using unordered_dense
2. some `noexcept`s for `NodeStorage::find_*_backend_view`, which I originally thought were a good idea (which is also what I told him) but in hindsight probably aren't, because while the functions in `Literal/IRI...` that access the node storage should be `noexcept`, the functions for accessing the node storage directly probably shouldn't as there might be some more extreme optimizations that won't be possible anymore (we don't do any of those yet btw). And I would at least hope the codegen looks identical whether the noexcept is just at the top level or all the way through.

src/rdf4cpp/rdf/Literal.cpp
bigerl: It is but I am thinking that we will need to touch it again when we start to support more platforms and compilers, probably. For now, I am also fine with creating an issue to keep track of it.

bigerl:
> The additional things in Nikos branch are:
> 
>     1. using unordered_dense
> 
>     2. some `noexcept`s for `NodeStorage::find_*_backend_view`, which I originally thought were a good idea (which is also what I told him) but in hindsight probably aren't, because while the functions in `Literal/IRI...` that access the node storage should be `noexcept`, the functions for accessing the node storage directly probably shouldn't as there might be some more extreme optimizations that won't be possible anymore (we don't do any of those yet btw). And I would at least hope the codegen looks identical whether the noexcept is just at the top level or all the way through.

The question is valid how compilers handle non-noexcept functions within noexcept functions. I would expect they just don't and link to the same non-noexcept function assembly. Otherwise, functions would need to be compiled twice depending on where they are called from.
So to really benefit from noexcept I think we need to use it all the way down. Maybe we can have a look together at your concerns regarding using noexcept in the inner guts in the next days. Or can you directly pinpoint what you think might be probematic already?

src/rdf4cpp/rdf/Literal.cpp
liss-h: Created an issue

liss-h:
I have 2 new insights regarding the exception specs:
1. calling a throwing function inside a noexcept function seems to basically generate a `try {} catch (...) { terminate() }` (if I read the assembly correctly, https://godbolt.org/z/ezcTov1oo)
2. I was not able to observe any difference in performance from adding the `noexcept`s there beyond what can be attributed test inaccuracy. Although my benchmark might not be optimal for that or something, not sure.

.github/workflows/reusable_packaging_and_deployment.yml
```diff
@@ -53,6 +53,7 @@ jobs:
           conan user
           conan profile new --detect default
           conan profile update settings.compiler.libcxx=libstdc++11 default
+          conan remote add -f dice-group https://conan.dice-research.org/artifactory/api/conan/tentris
```
bigerl, line 56: non-blocking comment:  What do we need from here that is not available at conan center? We should have everything we use for rdf4cpp on conan center. 

If we add this here we also must mention in the readme that you need to add the remote.

.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
```diff
@@ -59,6 +61,22 @@ jobs:
...
+          echo 'int main() { return __AVX2__; }' > /tmp/rdf4cpp-avx-test.cpp
+          
+          AVX_EXISTS=$(${CXX} ${{ fromJSON('["", "-march=native"]')[inputs.with-AVX] }} -o /dev/null /tmp/rdf4cpp-avx-test.cpp > /dev/null 2>&1; test $? -eq 0; echo $?)
+          AVX_SHOULD_EXIST=${{ fromJSON('["1", "0"]')[inputs.with-AVX] }}
+          
+          # Shell: true=zero-exit, false=non-zero-exit
+          echo "AVX exists: $AVX_EXISTS"
+          echo "AVX should exist: $AVX_SHOULD_EXIST"
+          
+          test $AVX_EXISTS -eq $AVX_SHOULD_EXIST
```
bigerl, line 81: Please add a short comment what this does.

bigerl (changes requested):
final review. When these are done or moved to new issues this is ready to merge.

.github/workflows/reusable_packaging_and_deployment.yml
liss-h: Updated the README. We need `dice-hash` which isn't on conancenter

PR #196, opened 2023-08-28
date time arithmetic

mcb5637:


src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1273,6 +1274,41 @@ bool Literal::is_string_like() const noexcept {
 }
 
 Literal Literal::add(Literal const &other, Node::NodeStorage &node_storage) const noexcept {
+    if ((this->datatype_eq<datatypes::xsd::DateTime>() || this->datatype_eq<datatypes::xsd::DateTimeStamp>() || this->datatype_eq<datatypes::xsd::Date>() || this->datatype_eq<datatypes::xsd::Time>()) &&
```
liss-h: This function is incredibly unreadable. Please add some spaces in appropriate places

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1295,6 +1331,57 @@ Literal &Literal::operator+=(const Literal &other) noexcept {
 }
 
 Literal Literal::sub(Literal const &other, Node::NodeStorage &node_storage) const noexcept {
+    if ((this->datatype_eq<datatypes::xsd::DateTime>() || this->datatype_eq<datatypes::xsd::Date>() || this->datatype_eq<datatypes::xsd::DateTimeStamp>() || this->datatype_eq<datatypes::xsd::Time>()) &&
```
liss-h: Same here

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1339,6 +1442,36 @@ Literal &Literal::operator*=(const Literal &other) noexcept {
 }
 
 Literal Literal::div(Literal const &other, Node::NodeStorage &node_storage) const noexcept {
+    if (this->datatype_eq<datatypes::xsd::YearMonthDuration>() && other.datatype_eq<datatypes::xsd::Double>()) {
```
liss-h: Same here

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1317,6 +1404,22 @@ Literal &Literal::operator-=(const Literal &other) noexcept {
 }
 
 Literal Literal::mul(Literal const &other, Node::NodeStorage &node_storage) const noexcept {
+    if (this->datatype_eq<datatypes::xsd::YearMonthDuration>() && other.datatype_eq<datatypes::xsd::Double>()) {
```
liss-h: Same here

src/rdf4cpp/rdf/datatypes/registry/util/DateTimeUtils.hpp
```diff
@@ -7,10 +7,47 @@
...
+using CheckedZonedTime = std::chrono::zoned_time<CheckedMilliseconds, rdf4cpp::rdf::util::Timezone>;
+using CheckedTimePointSys = std::chrono::time_point<std::chrono::system_clock, CheckedMilliseconds>;
+
+template<class R>
+std::chrono::duration<rdf4cpp::rdf::util::CheckedIntegral<int64_t>, R> to_checked(std::chrono::duration<int64_t, R> v) noexcept {
+    return std::chrono::duration<rdf4cpp::rdf::util::CheckedIntegral<int64_t>, R>{v.count()};
+}
+template<class R>
+std::chrono::duration<int64_t, R> from_checked(std::chrono::duration<rdf4cpp::rdf::util::CheckedIntegral<int64_t>, R> v) noexcept {
+    return std::chrono::duration<int64_t, R>{v.count().get_value()};
```
liss-h, line 47:
```c++
assert(!v.count().is_invalid());
```

Probably wouldn't hurt

src/rdf4cpp/rdf/util/CheckedInt.hpp
```diff
@@ -0,0 +1,93 @@
...
+    constexpr std::partial_ordering operator<=>(const CheckedIntegral &other) const noexcept {
+        if (this->invalid && other.invalid)
+            return std::partial_ordering::equivalent;
+        if (this->invalid != other.invalid)
+            return std::partial_ordering::unordered;
+        return this->value <=> other.value;
+    }
+
+    constexpr CheckedIntegral &operator+=(const CheckedIntegral &other) noexcept {
+        if (this->invalid || other.invalid) {
```
liss-h: Can avoid branches if you do something like this:
```c++
this->invalid |= other.invalid;
this->invalid |= __builtin_add_overflow(this->value, other.value, &this->value);
```

Same applies to functions below

tests/datatype/tests_time_types.cpp
```diff
@@ -554,3 +554,55 @@ TEST_CASE("Literal API") {
     CHECK(Literal::make_typed<datatypes::xsd::GDay>("3").as_tz() == Literal::make_simple(""));
     CHECK(Literal::make_simple("5").as_tz() == Literal{});
 }
+
+TEST_CASE("arithmetic") {
```
liss-h, line 558: Todo to self: check this

liss-h (changes requested):
Initial review, I've skipped the impl in Literal and the tests for now.

tests/datatype/tests_time_types.cpp
liss-h: If I remember correctly you based the behaviour on some reference implementations, correct? If so please mention somewhere what the current behaviour was inspired by.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1317,6 +1404,22 @@ Literal &Literal::operator-=(const Literal &other) noexcept {
...
+        auto ym = this->value<datatypes::xsd::YearMonthDuration>();
+        auto d = other.value<datatypes::xsd::Double>();
+        auto r = std::round(static_cast<double>(ym.count()) * d);
+        if (!datatypes::registry::util::fits_into<int64_t>(r))
+            return Literal{};
+        return make_typed_from_value<datatypes::xsd::YearMonthDuration>(std::chrono::months{static_cast<int64_t>(r)});
+    } else if (this->datatype_eq<datatypes::xsd::DayTimeDuration>() && other.datatype_eq<datatypes::xsd::Double>()) {
```
nkaralis: what about xsd:Duration? Can it participate in multiplications with doubles?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1339,6 +1442,36 @@ Literal &Literal::operator*=(const Literal &other) noexcept {
...
 
 Literal Literal::div(Literal const &other, Node::NodeStorage &node_storage) const noexcept {
+    if (this->datatype_eq<datatypes::xsd::YearMonthDuration>() && other.datatype_eq<datatypes::xsd::Double>()) {
+        auto ym = this->value<datatypes::xsd::YearMonthDuration>();
+        auto d = other.value<datatypes::xsd::Double>();
+        auto r = static_cast<int64_t>(std::round(static_cast<double>(ym.count()) / d));
+        if (!datatypes::registry::util::fits_into<int64_t>(r))
+            return Literal{};
+        return make_typed_from_value<datatypes::xsd::YearMonthDuration>(std::chrono::months{r});
+    } else if (this->datatype_eq<datatypes::xsd::DayTimeDuration>() && other.datatype_eq<datatypes::xsd::Double>()) {
```
nkaralis: as in `Literal::mul`. do we need cases for xsd:Duration?

src/rdf4cpp/rdf/util/CheckedInt.hpp
```diff
@@ -0,0 +1,93 @@
+#ifndef RDF4CPP_CHECKEDINT_HPP
+#define RDF4CPP_CHECKEDINT_HPP
+
+#endif  //RDF4CPP_CHECKEDINT_HPP
+
+namespace rdf4cpp::rdf::util {
+template<std::integral I>
+class CheckedIntegral {
```
nkaralis, line 11: you could add comments to describe the purpose of this class.

src/rdf4cpp/rdf/datatypes/registry/util/DateTimeUtils.hpp
```diff
@@ -7,10 +7,47 @@
...
+using CheckedTimePoint = std::chrono::time_point<std::chrono::local_t, CheckedMilliseconds>;
+using CheckedZonedTime = std::chrono::zoned_time<CheckedMilliseconds, rdf4cpp::rdf::util::Timezone>;
+using CheckedTimePointSys = std::chrono::time_point<std::chrono::system_clock, CheckedMilliseconds>;
+
+template<class R>
+std::chrono::duration<rdf4cpp::rdf::util::CheckedIntegral<int64_t>, R> to_checked(std::chrono::duration<int64_t, R> v) noexcept {
+    return std::chrono::duration<rdf4cpp::rdf::util::CheckedIntegral<int64_t>, R>{v.count()};
+}
+template<class R>
+std::chrono::duration<int64_t, R> from_checked(std::chrono::duration<rdf4cpp::rdf::util::CheckedIntegral<int64_t>, R> v) noexcept {
```
nkaralis, line 45: could you please add comments to describe the purpose of the functions?

src/rdf4cpp/rdf/Literal.cpp
nkaralis: i want to add here that having comments that describe the cases would make things easier to understand. 
i think currently it is easy to miss some cases.

if it does not affect the performance and it is easy to implement, i think it would be even better if you could move the implementation to separate functions.

nkaralis (changes requested):
I had a quick look.

For the time being, I mostly have some suggestions to make the code a bit easier to read.

Please provide us with a reference to the spec of the operations. With the spec it would be easier to validate the test cases.

mcb5637:
Implementation is based on the current XPath spec: https://www.w3.org/TR/xpath-functions-31/#dateTime-arithmetic and https://www.w3.org/TR/xpath-functions-31/#duration-arithmetic
(which is likely what any future SPARQL spec will just use).

src/rdf4cpp/rdf/Literal.cpp
mcb5637: According to XPath: No.
(It does not even specify a function for DateTime + Duration, but defined an algorythm to do the calculation.
I added it anyway, because it needs less type checks.)
(Maybe because Duration is a DayTimeDuration + a YearMonthDuration internaly anyway.)

liss-h:
I've added a commit to improve the formatting/readability. 

What I've noticed is that this does not handle subtypes correctly. I've made this exact mistake in the past and never noticed. Basically, the special cases for `yearMonthDuration` and such need to handle subtypes of those types correctly. (There are currently no subtypes but a user might add them). Added `TODO` comments in all the places I noticed this error. Some of them are not related to this PR.

This basically needs a function like `cast_to_value` specifically only casts to supertypes. If you need help with that just ping me, I can implement that for you if needed.

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -531,6 +531,53 @@ class Literal : public Node {
...
 
+        /**
+         * Tries to cast this literal to a literal of the given type and return the result without creating a Literal.
+         * Only considers casts from subtype to supertype.
+         *
+         * @return conversion result
+         */
+        template<datatypes::LiteralDatatype T>
+        requires (!std::same_as<T, datatypes::xsd::String>)
+        std::optional<typename T::cpp_type> cast_to_supertype_value() const noexcept {
```
liss-h, line 532: Looks good, would have done it the same way

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1384,77 +1382,66 @@ Literal &Literal::operator+=(const Literal &other) noexcept {
...
-        datatypes::registry::util::CheckedZonedTime const this_tp{this_dt->second.has_value() ? *this_dt->second : util::Timezone{},
-                                                                  datatypes::registry::util::to_checked(this_dt->first)};
-
-        datatypes::registry::util::CheckedZonedTime const other_tp{other_dt->second.has_value() ? *other_dt->second : util::Timezone{},
-                                                                   datatypes::registry::util::to_checked(other_dt->first)};
-
-        auto d = this_tp.get_sys_time() - other_tp.get_sys_time();
-        if (d.count().is_invalid())
-            return Literal{};
+    if (this->datatype_id() == other.datatype_id()) {
```
liss-h, line 1361: Does the spec specify that you can only do `sub` if the types are equal?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1384,77 +1382,66 @@ Literal &Literal::operator+=(const Literal &other) noexcept {
 }
 
 Literal Literal::sub(Literal const &other, Node::NodeStorage &node_storage) const noexcept {
-    // TODO subtype handling
-
     // DateTime(&Subclasses) - DateTime(&Subclasses)
-    if (this->is_timepoint_like() && other.is_timepoint_like() && this->datatype_id() == other.datatype_id()) {
```
liss-h: If `is_timepoint_like` and `is_duration_like` are no longer used you can remove them again.

liss-h (commented):
Subtype support looks good, just some small questions/suggestions

src/rdf4cpp/rdf/Literal.cpp
mcb5637: It specifies a function for each datatype with both parameters of the same type.
(Also, does `Time - Date` make any sense? Typecasting (as described in xsd) makes it `1972 - Date` which is likely not what the average user expects)

liss-h: No the `Time - Date` thing doesn't make much sense. But I would say subtracting two different date-like types does make sense. But if the spec says there is only a function for equal types then thats fine

liss-h (approved):

src/rdf4cpp/rdf/util/CheckedInt.hpp
```diff
@@ -0,0 +1,96 @@
+#ifndef RDF4CPP_CHECKEDINT_HPP
+#define RDF4CPP_CHECKEDINT_HPP
+
+#endif  //RDF4CPP_CHECKEDINT_HPP
```
nkaralis: is this out of place?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1281,10 +1282,59 @@ bool Literal::is_string_like() const noexcept {
...
+        auto other_v = other.cast_to_supertype_value<datatypes::xsd::YearMonthDuration>();
+        if (this_v.has_value() && other_v.has_value()) {
+            auto r = datatypes::registry::util::to_checked(*this_v) + datatypes::registry::util::to_checked(*other_v);
+            if (r.count().is_invalid())
+                return Literal{};
+
+            return make_typed_from_value<datatypes::xsd::YearMonthDuration>(datatypes::registry::util::from_checked(r));
+        }
+    }
+    // DateTime(&Subclasses) + Duration(&Subclasses)
```
nkaralis, line 1314: this covers more cases than the ones defined in the standard, right? the standard defines (dayTimeDuration+time), (dayTimeDuration+date), (yearMonthDuration+time), and (yearMonthDuration+time).

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -1307,6 +1357,68 @@ Literal &Literal::operator+=(const Literal &other) noexcept {
...
+                                                                       datatypes::registry::util::to_checked(other_dt->first)};
+
+            auto d = this_tp.get_sys_time() - other_tp.get_sys_time();
+            if (d.count().is_invalid())
+                return Literal{};
+
+            return make_typed_from_value<datatypes::xsd::DayTimeDuration>(datatypes::registry::util::from_checked(d));
+        }
+    }
+    // DateTime(&Subclasses) - Duration(&Subclasses)
```
nkaralis, line 1378: As above, this covers more cases than the ones defined in the standard, right?

nkaralis (commented):
thanks for addressing the comments. it looks good to me and there are many test cases. I have left some questions.

src/rdf4cpp/rdf/util/CheckedInt.hpp
mcb5637: Yes, good catch

src/rdf4cpp/rdf/Literal.cpp
mcb5637: Yes, my reasoning for that is: the algorythm to do the calculation is defined and it simplifies code to not differentiate between the 2 cases.

nkaralis (approved):

PR #197, opened 2023-09-06
Fix: boost dep in conan package

liss-h:
rdf4cpp specified `boost:header_only=True` explicitly which prevents packages using rdf4cpp and boost from building


liss-h:
gcc-12 segfaults, pretty sure thats unrelated to this PR

bigerl (approved):
Changes look straight-forward. 

segfault of gcc is really strange, though.

PR #200, opened 2023-09-11
decimal rounding fix

### LINKED ISSUE #199: Floor(), Ceiling() and Round() operations do not work as intented with negative decimal literals
Examples:

1. actual: `?num:"-1.6"^^<http://www.w3.org/2001/XMLSchema#decimal> ?round:"-1.0"^^<http://www.w3.org/2001/XMLSchema#decimal> ?s:<http://example.org/n2>`
    expected: `?num:"-1.6"^^<http://www.w3.org/2001/XMLSchema#decimal> ?round:"-2.0"^^<http://www.w3.org/2001/XMLSchema#decimal> ?s:<http://example.org/n2>`
2. actual: `?floor:"-1.0"^^<http://www.w3.org/2001/XMLSchema#decimal> ?num:"-1.6"^^<http://www.w3.org/2001/XMLSchema#decimal> ?s:<http://example.org/n2>`
    expected: `?floor:"-2.0"^^<http://www.w3.org/2001/XMLSchema#decimal> ?num:"-1.6"^^<http://www.w3.org/2001/XMLSchema#decimal> ?s:<http://example.org/n2> `
3. actual: `?ceil:"0.0"^^<http://www.w3.org/2001/XMLSchema#decimal> ?num:"-1.6"^^<http://www.w3.org/2001/XMLSchema#decimal> ?s:<http://example.org/n2>`
    expected: `?ceil:"-1.0"^^<http://www.w3.org/2001/XMLSchema#decimal> ?num:"-1.6"^^<http://www.w3.org/2001/XMLSchema#decimal> ?s:<http://example.org/n2>`

mcb5637:
close #199

### DIFF
--- a/src/rdf4cpp/rdf/util/BigDecimal.hpp
+++ b/src/rdf4cpp/rdf/util/BigDecimal.hpp
@@ -313,14 +313,29 @@ private:
     constexpr static BigDecimal handle_rounding(UnscaledValue_t v, Exponent_t e, UnscaledValue_t rem, RoundingMode m) noexcept {
         switch (m) {
             case RoundingMode::Floor:
-                return BigDecimal{v, e};
-            case RoundingMode::Ceil:
-                return BigDecimal{v + 1, e};
-            case RoundingMode::Round:
-                if (rem >= 5)
-                    return BigDecimal{v + 1, e};
-                else
+                if (v >= 0 || rem == 0)
                     return BigDecimal{v, e};
+                else
+                    return BigDecimal{v - 1, e};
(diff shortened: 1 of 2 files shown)


mcb5637:
please ignore gcc12 failing
gcc12 will get removed by #179 anyway

nkaralis:
It looks good to me thank you. 
This is me nitpicking, but I have a small request. Could you also please add test cases for -1.4? Since you already have test cases for -1.6, I think it would be better to also cover case where the decimal part is smaller than 0.5.

PR #202, opened 2023-09-13
Fix: decimal to integral cast

liss-h:
#200 introduced a bug that causes decimal to integral casts to yield invalid values. This fixes that.


bigerl (approved):
Nice solution with Trunc 👍🏻

Issue #204, opened 2023-09-20
Improve datatype parsing error message


liss-h:
Currently errors that occur during datatype parsing are often confusing or just not very good. This needs to be improved.

Addions by @bigerl:
The `ConstexprString datatype` parameter in some parsing functions does not enforce in any way to be set to correct values only. It's very easy to set it to some random value. I have no simple idea at hand but a more robust approach would be preferable.

In general a more robust way to generate messages would be preferable.

KonradHoeffner:
The errors are also too many, for example when I pipe them into a file it has 21 MB of "invalid date..." messages after loading an 1.2GB Turtle file.

liss-h:
> The errors are also too many, for example when I pipe them into a file it has 21 MB of "invalid date..." messages after loading an 1.2GB Turtle file.

Have you verified that the triples/dates in that file are actually correct? If so it would be very helpful if you could post them so we can see which triples we erroneously reject and fix it.

KonradHoeffner:
The file in question is <http://downloads.dbpedia.org/2016-10/core-i18n/en/persondata_en.ttl.bz2> and the dates do seem partially wrong (some are missing the leading zero for months and days, others even have 0 for the month or day) but in this case there are 182774 errors, I think it would be enough to report the first 10 or so, to not slow the program down, or to allow users to ignore the errors somehow.

Some of the triples

```
<http://dbpedia.org/resource/Buddie_Petit> <http://dbpedia.org/ontology/birthDate> "1890-0-0"^^<http://www.w3.org/2001/XMLSchema#date> .
<http://dbpedia.org/resource/Buddy_Emmons> <http://dbpedia.org/ontology/deathDate> "2015-7-0"^^<http://www.w3.org/2001/XMLSchema#date> .
<http://dbpedia.org/resource/Margaret_of_Foix-Candale> <http://dbpedia.org/ontology/birthDate> "1473-0-0"^^<http://www.w3.org/2001/XMLSchema#date> .
```

Having month or day as 0 seems indeed wrong, I guess the authors wanted to express that they didn't know the exact date.
So in summary, the triples are correct RDF but the datatypes are invalid dates, which are rightfully reported as incorrect, but from the data consumption side (e.g. using DBpedia where this is from) I am using the data as is so it would be nice if the errors could be silenced or abbreviated.

liss-h:
I mean, rdf4cpp is a library so you can make your program do whatever you want. 

[...]

PR #205, opened 2023-09-20
Fix: parsing of some date types

liss-h:


liss-h:
Error messages are also not very good in some places. This PR improves this slightly to at least not be confusing but still far from good. Related #204

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -19,7 +19,7 @@ namespace rdf4cpp::rdf::datatypes::registry::util {
  * @return the resulting value
  * @throws std::runtime_error if the string cannot be parsed
  */
-template<std::integral I>
+template<std::integral I, ConstexprString datatype>
```
bigerl, line 22: Please document the template parameters here and in DateTimeUtils. Can be adressed together with making the `ConstexprString datatype` parameters more robust.

bigerl, line 22: `ConstexprString datatype` does not enforce in any way to be set to correct values only. It's very easy to set it to some random value. I have no simple idea at hand but a more robust approach would be preferable. This does not need to be addressed here but can be moved to a separate issue.

bigerl (approved):
Changes look good. I found a thing that would need improvement but that should be covered in another issue.

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
liss-h: moved to error message issue

liss-h: moved to error message issue

PR #206, opened 2023-09-20
documentation & tutorials

### LINKED ISSUE #144: Feature: Documentation & Tutorials page
We need the following in the form of one (or more) websites:
- getting started
- tutorial
- Class, function, method and Concept Documentation
- in-depth technical documentation

Interesting projects that do all seem to be [Breathe](https://github.com/breathe-doc/breathe) and [Exhale](https://github.com/svenevs/exhale). Main obstacle is that things around concepts are not yet completely supported, see https://github.com/mpusz/units/issues/167 and https://github.com/breathe-doc/breathe/issues/482 .

CI should be built for develop, main (whenever we reach 0.1.0) and releases. Ideally, one can select the version for the documentation. 

This might be a good starting point. I think it looks quite good. https://github.com/mpusz/units

mcb5637:
close #144 
also added the documentation for #210


mcb5637:
@lennartaustenfeld could you please review the documentation build process including doxygen/sphinx?

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
```diff
@@ -2,6 +2,7 @@
 
 namespace rdf4cpp::rdf::datatypes::registry {
 
+#ifndef DOXYGEN_PARSER
```
liss-h, line 5: Can we move this somewhere more top level, or find another workaround? Something like you did for `mz::` maybe? (What I had in mind there would be to blacklist all of `rdf4cpp::rdf::datatypes::registry` and then whitelist only `rdf4cpp::rdf::datatypes::registry::DatatypeRegistry` or something)

Point mainly being that I would prefer if we could work around this without modifying the source if at all possible. Well if not then I guess we don't have much of a choice

liss-h (changes requested):
Some suggestions:

Could you prevent the `*_detail` namespaces from showing up in the documentation (e.g. `rdf4cpp::rdf::storage::node::node_storage_detail`). There is also some other stuff in there that users shouldn't be able to see, (e.g. `SelectOpRes`).

In general anything that is not part of the official public interface should be hidden imo, because it will scare off users with that amount of information that is at the end of the day not useful for them anyways, these things are after all implementation details.

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
mcb5637: Unfortunately neither doxygen nor breathe+exhale can distinguish between the template declarations and their specializations. So our only other option would be to ignore the complete registry namespace. But the technical documentation refers to it a lot, to make sure any advanced user will implement the proper interfaces for their custom types.
So in the end this seems to be the only option.

lennartaustenfeld (approved):
To my limited knowledge on doxygen and sphinx, it looks fine.

src/rdf4cpp/rdf/datatypes/owl/Rational.cpp
liss-h: Ok then

liss-h (approved):

PR #212, opened 2023-09-28
date time invalid_argument to runtime_error

### LINKED ISSUE #211: Parser crash
This triple crashes the parser:
```
<http://dbpedia.org/resource/Nikolai_Skoblin> <http://dbpedia.org/ontology/deathDate> "1937-0-0"^^<http://www.w3.org/2001/XMLSchema#date> .
```

used:
```c++
{
    using namespace rdf4cpp::rdf;
    parser::ParsingFlags flags = parser::ParsingFlag::Turtle;
    for (const auto &e: parser::RDFFileParser{filename, flags}) {
      if (e.has_value()) {
        count++;
      } else
        std::cerr << e.error();
    }
  }
```

mcb5637:
fix #211

### DIFF
--- a/src/rdf4cpp/rdf/datatypes/registry/util/DateTimeUtils.hpp
+++ b/src/rdf4cpp/rdf/datatypes/registry/util/DateTimeUtils.hpp
@@ -17,7 +17,7 @@ ResultType parse_date_time_fragment(std::string_view &s) {
     if constexpr (Separator != '\0') {
         auto p = s.find(Separator, 1);
         if (p == std::string::npos)
-            throw std::invalid_argument(std::format("{} parse error: missing {}", datatype, Separator));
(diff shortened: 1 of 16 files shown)


bigerl:
What is the idea here? `std::invalid_argument` seems more specific than `std::runtime_error`. I think it would be better to handle `std::invalid_argument` instead of replacing it.

mcb5637:
As we just learned, `std::invalid_argument` breaks existing code. And finding all instances of me throwning it is a lot easier/quicker than finding every instance where a `std::runtime_error` is cought, but a `std::invalid_argument` is not.

liss-h (changes requested):
Please add a test case that covers the specific error from the issue

liss-h (approved):

PR #218, opened 2023-10-19
Feature: Faster serialization and deserialization interfaces

liss-h:
Makes the usage of `std::iostream`s optional and adds serialization without heap allocation.


liss-h:
No idea what that GCC warning is about

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
```diff
@@ -34,14 +34,14 @@ class RDFFileParser {
     explicit RDFFileParser(std::string &&file_path, ParsingFlags flags = ParsingFlags::none(),
                            storage::node::NodeStorage node_storage = storage::node::NodeStorage::default_instance());
 
-    class Iterator {
+    class iterator {
         friend class RDFFileParser;
-        friend bool operator==(const RDFFileParser::Iterator &iter, std::default_sentinel_t) noexcept;
-        std::unique_ptr<std::ifstream> stream_;
+        friend bool operator==(const RDFFileParser::iterator &iter, std::default_sentinel_t) noexcept;
+        std::unique_ptr<FILE, decltype([](FILE *f) { fclose(f); })> stream_;
```
bigerl: `FILE` should actually not be available here but only `std::FILE`. Maybe this is what confuses gcc?
- https://en.cppreference.com/w/cpp/header/cstdio
- https://en.cppreference.com/w/cpp/io/c/FILE

bigerl (changes requested):
just had a quick look

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
liss-h: No, thats not it. (And yes technically only std::FILE should be available but libstdc++ exposes everything from the c headers in the global namespace as well. Probably because they don't want to annoy people)

liss-h: Removed the unique_ptr, gcc seems to not like in in combination with `FILE`

liss-h:
Benchmark results:
Benchmark: load dbpedia2015 (not cleaned) from a file into a `rdf4cpp::rdf::Dataset` and then serialize the dataset back out to a file.

develop: 78.1777 MB/s deserializing, 124.281 MB/s serializing
this branch: 81.4333 MB/s deserializing, 143.108 MB/s serializing

Better than nothing, but kinda meh

tests/parser/bench_RDFFileParser.cpp
```diff
@@ -0,0 +1,67 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include <doctest/doctest.h>
+
+#include <rdf4cpp/rdf.hpp>
+
+#include <chrono>
+#include <filesystem>
+
+TEST_CASE("Benchmark") {
```
bigerl: I would suggest https://github.com/martinus/nanobench for reliable numbers

bigerl:
> Benchmark results: Benchmark: load dbpedia2015 (not cleaned) from a file into a `rdf4cpp::rdf::Dataset` and then serialize the dataset back out to a file.
> 
> develop: 78.1777 MB/s deserializing, 124.281 MB/s serializing this branch: 81.4333 MB/s deserializing, 143.108 MB/s serializing
> 
> Better than nothing, but kinda meh

As reference point, pure serd deserialization happening under the hood would be good so we know where the sonic wall is that we cannot break through without replacing the backend.

liss-h:
New results:

develop:
```
|             ns/SWDF |              SWDF/s |    err% |        ins/SWDF |        cyc/SWDF |    IPC |       bra/SWDF |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|----------------:|-------:|---------------:|--------:|----------:|:----------
|      348,421,110.00 |                2.87 |    0.2% |5,589,615,019.00 |1,015,589,094.00 |  5.504 |1,159,627,002.00 |    0.8% |      3.84 | `deserialization`
|       85,897,593.00 |               11.64 |    0.6% |  917,672,728.00 |  221,854,932.00 |  4.136 | 181,551,670.00 |    0.4% |      0.96 | `serialization`
```

This branch:
```
|             ns/SWDF |              SWDF/s |    err% |        ins/SWDF |        cyc/SWDF |    IPC |       bra/SWDF |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|----------------:|-------:|---------------:|--------:|----------:|:----------
|      308,382,796.00 |                3.24 |    0.3% |4,872,342,245.00 |  897,918,138.00 |  5.426 | 996,055,847.00 |    0.9% |      3.40 | `deserialization`
|       71,300,434.00 |               14.03 |    1.1% |  711,314,709.00 |  179,003,682.00 |  3.974 | 141,446,458.00 |    0.5% |      0.79 | `serialization`
```

Still meh

liss-h:
Fiddle around with the serialization interface a bunch more, now we get:
```
|             ns/SWDF |              SWDF/s |    err% |        ins/SWDF |        cyc/SWDF |    IPC |       bra/SWDF |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|----------------:|-------:|---------------:|--------:|----------:|:----------
|      309,859,986.36 |                3.23 |    0.3% |4,872,405,401.08 |  902,468,480.73 |  5.399 | 996,053,280.30 |    1.0% |     37.35 | `deserialization`
|       37,519,710.40 |               26.65 |    0.3% |  419,170,423.82 |  112,092,856.20 |  3.739 |  71,644,882.55 |    0.4% |      4.51 | `serialization`
```

liss-h:
TODO:
- [x] add tests

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    constexpr void repoint(char *new_data, size_t new_size) noexcept {
+        data_ = new_data;
+        size_ = new_size;
+    }
+};
+
+template<typename T>
+concept Serializer = requires (T &ser) {
```
bigerl: This one also needs documentation. As a standalone concept it is no clear how it can be a serializer (e.g., doesn't have a write method)
Would be good to have concept documentation and also to cover the funcitons it has.

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+    typename T::Output;
+    typename T::Buffer;
+
+    { ser.buffer() } -> std::same_as<typename T::Buffer &>;
+    { ser.cursor() } -> std::same_as<Cursor &>;
+    { ser.finalize() } -> std::convertible_to<typename T::Output>;
+    T::flush(&ser.buffer(), ser.cursor());
```
bigerl: why is flush static and not a method or member?

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+};
+
+/**
+ * Flushes user defined data
+ */
+using FlushFunc = void (*)(void *buffer, Cursor &cursor) noexcept;
```
bigerl: from the definition it's not clear what the relation between buffer and curser is.

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * Serializes the given string as is
+ * See Node::serialize for more details on the signature/usage
+ */
+inline bool serialize_str(std::string_view str, void *const buffer, Cursor &cursor, FlushFunc const flush) noexcept {
```
bigerl: please document args

src/rdf4cpp/rdf/Node.hpp
```diff
@@ -68,6 +71,36 @@ class Node {
      */
     [[nodiscard]] static Node make_null() noexcept;
 
+    /**
+     * Serialize the string representation of the given node in N-format as defined by <a href="https://www.w3.org/TR/n-triples/">N-Triples</a> and <a href="https://www.w3.org/TR/n-quads/">N-Quads</a>.
+     *
+     * Takes as input an arbitrary, flushable/extendable buffer type (e.g. a std::string) a cursor
```
bigerl: , a cursor

src/rdf4cpp/rdf/Node.hpp
```diff
@@ -68,6 +71,36 @@ class Node {
...
+     * into the buffer (that describes its base address and length) and a function to flush the buffer (i.e. to either flush to io or extend the buffer).
```
bigerl: > that describes its base address and length

Doesn't it describe the current position in the buffer (I think that is what cursor means)?

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -158,10 +158,10 @@ class Literal : public Node {
      *      which prevents the template version of datatype_eq from using IRIs
      */
     [[nodiscard]] bool dynamic_datatype_eq_impl(std::string_view datatype) const noexcept;
-protected:
-    explicit Literal(Node::NodeBackendHandle handle) noexcept;
 
 public:
+    explicit Literal(Node::NodeBackendHandle handle) noexcept;
```
bigerl, line 162: why is this one required to be public now?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -609,75 +609,67 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
     return Literal::make_boolean(this->language_tag_eq(other), node_storage);
 }
 
-Literal::operator std::string() const noexcept {
+// https://www.w3.org/TR/n-triples/#grammar-production-STRING_LITERAL_QUOTE
+#define RDF4CPP_DETAIL_TRY_SER_QUOTED_LEXICAL(lexical) \
```
bigerl, line 613: why a macro and not a function or lambda?

private/rdf4cpp/rdf/TrySerialize.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef RDF4CPP_PRIVATE_TRY_SERIALIZE_HPP
+#define RDF4CPP_PRIVATE_TRY_SERIALIZE_HPP
+
+#include <rdf4cpp/rdf/Serialize.hpp>
+
+#define RDF4CPP_DETAIL_TRY_SER(...)                             \
```
bigerl: This is a macro so it can return, right?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -733,17 +724,24 @@ Literal::operator std::string() const noexcept {
...
+                    RDF4CPP_DETAIL_TRY_SER(lexical_form);
+                    RDF4CPP_DETAIL_TRY_SER("\"^^<");
+                    RDF4CPP_DETAIL_TRY_SER(datatype_iri);
+                    RDF4CPP_DETAIL_TRY_SER(">");
+                    return true;
                 });
     }
+}
+
+#undef RDF4CPP_DETAIL_TRY_SER_QUOTED_LEXICAL
```
bigerl, line 737: haven't seen that so far. But it's a very good idea!

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+#include <array>
+#include <cstdio>
+#include <cstring>
+#include <ostream>
+#include <string>
+#include <string_view>
+#include <system_error>
+
+namespace rdf4cpp::rdf {
```
bigerl: I would move it to rdf4cpp::rdf::serialization/serializer

src/rdf4cpp/rdf/Serialize.hpp
bigerl: It's rather a BufferedWriter than a Serializer. Different serializers would be for ntriple, turtle, rdfxml, etc.

src/rdf4cpp/rdf/Serialize.hpp
bigerl: name: write_str

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+    }
+    return true;
+}
+
+/**
+ * Serializes a string as is
+ * See Node::serialize<> for more details on the signature/usage
+ */
+template<Serializer Ser>
+bool serialize_str(std::string_view str, Ser &ser) {
```
bigerl: write_str

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+    return serialize_str(str, &ser.buffer(), ser.cursor(), &Ser::flush);
+}
+
+/**
+ * (Optional) base class for serializers.
+ * Encapsulates a cursor and some buffer type and makes the implementation of flush less error prone.
+ */
+template<typename CRTP, typename Buffer>
+struct SerializerBase {
```
bigerl: BufferedWriterBase

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+template<Serializer Ser>
+bool serialize_str(std::string_view str, Ser &ser) {
+    return serialize_str(str, &ser.buffer(), ser.cursor(), &Ser::flush);
+}
+
+/**
+ * (Optional) base class for serializers.
+ * Encapsulates a cursor and some buffer type and makes the implementation of flush less error prone.
+ */
+template<typename CRTP, typename Buffer>
```
bigerl: Add doc for template args

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+        CRTP::flush_impl(*reinterpret_cast<typename CRTP::Buffer *>(buffer), cursor);
+    }
+};
+
+using StringBuffer = std::string;
+
+/**
+ * A serializer that serializes to a std::string
+ */
+struct StringSerializer : SerializerBase<StringSerializer, StringBuffer> {
```
bigerl: BufferedStringWriter

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+    static void flush_impl(Buffer &buffer, Cursor &cursor) noexcept {
+        auto const bytes_written = cursor.data() - buffer.data();
+
+        buffer.resize(buffer.size() * 2);
+        cursor.repoint(buffer.data() + bytes_written,
+                       cursor.size() + buffer.size());
+    }
+};
+
+struct CFileBuffer {
```
bigerl: That it is the C file interface is a implementation detail. I would just call it BufferedFile

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+    explicit constexpr CFileBuffer(FILE *file) noexcept : file_{file} {}
+};
+
+/**
+ * A serializer that serializes to a C FILE
+ * It is advisable to set the internal file buffer to nullptr
+ * because this serializer has an internal buffer
+ */
+struct CFileSerializer : SerializerBase<CFileSerializer, CFileBuffer> {
```
bigerl: BufferedFileWriter

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+    void finalize() {
+        auto const to_write = static_cast<size_t>(cursor().data() - buffer().buffer_.data());
+        if (fwrite(buffer().buffer_.data(), 1, to_write, buffer().file_) < to_write) {
+            throw std::system_error{std::error_code{errno, std::system_category()}};
+        }
+    }
+
+    static void flush_impl(Buffer &buffer, Cursor &cursor) noexcept {
+        auto const bytes_written = fwrite(buffer.buffer_.data(), 1, static_cast<size_t>(cursor.data() - buffer.buffer_.data()), buffer.file_);
```
bigerl: If bytes are available and bytes_written does repeatedly not write this should throw. 
Otherwise we are locked in an infinite loop when the disk is full.

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+        cursor.repoint(cursor.data() - bytes_written,
+                       cursor.size() + bytes_written);
+    }
+};
+
+struct OStreamBuffer {
```
bigerl: BufferedStl OStream

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+    explicit constexpr OStreamBuffer(std::ostream &os) noexcept : os_{&os} {}
+};
+
+/**
+ * A serializer that serializes to an std::ostream.
+ * It is advisable to set the associated stream buffer to nullptr
+ * because this serializer has an internal buffer.
+ */
+struct OStreamSerializer : SerializerBase<OStreamSerializer, OStreamBuffer> {
```
bigerl: StlOStreamWriter

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    void finalize() {
+        if (!buffer().os_->write(buffer().buffer_.data(), static_cast<std::streamsize>(cursor().data() - buffer().buffer_.data()))) {
+            throw std::system_error{std::error_code{errno, std::system_category()}};
+        }
+    }
+
+    static void flush_impl(Buffer &buffer, Cursor &cursor) noexcept {
+        if (!buffer.os_->write(buffer.buffer_.data(), static_cast<std::streamsize>(cursor.data() - buffer.buffer_.data()))) {
```
bigerl: similar to writing to a file we need to somehow handle if we repeatedly cannot write to os_

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
```diff
@@ -34,16 +34,18 @@ class RDFFileParser {
...
-        friend bool operator==(const RDFFileParser::Iterator &iter, std::default_sentinel_t) noexcept;
-        std::unique_ptr<std::ifstream> stream_;
+        friend bool operator==(const RDFFileParser::iterator &iter, std::default_sentinel_t) noexcept;
+        FILE *stream_;
         std::unique_ptr<IStreamQuadIterator> iter_;
 
-        Iterator();
-        Iterator(std::ifstream &&stream, ParsingFlags flags, const storage::node::NodeStorage &node_storage);
+        iterator();
+        iterator(FILE *stream, ParsingFlags flags, const storage::node::NodeStorage &node_storage);
```
bigerl: rename stream to file
who is responsible to close it? Please add a docstring that documents that

src/rdf4cpp/rdf/parser/IStreamQuadIterator.hpp
```diff
@@ -66,13 +77,20 @@ struct IStreamQuadIterator {
...
      */
-    IStreamQuadIterator(std::default_sentinel_t) noexcept;
+    explicit IStreamQuadIterator(std::default_sentinel_t) noexcept;
 
     IStreamQuadIterator(IStreamQuadIterator const &) = delete;
     IStreamQuadIterator(IStreamQuadIterator &&) noexcept;
 
     IStreamQuadIterator &operator=(IStreamQuadIterator &&) noexcept = default;
 
+    IStreamQuadIterator(void *stream,
```
bigerl, line 112: needs a docstring

src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
```diff
@@ -3,6 +3,27 @@
 
 namespace rdf4cpp::rdf::parser {
 
+/**
+ * Adaptor function so that serd can read from std::istreams.
+ * Matches the interface of SerdSource
+ */
+static size_t istream_read(void *buf, [[maybe_unused]] size_t elem_size, size_t count, void *voided_self) noexcept {
```
bigerl, line 15: please add documentation for the args

tests/serializer/tests_Serialize.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include <filesystem>
+#include <format>
+#include <random>
+
+using namespace rdf4cpp::rdf;
+
+TEST_SUITE("Serialize") {
+    template<typename NodeT>
+    void run_ser_test(NodeT node) {
+        auto expected = std::string{node};
```
bigerl: This tests the serializer by comparing with the serializer because std::string also uses a serializer. Expected should be hardcoded.

src/rdf4cpp/rdf/Serialize.hpp
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * Memory layout is guaranteed to be identical to:
+ * @code
+ * struct Cursor {
+ *      char *data;
+ *      size_t size;
+ * };
+ * @endcode
+ */
+struct Cursor {
```
bigerl: maybe directly call it BufferCurser?

bigerl (changes requested):
thx for the PR. 28 comments may seem devastating but I actually like the design. :) 
It's mainly naming, documentation and smaller things.

src/rdf4cpp/rdf/Serialize.hpp
liss-h: Because that makes it way easier to map to C. I don't think you can cast a normal function pointer to a member function pointer

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
liss-h: Added docstring, will keep name stream because thats also how C names objects of type `FILE *`

src/rdf4cpp/rdf/Serialize.hpp
liss-h: That is already handled, the write functions return when the cursor size after flushing is still zero.
See: https://github.com/rdf4cpp/rdf4cpp/blob/19759736c57df779fbde04ade045717e8d00e2ef/src/rdf4cpp/rdf/Serialize.hpp#L75-L77

src/rdf4cpp/rdf/Serialize.hpp
liss-h: No, we are not. If its incorrectly implemented we would be. Increasing the buffer space by 0 bytes is allowed.

https://github.com/rdf4cpp/rdf4cpp/blob/19759736c57df779fbde04ade045717e8d00e2ef/src/rdf4cpp/rdf/Serialize.hpp#L75-L77

src/rdf4cpp/rdf/Serialize.hpp
liss-h: I mean, its not really since you have to pass in `FILE *`. I would also keep it that way because then error handling can be done however the user wants.

src/rdf4cpp/rdf/Node.hpp
liss-h: Yes, base address as in base addr of the free segment.

src/rdf4cpp/rdf/Literal.hpp
liss-h: Left in from previous attempt, reverting

private/rdf4cpp/rdf/TrySerialize.hpp
liss-h: It returns from the function its called from, the only purpose of this is that i don't have to write the contents of this macro on every single call.

liss-h:
Question: do we want to remove the things in rdf/writer or rewrite them?

src/rdf4cpp/rdf/Serialize.hpp
liss-h: I think cursor is unambiguous, esp if we move this stuff into a subnamespace

src/rdf4cpp/rdf/Literal.cpp
liss-h: Because it composes better, otherwise I would have to do explicit error checking on every single call

src/rdf4cpp/rdf/Serialize.hpp
liss-h: Also making it a member function doesn't work in rust because you would need to mutably borrow members of self multiple times, which is illegal even in unsafe code.

(first mutable borrow when passing cursor to function, second mutable borrow when passing self to the voided self param)

src/rdf4cpp/rdf/Node.hpp
liss-h: added note about it being base of free segment

src/rdf4cpp/rdf/Serialize.hpp
liss-h: Moved to rdf4cpp::rdf::writer

src/rdf4cpp/rdf/Serialize.hpp
liss-h: Now called BufWriter

src/rdf4cpp/rdf/Serialize.hpp
liss-h: BufWriterBase

src/rdf4cpp/rdf/Serialize.hpp
liss-h: StringWriter because its not buffered

src/rdf4cpp/rdf/Serialize.hpp
liss-h: BufCFileWriter

src/rdf4cpp/rdf/Serialize.hpp
Clueliss: BufOStreamWriter

src/rdf4cpp/rdf/Serialize.hpp
liss-h: BufOStreamWriter

liss-h:
> Question: do we want to remove the things in rdf/writer or rewrite them?

Replaced serialization interfaces for Graph and Dataset

src/rdf4cpp/rdf/Serialize.hpp
bigerl: OK, I am convinced :)

src/rdf4cpp/rdf/Node.hpp
```diff
@@ -7,7 +7,7 @@
  */
 
 
-#include <rdf4cpp/rdf/Serialize.hpp>
+#include "rdf4cpp/rdf/writer/BufWriter.hpp"
```
bigerl: should be in `<>`

src/rdf4cpp/rdf/writer/BufWriter.hpp
```diff
@@ -43,27 +43,69 @@ struct Cursor {
...
-    typename T::Buffer;
-
-    { ser.buffer() } -> std::same_as<typename T::Buffer &>;
-    { ser.cursor() } -> std::same_as<Cursor &>;
-    { ser.finalize() } -> std::convertible_to<typename T::Output>;
-    T::flush(&ser.buffer(), ser.cursor());
+/**
+ * A type that can be used for buffered output.
+ * Conceptually, it is three things.
+ * Some buffer-like object a cursor into that buffer and a way to flush the buffer somewhere to make some room.
```
bigerl:
```suggestion
 * Some buffer-like object, a cursor into that buffer, and a way to flush the buffer somewhere to make some room.
```

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
```diff
@@ -41,7 +41,10 @@ class RDFFileParser {
         std::unique_ptr<IStreamQuadIterator> iter_;
 
         iterator();
-        iterator(FILE *stream, ParsingFlags flags, const storage::node::NodeStorage &node_storage);
+        /**
+         * Constructs an iterator by taking ownership of the given file
+         */
+        iterator(FILE *&&stream, ParsingFlags flags, const storage::node::NodeStorage &node_storage);
```
bigerl: why the rvalue for stream? It's a pointer already. 

Did you mean `FILE &&stream`?

src/rdf4cpp/rdf/writer/BufWriter.hpp
```diff
@@ -43,27 +43,69 @@ struct Cursor {
...
+
+    /**
+     * The underlying buffer, the only thing this need to be able to do is be used in flush
+     */
+    { bw.buffer() } -> std::same_as<typename W::Buffer &>;
+
+    /**
+     * The cursor pointing inside of w.buffer()
+     */
+    { bw.cursor() } -> std::same_as<Cursor &>;
```
bigerl, line 77: I am missing a function that allows to write with a BufWriter. That will confuse people. 

it could look like:
```
pair<size_t, bool> write(void* in_buf, size_t len)
```

and return the number of bytes written and if the buffer is exhausted (flush is required). 
Alternatively, it could also take care of the flushes. 

I am not sure where to actually implement it. The best spot would probably be cursor with code similar to that one:
```
auto const max_write = std::min(str.size(), cursor.size());
        memcpy(cursor.data(), str.data(), max_write);
        cursor.advance(max_write);
```

(from write_str below)

src/rdf4cpp/rdf/query/Variable.cpp
```diff
@@ -1,5 +1,5 @@
+#include "rdf4cpp/rdf/writer/TryWrite.hpp"
```
bigerl: `<>`

please check this through out the code base.

bigerl (changes requested):
one major thing, otherwise minor stuff

src/rdf4cpp/rdf/parser/RDFFileParser.hpp
liss-h: mostly to make it clear that this function takes ownership. This means you must `std::move` the FILE * into the iterator, this doesn't make a difference, its just for making it really visually clear on the callsite that this is the behaviour

src/rdf4cpp/rdf/query/Variable.cpp
liss-h: CLion always does this when renaming files.. really annoying

src/rdf4cpp/rdf/writer/BufWriter.hpp
liss-h: Well that function is basically `write_str`. Should I just mention it in the doc?

bigerl: yes that sounds good.

src/rdf4cpp/rdf/BlankNode.cpp
```diff
@@ -1,4 +1,5 @@
-#include "BlankNode.hpp"
```
bigerl, line 1: Sorry, I didn't make that what I wanted with regards to includes. 
I would suggest to use "" only in .cpp files and there only for the header that declares what is defined in the .cpp file. Use a relative path here, meaning for in my.cpp sth. like `#include "my.hpp"`.

loosely based on http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else

The main reason why we are using <> everywhere in public headers is that I had some problems with relative paths and CMake at some point and looking for a absolute path in the includes is very solid and worked always, so far.

tests/serializer/tests_Serialize.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <format>
+#include <random>
+
+using namespace rdf4cpp::rdf;
+
+TEST_SUITE("Serialize") {
+    template<typename NodeT>
+    void run_ser_test(NodeT node, std::optional<std::string> expected = std::nullopt) {
+        if (!expected.has_value()) {
+            expected = std::string{node};
```
bigerl, line 15: I checked for the types how `::operator std::string()` is implemented:
- IRI backend_view writes to preallocated string 
- BNode backend_view writes to preallocated string 
- Literal uses serialize()
- Variable backend_view writes to preallocated string 

Literal is tested with custom input. For the others, we are sure that the write function is correct, I think. 

It also makes the test sensitive to any of them being changed. So, I would say it is good now. :)

tests/bench_SerDe.cpp
```diff
@@ -0,0 +1,90 @@
...
+    writer::BufCFileWriter ser{out_file};
+
+    for (auto const &quad : ds) {
+        quad.subject().serialize(ser);
+        write_str(" ", ser);
+        quad.predicate().serialize(ser);
+        write_str(" ", ser);
+        quad.object().serialize(ser);
+        write_str(" .\n", ser);
+    }
```
bigerl:
```suggestion
    ds.serialize(ser);
```

bigerl (changes requested):

src/rdf4cpp/rdf/BlankNode.cpp
liss-h: Should I revert it or not? Its at consistent now at least, and probably works right?

bigerl: It is but I would feel safer with that in place (it's probably only reverting some changes):
> I would suggest to use "" only in .cpp files and there only for the header that declares what is defined in the .cpp file. Use a relative path here, meaning for in my.cpp sth. like #include "my.hpp".

I would also expect it to work as it is right now but I have never seen it done like this elsewhere and that makes me a bit nervous.

bigerl (approved):
Puh. That's a lot of files. thx for your patience going through them.

PR #220, opened 2023-10-27
Feature: Unsync node storage backend

liss-h:
Implements an alternative, non-synchronized node storage backend.


src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/UnsyncNodeTypeStorage.hpp
```diff
@@ -1,22 +1,20 @@
...
+#include <rdf4cpp/rdf/storage/node/identifier/NodeID.hpp>
 
 #include <memory>
-#include <shared_mutex>
-
 
 namespace rdf4cpp::rdf::storage::node::reference_node_storage {
 
 /**
  * Storage for one of the Node Backend types. Includes a shared mutex to synchronize access and bidirectional mappings between the Backend type and identifier::NodeID.
```
nkaralis: the shared mutex is available only in SyncNodeType, right?

src/rdf4cpp/rdf/storage/node/reference_node_storage/UnsyncReferenceNodeStorageBackend.cpp
```diff
@@ -0,0 +1,239 @@
...
+           });
+}
+
+bool UnsyncReferenceNodeStorageBackend::has_specialized_storage_for(identifier::LiteralType const datatype) const noexcept {
+    static constexpr auto specialization_lut = specialization_detail::make_storage_specialization_lut<decltype(specialized_literal_storage_)>();
+    return specialization_lut[datatype.to_underlying()];
+}
+
+/**
+ * Synchronized lookup (and creation) of IDs by a provided view of a Node Backend.
```
nkaralis: it is not synchronized here, right?

tests/nodes/tests_NodeStorage_specialization.cpp
```diff
@@ -10,6 +12,14 @@ using namespace rdf4cpp::rdf;
 using namespace storage::node;
 using namespace datatypes;
 
+int main(int argc, char **argv) {
+    NodeStorage::set_default_instance(NodeStorage::new_instance<reference_node_storage::SyncReferenceNodeStorageBackend>());
+    doctest::Context{argc, argv}.run();
+
+    NodeStorage::set_default_instance(NodeStorage::new_instance<reference_node_storage::UnsyncReferenceNodeStorageBackend>());
+    doctest::Context{argc, argv}.run();
```
nkaralis, line 20: do the tests run twice (once for each reference impl)?

nkaralis (changes requested):
The PR looks good. Only a few comments need to be addressed.

Question: Could the "syncness" have been implemented using the previous implementation by adding an additional template parameter and `if constexpr` at the appropriate places?

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/UnsyncNodeTypeStorage.hpp
liss-h: yes, is unused for UnsyncNodeTypeStorage

src/rdf4cpp/rdf/storage/node/reference_node_storage/UnsyncReferenceNodeStorageBackend.cpp
liss-h: yes, copy paste error

tests/nodes/tests_NodeStorage_specialization.cpp
liss-h: Yes, was the easiest way to have some tests without rewriting much of the test code

liss-h: Do you think the tests are sufficient, or should I also run some other test suites with both?

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/UnsyncNodeTypeStorage.hpp
nkaralis: the doc string needs to be updated. it seems that i did not highlight the right line.

tests/nodes/tests_NodeStorage_specialization.cpp
nkaralis: I think they are sufficient. The synced one, which is the tricky one, is tested in `tests_NodeStorage_lifetime`.

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/UnsyncNodeTypeStorage.hpp
liss-h: Ah no I just didn't see, fixed now

tests/nodes/tests_NodeStorage_specialization.cpp
liss-h: That specific test actually has nothing to do with the backend. It tests that the frontend (`NodeStorage`) is properly synchronized

nkaralis: I was referring to the synchronization. In this one, i think there are enough test cases, which are executed for both. I guess the non-specialized literal backend interfaces are not tested using both. Could we add a similar main function in the other files of `tests/nodes`?

liss-h: I looked through the tests and I think it only makes sense for `tests_Node`, `tests_Literal` and `tests_NodeStorage_specialization. So all 3 now run with both backends

nkaralis (approved):

PR #223, opened 2023-11-14
Fix: Serialization interfaces

liss-h:
Makes serialization interfaces actually FFI friendly.
The original design did not consider that references are not FFI safe.


bigerl (approved):
thx for the PR. Change is simple, tests complete successfully. This looks good :)

PR #225, opened 2023-11-20
Fix: nonstd/expected include and CI

liss-h:
Various fixes that prevented building (with) rdf4cpp

1. Forces the use of nonstd expected implementation instead of `std::expected` because the APIs are different
2. Fixes overcaching in the CI that would cause it to fail e.g. due to dependencies having (unresolved) coverage symbols in them even when not building with coverage (happens if the coverage workflow finishes first and populates the cache)
3. Add clang 17 to CI

### DIFF
--- a/.github/workflows/reusable_fetch_content.yml
+++ b/.github/workflows/reusable_fetch_content.yml
@@ -28,12 +28,15 @@ jobs:
           # gcc-13
           sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
 
-          # clang-16
           source /etc/os-release
-          echo "deb http://apt.llvm.org/${UBUNTU_CODENAME}/ llvm-toolchain-${UBUNTU_CODENAME}-16 main" | sudo tee /etc/apt/sources.list.d/llvm-16.list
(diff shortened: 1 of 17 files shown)


.github/workflows/reusable_fetch_content.yml
```diff
@@ -23,17 +23,20 @@ jobs:
         shell: bash
 
     steps:
-      - name: Add Repos for for gcc-13 and clang-16
+      - name: Add Repos for for gcc-13 and clang-16,17
         run: |
           # gcc-13
           sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+          
+          source /etc/os-release
```
bigerl, line 31: It's the first time I noticed this. Whatis it good for? what is in os-release?

bigerl (approved):
Looks good. Makes a bunch of things cleaner.

.github/workflows/reusable_fetch_content.yml
liss-h: The value for `${UBUNTU_CODENAME}` thats used below to select the correct repo

PR #227, opened 2023-11-27
Node::find

### LINKED ISSUE #226: Optimization for usecases of `try_get_in_node_storage`
Currently code that uses `try_get_in_node_storage` usually has this usage pattern:

```c++
IRI iri = IRI::make("some-iri", temp_node_storage);
...Later
IRI iri_for_use = iri.try_get_in_node_storage(actual_node_storage);
```

This constitutes two relatively expensive node storage accesses.
It could be improved by adding new interfaces to rdf4cpp:
```c++
// returns null if not exists, uses NodeStorage::find_id internally
IRI iri_for_use = IRI::find("some-iri", actual_node_storage); 
```

This would turn the two accesses into a single one.

mcb5637:
close #226

### DIFF
--- a/src/rdf4cpp/rdf/Literal.hpp
+++ b/src/rdf4cpp/rdf/Literal.hpp
@@ -126,6 +126,8 @@ private:
      */
     [[nodiscard]] static Literal make_lang_tagged_unchecked(std::string_view lexical_form, bool needs_escape, std::string_view lang, NodeStorage &node_storage) noexcept;
 
+    [[nodiscard]] static Literal make_lang_tagged_unchecked_from_node_id(std::string_view lang, const Node::NodeStorage &node_storage, storage::node::identifier::NodeID node_id) noexcept;
+
     /**
      * Creates a string like type with contents of str.
      * Will either create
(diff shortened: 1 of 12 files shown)


liss-h (changes requested):
Sorry the issue wasn't formulated very well. This is meant as a function for all node types.

So also Literal::find, BNode::find and so on

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -334,6 +336,85 @@ class Literal : public Node {
...
+     * searches for a Literal of type T in the specified node storage and returns it.
+     * returns a null Literal, if not found.
+     * if T is inlineable (and not rdf::LangString) always returns the inlined Literal.
+     * @tparam T
+     * @param compatible_value
+     * @param node_storage
+     * @return
+     */
+    template<datatypes::LiteralDatatype T>
+    [[nodiscard]] static Literal find(typename T::cpp_type const &compatible_value,
```
liss-h: Missing templated version that takes `string_view` (== `find_typed`)
+ rename this to `find_typed_from_value`

tests/nodes/tests_Node.cpp
```diff
@@ -200,3 +200,13 @@ TEST_CASE("IRI UUID") {
     CHECK(uuid != uuid2);  // note: non-deterministic but should basically never fail
     CHECK(regex::Regex{"^urn:uuid:[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"}.regex_match(uuid.identifier()));
 }
+
+TEST_CASE_TEMPLATE("IRI/BlankNode::find", T, IRI, BlankNode) {
```
liss-h: This test case will fail if we ever implement bnode validation (I think). Please seperate the 2 cases

liss-h (changes requested):
Implementation looks correct to me, just 2 things missing
1. the additional function in literal (added a seperate comment for that)
2. `query::Variable::find`

liss-h:
Also please fix the branch package workflow, it seems that it cannot handle branche names with uppercase letters

tests/nodes/tests_Node.cpp
mcb5637: I assume separating the data of the 2 is enough.
If not, we might want to wait until we know what bnode validation looks like.

mcb5637:
please only merge, when https://github.com/dice-group/cpp-conan-release-reusable-workflow/pull/17 is merged into main and the CI points to main again

liss-h (approved):
TODO:
- [x] point workflow to main when merged

PR #228, opened 2023-12-04
Fix: parse to non-default node storage and borrow optimization from serd

liss-h:


src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
```diff
@@ -15,7 +21,7 @@ namespace rdf4cpp::rdf::parser {
 static size_t istream_read(void *buf, [[maybe_unused]] size_t elem_size, size_t count, void *voided_self) noexcept {
     assert(elem_size == 1);
 
-    auto *self = reinterpret_cast<std::istream *>(voided_self);
+    auto *self = static_cast<std::istream *>(voided_self);
```
bigerl, line 24: Just read-up upon this. Seems much better to use static_cast everywhere to cast from voided to concrete pointers.

src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
```diff
@@ -47,17 +53,15 @@ IStreamQuadIterator::IStreamQuadIterator(void *stream,
                                          ReadFunc read,
                                          ErrorFunc error,
                                          flags_type flags,
-                                         state_type *state,
-                                         storage::node::NodeStorage node_storage) noexcept
-    : impl{std::make_unique<Impl>(stream, read, error, flags, state, std::move(node_storage))} {
+                                         state_type *state) noexcept
```
bigerl, line 56: I'll use this spot to phrase the general problem: the configuration of the node_storage was moved to the state at some point. But the constructor parameter remained and was still set. So, even when it was set nodes were written to the default node storage. 

Correct?

src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
```diff
@@ -97,4 +101,19 @@ bool IStreamQuadIterator::operator!=(std::default_sentinel_t) const noexcept {
...
 
+FILE *fopen_fastseq(char const *path, char const *mode) noexcept {
+    // inspired by <serd/system.c> (serd_fopen)
+
+    FILE *fd = fopen(path, mode);
+    if (fd == nullptr) {
+        return fd;
+    }
+
+#if _POSIX_C_SOURCE >= 200112L
```
bigerl: the header guard `#if __has_include(<fcntl.h>)` and this should have the same condition. Safest would be to use the conjunction.

src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
```diff
@@ -97,4 +101,19 @@ bool IStreamQuadIterator::operator!=(std::default_sentinel_t) const noexcept {
...
+    (void) posix_fadvise(fileno(fd), 0, 0, POSIX_FADV_SEQUENTIAL | POSIX_FADV_NOREUSE);
```
bigerl: We could even think about adding `POSIX_FADV_WILLNEED` (non-blocking)

src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
```diff
@@ -97,4 +101,19 @@ bool IStreamQuadIterator::operator!=(std::default_sentinel_t) const noexcept {
     return !this->is_at_end();
 }
 
+FILE *fopen_fastseq(char const *path, char const *mode) noexcept {
```
bigerl, line 104: Did you measure the difference? (non-blocking)

bigerl (changes requested):
thx, looks good.  Just some clarifications in minor things.

src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
liss-h: Yes

src/rdf4cpp/rdf/parser/IStreamQuadIterator.cpp
liss-h: No I was just looking up how serd opens files

liss-h: Seemed like a potentially useful addition

bigerl (approved):

PR #230, opened 2023-12-11
dbpedia mode

### LINKED ISSUE #214: Feature: DBpedia mode for literal parsing
DBpedia is important to support but it's literals are messy. 

How to handle it: a DBpedia mode
- [ ] extract all literals from DBpedia (plain text, not normalized or anything)
- [ ] add an relaxed-parsing flag, let's call it `dbpediamode`, that accepts all the wrong things in DBpedia's literals and makes them right.

mcb5637:
also disables the new IRI validation, because that one causes even more errors on loading

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,19 +248,6 @@ set_target_properties(rdf4cpp PROPERTIES
         CXX_STANDARD_REQUIRED ON
 )
 
-option(DBPEDIA_MODE "relaxed parsing for dbpedia" OFF)
-if (PROJECT_IS_TOP_LEVEL AND BUILD_TESTING)
-    message("dbpedia mode set to runtime (required by test cases)")
-elseif (DBPEDIA_MODE EQUAL RUNTIME)
-    message("dbpedia mode set to runtime")
-elseif (DBPEDIA_MODE)
-    message("dbpedia mode on")
-    add_compile_definitions(RDF4CPP_DBPEDIA_MODE=true)
-else ()
-    message("dbpedia mode off")
-    add_compile_definitions(RDF4CPP_DBPEDIA_MODE=false)
-endif ()
-
 if (PROJECT_IS_TOP_LEVEL)
     include(cmake/install_library.cmake)
     install_cpp_library(rdf4cpp src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.22)
-project(rdf4cpp VERSION 0.0.20)
+project(rdf4cpp VERSION 0.0.22)
 
 include(cmake/boilerplate_init.cmake)
 boilerplate_init()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,6 +248,19 @@ set_target_properties(rdf4cpp PROPERTIES
         CXX_STANDARD_REQUIRED ON
 )
 
+option(DBPEDIA_MODE "relaxed parsing for dbpedia" OFF)
+if (PROJECT_IS_TOP_LEVEL AND BUILD_TESTING)
+    message("dbpedia mode set to runtime (required by test cases)")
+elseif (DBPEDIA_MODE EQUAL RUNTIME)
+    message("dbpedia mode set to runtime")
+elseif (DBPEDIA_MODE)
+    message("dbpedia mode on")
+    add_compile_definitions(RDF4CPP_DBPEDIA_MODE=true)
+else ()
+    message("dbpedia mode off")
+    add_compile_definitions(RDF4CPP_DBPEDIA_MODE=false)
+endif ()
+
 if (PROJECT_IS_TOP_LEVEL)
     include(cmake/install_library.cmake)
     install_cpp_library(rdf4cpp src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.22)
-project(rdf4cpp VERSION 0.0.22)
+project(rdf4cpp VERSION 0.0.24)
 
 include(cmake/boilerplate_init.cmake)
 boilerplate_init()
--- a/docs/source/users_guide.rst
+++ b/docs/source/users_guide.rst
@@ -48,6 +48,7 @@ Relaxed Parsing Mode
 --------------------
 
 The setting :var:`rdf4cpp::rdf::datatypes::registry::relaxed_parsing_mode` disables IRI validity checks and allows rdf4cpp to automatically try to correct some faulty Literals.
+See the linked relaxed_parsing_mode for a full list of changes.
 Currently aimed at reducing loading errors with DBPedia, more might be added in future versions.
 
 (since rdf4cpp v0.0.24)
--- a/docs/source/users_guide.rst
+++ b/docs/source/users_guide.rst
@@ -43,3 +43,11 @@ Parsing Files
 The class :class:`rdf4cpp::rdf::parser::RDFFileParser` allows reading files containing rdf Statements and iterate over them.
 Supported Formats: Turtle, TriG, NTriple and NQuad.
 :class:`rdf4cpp::rdf::parser::IStreamQuadIterator` allows doing the same over arbitrary data streams.
+
+Relaxed Parsing Mode
+--------------------
+
+The setting :var:`rdf4cpp::rdf::datatypes::registry::relaxed_parsing_mode` disables IRI validity checks and allows rdf4cpp to automatically try to correct some faulty Literals.
+Currently aimed at reducing loading errors with DBPedia, more might be added in future versions.
+
+(since rdf4cpp v0.0.24)
--- a/src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.cpp
+++ b/src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.cpp
@@ -2,9 +2,7 @@
 
 namespace rdf4cpp::rdf::datatypes::registry {
 
-#ifndef RDF4CPP_DBPEDIA_MODE
-bool dbpedia_mode = false;
-#endif
+bool relaxed_parsing_mode = false;
 
 DatatypeRegistry::registered_datatypes_t &DatatypeRegistry::get_mutable() noexcept {
     static registered_datatypes_t registry_ = []() {
--- a/src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.cpp
+++ b/src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.cpp
@@ -2,6 +2,10 @@
 
 namespace rdf4cpp::rdf::datatypes::registry {
 
+#ifndef RDF4CPP_DBPEDIA_MODE
+bool dbpedia_mode = false;
+#endif
+
 DatatypeRegistry::registered_datatypes_t &DatatypeRegistry::get_mutable() noexcept {
     static registered_datatypes_t registry_ = []() {
         registered_datatypes_t r;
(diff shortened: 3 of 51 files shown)


liss-h:
Before I review this it would be interesting to know how much of a difference it makes.
For example: number of (dbpedia) triples accepted with and without this PR

mcb5637:
The dbpedia version i tested with has no errors remaining in dbpedia mode.
Ill have my test run again over night with it off, to give you a count of triples thrown out.

mcb5637:
Ok, its 109330 lines of error messages at an estimated 1 faulty triple per line.
Some more info: quick filtering with grep shows, that most (109249) of these fail iri vaidation.

liss-h:
Thats definitely an improvement, thank you

CMakeLists.txt
```diff
@@ -248,6 +248,19 @@ set_target_properties(rdf4cpp PROPERTIES
         CXX_STANDARD_REQUIRED ON
 )
 
+option(DBPEDIA_MODE "relaxed parsing for dbpedia" OFF)
```
liss-h: We probably won't need a compiletime switch. It will most probably always be set to runtime. Given how people usually handle data they probably don't want a seperate executable/compiled library for dbpedia.

tests/nodes/tests_DBpediaMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+TEST_CASE("integers") {
+    Literal u{};
+    datatypes::registry::dbpedia_mode = false;
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::NonNegativeInteger>("-5"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::PositiveInteger>("0"), std::runtime_error);
+    CHECK(u == Literal{});
+
+    datatypes::registry::dbpedia_mode = true;
+    CHECK(Literal::make_typed<datatypes::xsd::NonNegativeInteger>("-5") == Literal::make_typed<datatypes::xsd::NonNegativeInteger>("0"));
```
liss-h: This behaviour is problematic because it changes the existing data. It would be better to just discard it.

Even better would be to make it into an `xsd:integer` and keep the value, but as we already discussed this is a little more complicated so I will create an issue for that specific approach, no need to implement it here.

tests/nodes/tests_DBpediaMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::Date>("1742-0-0"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::Date>("1742-2-40"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::Time>("10:90:00"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::Time>("26:90:60"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::DateTime>("1742-2-40T26:90:60"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::DateTimeStamp>("1742-2-40T26:90:60Z"), std::runtime_error);
+    CHECK(u == Literal{});
+
+    datatypes::registry::dbpedia_mode = true;
+    CHECK(Literal::make_typed<datatypes::xsd::GDay>("---40") == Literal::make_typed<datatypes::xsd::GDay>("---31"));
```
liss-h: Same here, we don't really know what they meant by that. They probably just had a buggy datetime implementation that just added days together without checking if it makes sense that way.

But since this is gDay we don't really have a way to fix what they meant => discard

tests/nodes/tests_DBpediaMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+    CHECK(Literal::make_typed<datatypes::xsd::GMonth>("--40") == Literal::make_typed<datatypes::xsd::GMonth>("--12"));
+    CHECK(Literal::make_typed<datatypes::xsd::GMonthDay>("--40-40") == Literal::make_typed<datatypes::xsd::GMonthDay>("--12-31"));
+    CHECK(Literal::make_typed<datatypes::xsd::GMonthDay>("--2-40") == Literal::make_typed<datatypes::xsd::GMonthDay>("--2-29"));
+    CHECK(Literal::make_typed<datatypes::xsd::GYearMonth>("1742-0") == Literal::make_typed<datatypes::xsd::GYearMonth>("1742-1"));
+    CHECK(Literal::make_typed<datatypes::xsd::Date>("1742-0-0") == Literal::make_typed<datatypes::xsd::Date>("1742-1-1"));
+    CHECK(Literal::make_typed<datatypes::xsd::Date>("1742-2-40") == Literal::make_typed<datatypes::xsd::Date>("1742-2-28"));
```
liss-h: This one should be fixed by adding the additional days onto the last valid date.
I.e. 1742.2.28 + (40 - 28)days = 1742.3.12

Again under the assumtion that their datetime implementation just naively added some number of days onto the date

tests/nodes/tests_DBpediaMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::DateTime>("1742-2-40T26:90:60"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::DateTimeStamp>("1742-2-40T26:90:60Z"), std::runtime_error);
+    CHECK(u == Literal{});
+
+    datatypes::registry::dbpedia_mode = true;
+    CHECK(Literal::make_typed<datatypes::xsd::GDay>("---40") == Literal::make_typed<datatypes::xsd::GDay>("---31"));
+    CHECK(Literal::make_typed<datatypes::xsd::GMonth>("--40") == Literal::make_typed<datatypes::xsd::GMonth>("--12"));
+    CHECK(Literal::make_typed<datatypes::xsd::GMonthDay>("--40-40") == Literal::make_typed<datatypes::xsd::GMonthDay>("--12-31"));
+    CHECK(Literal::make_typed<datatypes::xsd::GMonthDay>("--2-40") == Literal::make_typed<datatypes::xsd::GMonthDay>("--2-29"));
+    CHECK(Literal::make_typed<datatypes::xsd::GYearMonth>("1742-0") == Literal::make_typed<datatypes::xsd::GYearMonth>("1742-1"));
```
liss-h: this one probably makes sense (can stay this way)

tests/nodes/tests_DBpediaMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::Time>("10:90:00"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::Time>("26:90:60"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::DateTime>("1742-2-40T26:90:60"), std::runtime_error);
+    CHECK_THROWS_AS(u = Literal::make_typed<datatypes::xsd::DateTimeStamp>("1742-2-40T26:90:60Z"), std::runtime_error);
+    CHECK(u == Literal{});
+
+    datatypes::registry::dbpedia_mode = true;
+    CHECK(Literal::make_typed<datatypes::xsd::GDay>("---40") == Literal::make_typed<datatypes::xsd::GDay>("---31"));
+    CHECK(Literal::make_typed<datatypes::xsd::GMonth>("--40") == Literal::make_typed<datatypes::xsd::GMonth>("--12"));
+    CHECK(Literal::make_typed<datatypes::xsd::GMonthDay>("--40-40") == Literal::make_typed<datatypes::xsd::GMonthDay>("--12-31"));
```
liss-h: This has the same issue as gDay. There is no way we can fix this properly without year information => discard

tests/nodes/tests_DBpediaMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+    CHECK(Literal::make_typed<datatypes::xsd::GMonthDay>("--2-40") == Literal::make_typed<datatypes::xsd::GMonthDay>("--2-29"));
+    CHECK(Literal::make_typed<datatypes::xsd::GYearMonth>("1742-0") == Literal::make_typed<datatypes::xsd::GYearMonth>("1742-1"));
+    CHECK(Literal::make_typed<datatypes::xsd::Date>("1742-0-0") == Literal::make_typed<datatypes::xsd::Date>("1742-1-1"));
+    CHECK(Literal::make_typed<datatypes::xsd::Date>("1742-2-40") == Literal::make_typed<datatypes::xsd::Date>("1742-2-28"));
+    CHECK(Literal::make_typed<datatypes::xsd::Time>("10:90:00") == Literal::make_typed<datatypes::xsd::Time>("11:30:00"));
```
liss-h, line 33: makes sense

tests/nodes/tests_DBpediaMode.cpp
```diff
@@ -0,0 +1,60 @@
...
+    CHECK(Literal::make_typed<datatypes::xsd::Time>("26:90:60") == Literal::make_typed<datatypes::xsd::Time>("23:59:59.999"));
```
liss-h: I think we need to discard that one

src/rdf4cpp/rdf/datatypes/registry/DatatypeRegistry.cpp
```diff
@@ -2,6 +2,10 @@
 
 namespace rdf4cpp::rdf::datatypes::registry {
 
+#ifndef RDF4CPP_DBPEDIA_MODE
+bool dbpedia_mode = false;
```
liss-h: rename to something like
- relaxed_parsing_mode
- lax_parsing_mode

or similar

liss-h (changes requested):
Thx for the PR
I have discussed this with @bigerl and came to some conclusions

1. We will probably never need the option to disable lax parsing at compiletime. It will always be in runtime mode anyway. But: have you tested if it makes any perf difference if the variable is runtime?
2. Some value corrections performed here are problematic, since they change the original data. I've added separate comments to the problematic cases. (Not all of them but you probably get the idea)



TODOS:
- [x] add documentation for relaxed mode (i.e. what does it to correct things). Reasoning being that people will not read the code to figure that out. Also: add a "since rdf4cpp-0.0.x" to the top, so that its easy to tell when behaviour changes and since when it behaves this way

docs/source/users_guide.rst
```diff
@@ -43,3 +43,11 @@ Parsing Files
 The class :class:`rdf4cpp::rdf::parser::RDFFileParser` allows reading files containing rdf Statements and iterate over them.
 Supported Formats: Turtle, TriG, NTriple and NQuad.
 :class:`rdf4cpp::rdf::parser::IStreamQuadIterator` allows doing the same over arbitrary data streams.
+
+Relaxed Parsing Mode
+--------------------
+
+The setting :var:`rdf4cpp::rdf::datatypes::registry::relaxed_parsing_mode` disables IRI validity checks and allows rdf4cpp to automatically try to correct some faulty Literals.
+Currently aimed at reducing loading errors with DBPedia, more might be added in future versions.
```
liss-h, line 52: Please specify exactly what it does (i.e. list all transformations)

liss-h (changes requested):
One documentation thing left

docs/source/users_guide.rst
mcb5637: i purposefully choose not to add it here, but instead link to the generated page for the variable. it is specified there in the comment.

liss-h: Oh ok, didn't recognize that syntax

liss-h (approved):

liss-h:
Waiting with merge until we have our parsing perf problems figured out.

PR #231, opened 2023-12-20
Fix: Parser malformed input rejection

liss-h:
Fixes various issues that caused malformed inputs to sometimes not be rejected during parsing.
Additionally this PR simplifies the interface and implementation of the parser.

Important takeaways:
- serd does not like it if you do not return `SERD_SUCCESS` in event handlers. I couldn't make sense of what it does but it somehow messes up the parsing state sometimes
- the iterator sometimes ate errors without surfacing them

### DIFF
--- a/private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -33,11 +33,10 @@ nonstd::expected<Node, SerdStatus> IStreamQuadIterator::Impl::get_bnode(std::str
         } catch (std::runtime_error const &e) {
             // TODO: check when actual blank node validation implemented
             // NOTE: line, col not entirely accurate as this function is called after a triple was parsed
-            this->last_error = ParsingError{
-                    .error_type = ParsingError::Type::BadBlankNode,
-                    .line = serd_reader_get_current_line(this->reader),
-                    .col = serd_reader_get_current_col(this->reader),
-                    .message = std::string{e.what()} + ". note: position may not be accurate and instead point to the end of the triple."};
+            this->last_error = ParsingError{.error_type = ParsingError::Type::BadBlankNode,
+                                            .line = serd_reader_get_current_line(this->reader),
+                                            .col = serd_reader_get_current_col(this->reader),
+                                            .message = std::string{e.what()} + ". note: position may not be accurate and instead point to the end of the triple."};
 
             return nonstd::make_unexpected(SERD_ERR_BAD_SYNTAX);
         }
@@ -53,11 +52,10 @@ nonstd::expected<Node, SerdStatus> IStreamQuadIterator::Impl::get_bnode(std::str
     } catch (std::runtime_error const &e) {
         // TODO: check when actual blank node validation implemented
         // NOTE: line, col not entirely accurate as this function is called after a triple was parsed
-        this->last_error = ParsingError{
-                .error_type = ParsingError::Type::BadBlankNode,
-                .line = serd_reader_get_current_line(this->reader),
-                .col = serd_reader_get_current_col(this->reader),
-                .message = std::string{e.what()} + ". note: position may not be accurate and instead point to the end of the triple."};
+        this->last_error = ParsingError{.error_type = ParsingError::Type::BadBlankNode,
+                                        .line = serd_reader_get_current_line(this->reader),
+                                        .col = serd_reader_get_current_col(this->reader),
+                                        .message = std::string{e.what()} + ". note: position may not be accurate and instead point to the end of the triple."};
 
         return nonstd::make_unexpected(SERD_ERR_BAD_SYNTAX);
     }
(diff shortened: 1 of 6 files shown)


private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -185,19 +186,21 @@ SerdStatus IStreamQuadIterator::Impl::on_base(void *voided_self, const SerdNode
...
-                    .col = serd_reader_get_current_col(self->reader.get()),
-                    .message = std::format("Error setting base: {}. hint: prefix parsing is currently deactivated. note: position may not be accurate and instead point to the end of the line.", e)};
-            return SERD_ERR_BAD_SYNTAX;
-        }
+
+        return SERD_SUCCESS;
+    }
+
+    auto e = self->state->iri_factory.set_base(node_into_string_view(uri));
+    if (e != IRIFactoryError::Ok) {
```
bigerl:
```suggestion
    } else if (auto e = self->state->iri_factory.set_base(node_into_string_view(uri));
                e != IRIFactoryError::Ok) {
```

would also do it. Then you can get rid of all but the final SERD_SUCCESS. Returning in 3 branches the same SERD_SUCCESS confuses me :)

private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -209,13 +212,13 @@ SerdStatus IStreamQuadIterator::Impl::on_prefix(void *voided_self, SerdNode cons
...
         self->last_error = ParsingError{
                 .error_type = ParsingError::Type::BadSyntax,
-                .line = serd_reader_get_current_line(self->reader.get()),
-                .col = serd_reader_get_current_col(self->reader.get()),
+                .line = serd_reader_get_current_line(self->reader),
+                .col = serd_reader_get_current_col(self->reader),
                 .message = "Encountered prefix while parsing. hint: prefix parsing is currently deactivated. note: position may not be accurate and instead point to the end of the line."};
-    } else {
-        self->state->iri_factory.assign_prefix(node_into_string_view(name), node_into_string_view(uri));
+        return SERD_SUCCESS;
```
bigerl: personally, I preferred here the if-else. If there is an if with a return at the end I expect a short cut. this is none.

private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -321,41 +324,53 @@ IStreamQuadIterator::Impl::Impl(void *stream,
...
+                serd_reader_skip_until_byte(this->reader, '\n');
+            }
+            return nonstd::make_unexpected(*std::exchange(this->last_error, std::nullopt));
+        } else if (this->end_flag) {
+            return std::nullopt;
+        }
+
+        SerdStatus const st = serd_reader_read_chunk(this->reader);
+
+        if (st == SERD_SUCCESS) {
```
bigerl, line 340: Where do those st != SERD_SUCCESS come from? From within serd? They don't come from rdf4cpp anymore.

private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -321,41 +324,53 @@ IStreamQuadIterator::Impl::Impl(void *stream,
...
-                serd_reader_skip_until_byte(this->reader.get(), '\n');
-                return nonstd::make_unexpected(*this->last_error);
-            } else if (this->last_error.has_value()) {
-                // non-fatal, artificially inserted error
-                return nonstd::make_unexpected(*this->last_error);
+        if (this->last_error.has_value()) {
+            // handle error from last time
+            if (this->last_error_requires_skip) {
+                this->last_error_requires_skip = false;
+                serd_reader_skip_until_byte(this->reader, '\n');
```
bigerl, line 331: Might jump to far for turtle. Looking for the next `.` would be better.

bigerl (changes requested):
Thx for the PR. Looks good. Most things are more style and I have some remarks.

private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
liss-h: yes, all those now actual parse errors from serd. Not validation errors from us anymore

private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
liss-h: Turtle is also usually written with newlines on each virtual triple, right? problem with looking for dots is that you could sometimes land inside an IRI

bigerl: yes usually. but it is not required AFAIK. Ending up inside a query sounds like a thing that could happen quite frequently. OK, then let's keep it as it is for now.

bigerl (approved):

PR #232, opened 2024-01-02
Feature: `ParsingFlags::NoParseBlankNode`, `current_line()` and `current_column()` for `IStreamQuadIterator`

liss-h:


tests/parser/tests_IStreamQuadIterator.cpp
```diff
@@ -308,6 +308,20 @@ TEST_SUITE("IStreamQuadIterator") {
...
 
+    TEST_CASE("no bnode parsing") {
+        constexpr char const *triples = "<a> <b> _:bnode .\n";
+
+        std::istringstream iss{triples};
+        IStreamQuadIterator qit{iss, ParsingFlag::NoParseBlankNode};
+
+        CHECK_NE(qit, std::default_sentinel);
+        CHECK(!qit->has_value());
+        std::cout << qit->error() << std::endl;
```
bigerl, line 320: check that the message contains something from `"Encountered blank node while parsing"`.

bigerl (changes requested):
Thx for the PR. Changes look straight forward. I only have one suggestions to improve the test case for the `NoParseBlankNode` mode.

src/rdf4cpp/rdf/parser/ParsingError.hpp
```diff
@@ -25,41 +25,60 @@ struct ParsingError {
...
-                os << "bad blank node";
-                break;
-        }
+} // namespace rdf4cpp::rdf::parser
 
-        return os;
+namespace std {
+
+template<>
+struct formatter<::rdf4cpp::rdf::parser::ParsingError::Type> : formatter<string_view> {
```
bigerl, line 35: Why is this one inheriting from `std::formatter<std::string_view>`?
Same for ParsingError's formatter

bigerl (changes requested):
test looks good now. One more question below

src/rdf4cpp/rdf/parser/ParsingError.hpp
liss-h: I saw that in some example at some point but had no idea so just looked it up now:
`formatter` also needs the `parse` function, since there are no arguments we might want to parse we just steal the one for formatting `string_view`

bigerl (approved):

PR #233, opened 2024-01-04
Fix: heap buffer overflow on `write_str` with `StringWriter` and large input

liss-h:
Fixes 1 issue and adds 1 improvement:
1. The cursor was not correctly repointed in `StringWriter::flush_impl` which caused a heap buffer overflow on `write_str` calls with large strings (larger than new buffer space)
2. reduces the amount of reallocations needed for `StringWriter` by adding an additional size parameter to `flush`

### DIFF
--- a/src/rdf4cpp/rdf/writer/BufWriter.hpp
+++ b/src/rdf4cpp/rdf/writer/BufWriter.hpp
@@ -55,7 +55,7 @@ public:
  * If the cursor says that there is no more room flush should be called.
  */
 template<typename W>
-concept BufWriter = requires (W &bw) {
+concept BufWriter = requires (W &bw, size_t additional_cap) {
     /**
      * Return type of finalize, typically void
      */
@@ -84,7 +84,7 @@ concept BufWriter = requires (W &bw) {
     /**
      * Flushes bw.buffer() and repoints bw.cursor() to the new free space
      */
-    W::flush(&bw.buffer(), bw.cursor());
+    W::flush(&bw.buffer(), bw.cursor(), additional_cap);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/rdf4cpp/rdf/writer/BufWriter.hpp
```diff
@@ -115,12 +116,12 @@ inline bool write_str(std::string_view str, void *const buffer, Cursor &cursor,
...
-        } else {
-            if (flush(buffer, cursor); cursor.size() == 0) [[unlikely]] {
-                return false;
-            }
+        }
+
+        str.remove_prefix(max_write);
 
-            str.remove_prefix(max_write);
+        if (flush(buffer, cursor, str.size()); cursor.size() == 0) [[unlikely]] {
```
bigerl, line 123: remove_prefix moved before flush where `str.size()` is used to assess the remaining capacity requirement. Looks good to me.

src/rdf4cpp/rdf/writer/BufWriter.hpp
```diff
@@ -181,12 +181,12 @@ struct StringWriter : BufWriterBase<StringWriter, StringBuffer> {
...
-        auto const bytes_written = cursor.data() - buffer.data();
+    static void flush_impl(Buffer &buffer, Cursor &cursor, size_t additional_cap) noexcept {
+        auto const bytes_filled = cursor.data() - buffer.data();
 
-        buffer.resize(buffer.size() * 2);
-        cursor.repoint(buffer.data() + bytes_written,
-                       cursor.size() + buffer.size());
+        buffer.resize(std::bit_ceil(buffer.size() + additional_cap));
+        cursor.repoint(buffer.data() + bytes_filled,
+                       buffer.size() - bytes_filled);
```
bigerl, line 189: change looks sensible. Growth is now controlled by the data not by exponential growth.

bigerl (approved):
looks good. thx

src/rdf4cpp/rdf/writer/BufWriter.hpp
liss-h: still exponential (just sped up) because of `bit_ceil`

PR #236, opened 2024-01-09
fix for IRI validation unserinfo and host

mcb5637:


liss-h (approved):
Well, if the spec says so

PR #237, opened 2024-01-09
Fix: IRI validation performance

liss-h:
Changes:
1. use unordered_dense instead of dice sparse_map for storing prefixes
2. fixes some suboptimal (or nonexistent) buffer prereserves in `IRIFactory`
3. introduces a base parts cache to `IRIFactory` to avoid reparsing the base for every IRI in `IRIFactory::from_relative`
4. Conditionally call `IRIFactory::from_relative` in the parser only if the format allows relative IRIs. Otherwise fall back to the faster `IRIFactory::create_and_validate`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,6 @@ find_package(OpenSSL REQUIRED)
 find_package(uni-algo QUIET)
 find_package(dice-hash QUIET)
 find_package(dice-sparse-map QUIET)
-find_package(unordered_dense QUIET)
 
 include(FetchContent)
 
@@ -80,16 +79,6 @@ if (NOT dice-sparse-map_FOUND)
             GIT_SHALLOW TRUE)
 
     FetchContent_MakeAvailable(fetchdice-sparse-map)
-endif ()
-
-if (NOT unordered_dense_FOUND)
-    FetchContent_Declare(
-            fetchunordered_dense
-            GIT_REPOSITORY https://github.com/martinus/unordered_dense.git
-            GIT_TAG v4.4.0
-            GIT_SHALLOW TRUE)
-
-    FetchContent_MakeAvailable(fetchunordered_dense)
 endif ()
 
 
@@ -245,7 +234,6 @@ target_link_libraries(rdf4cpp
         Boost::headers
         dice-hash::dice-hash
         dice-sparse-map::dice-sparse-map
-        unordered_dense::unordered_dense
         PRIVATE
         re2::re2
         OpenSSL::Crypto
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,7 +89,7 @@ if (NOT unordered_dense_FOUND)
             GIT_TAG v4.4.0
             GIT_SHALLOW TRUE)
 
-    FetchContent_MakeAvailable(fetchdice-sparse-map)
+    FetchContent_MakeAvailable(fetchunordered_dense)
 endif ()
 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ find_package(OpenSSL REQUIRED)
 find_package(uni-algo QUIET)
 find_package(dice-hash QUIET)
 find_package(dice-sparse-map QUIET)
+find_package(unordered_dense REQUIRED)
 
 include(FetchContent)
 
@@ -234,6 +235,7 @@ target_link_libraries(rdf4cpp
         Boost::headers
         dice-hash::dice-hash
         dice-sparse-map::dice-sparse-map
+        unordered_dense::unordered_dense
         PRIVATE
         re2::re2
         OpenSSL::Crypto
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ find_package(OpenSSL REQUIRED)
 find_package(uni-algo QUIET)
 find_package(dice-hash QUIET)
 find_package(dice-sparse-map QUIET)
-find_package(unordered_dense REQUIRED)
+find_package(unordered_dense QUIET)
 
 include(FetchContent)
 
@@ -77,6 +77,16 @@ if (NOT dice-sparse-map_FOUND)
             fetchdice-sparse-map
             GIT_REPOSITORY https://github.com/dice-group/dice-sparse-map.git
             GIT_TAG v0.2.4
+            GIT_SHALLOW TRUE)
+
+    FetchContent_MakeAvailable(fetchdice-sparse-map)
+endif ()
+
+if (NOT unordered_dense_FOUND)
+    FetchContent_Declare(
+            fetchunordered_dense
+            GIT_REPOSITORY https://github.com/martinus/unordered_dense.git
+            GIT_TAG v4.4.0
             GIT_SHALLOW TRUE)
 
     FetchContent_MakeAvailable(fetchdice-sparse-map)
--- a/private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -6,7 +6,7 @@
 namespace rdf4cpp::rdf::parser {
 
 std::string_view IStreamQuadIterator::Impl::node_into_string_view(SerdNode const *node) noexcept {
-    return std::string_view{reinterpret_cast<char const *>(node->buf), static_cast<size_t>(node->n_bytes)};
+    return std::string_view{reinterpret_cast<char const *>(node->buf), node->n_bytes};
 }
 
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ find_package(OpenSSL REQUIRED)
 find_package(uni-algo QUIET)
 find_package(dice-hash QUIET)
 find_package(dice-sparse-map QUIET)
+find_package(unordered_dense QUIET)
```
liss-h: Motivation for including unordered_dense: The number of prefixes is typically low, but their usage might occur often => faster map better

But I actually have no idea if this makes any appreciable difference because we don't really have datasets in turtle format

src/rdf4cpp/rdf/IRIFactory.cpp
```diff
@@ -1,55 +1,70 @@
...
-        iri = construct(*b_scheme, r_auth, remove_dot_segments(r_path), r_query, r_frag);
-    } else if (r_path.empty()) {
-        iri = construct(*b_scheme, b_auth, b_path, r_query.has_value() ? r_query : b_query, r_frag);
-    } else if (r_path.starts_with('/')) {
-        iri = construct(*b_scheme, b_auth, remove_dot_segments(r_path), r_query, r_frag);
-    } else {
-        auto m = merge_paths(IRIView{base}, r_path);
-        iri = construct(*b_scheme, b_auth, remove_dot_segments(m), r_query, r_frag);
-    }
+    std::string iri = [this, rel]() noexcept {
```
bigerl: I think this lambda should live elsewhere. As a function then. IRIView, maybe?

src/rdf4cpp/rdf/IRIFactory.cpp
```diff
@@ -1,55 +1,70 @@
...
 nonstd::expected<IRI, IRIFactoryError> IRIFactory::from_prefix(std::string_view prefix, std::string_view local, storage::node::NodeStorage &storage) const {
     auto i = prefixes.find(prefix);
-    if (i == prefixes.end())
+    if (i == prefixes.end()) {
         return nonstd::make_unexpected(IRIFactoryError::UnknownPrefix);
-    std::string deref {i->second};
+    }
+
+    std::string deref;
+    deref.reserve(i->second.size() + local.size());
```
bigerl: good idea. Could even be static thread local. That would amortize the allocations. 
would start with clear then. 

note from string::clear:

> Unlike for [std::vector::clear](https://en.cppreference.com/w/cpp/container/vector/clear), the C++ standard does not explicitly require that [capacity](https://en.cppreference.com/w/cpp/string/basic_string/capacity) is unchanged by this function, but existing implementations do not change capacity. This means that they do not release the allocated memory (see also [shrink_to_fit](https://en.cppreference.com/w/cpp/string/basic_string/shrink_to_fit)).

I think we can just ignore that for now. The safe solution, though, would be to use a `vector<char>`.

src/rdf4cpp/rdf/IRIFactory.cpp
```diff
@@ -62,7 +77,7 @@ void IRIFactory::clear_prefix(std::string_view prefix) {
 }
 
 std::string IRIFactory::remove_dot_segments(std::string_view path) noexcept {
-    std::string r{};
+    std::string r;
```
bigerl: same here. static thread local to avoid allocations.

src/rdf4cpp/rdf/IRIFactory.cpp
```diff
@@ -121,51 +136,65 @@ void IRIFactory::remove_last_path_segment(std::string &path) noexcept {
...
+    r.push_back('/');
     r.append(path);
     return r;
 }
 
 std::string IRIFactory::construct(std::string_view scheme, std::optional<std::string_view> auth, std::string_view path,
                                   std::optional<std::string_view> query, std::optional<std::string_view> frag) noexcept {
-    size_t l = scheme.length() + 1 + path.length();
+    size_t l = scheme.size() + 1 + path.size();
     if (auth.has_value()) {
```
bigerl: I am not sure if saving this little memory is worth the additional branches. Just add +4

CMakeLists.txt
bigerl: I would only switch to unordered_dense when we see that it is a bottle neck. For the typically low amount of prefixes (<20 from my experience) I think boost's flat_map should be just as fast. 

For now, I prefer to keep our dependencies short.

private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -6,7 +6,7 @@
 namespace rdf4cpp::rdf::parser {
 
 std::string_view IStreamQuadIterator::Impl::node_into_string_view(SerdNode const *node) noexcept {
-    return std::string_view{reinterpret_cast<char const *>(node->buf), static_cast<size_t>(node->n_bytes)};
+    return std::string_view{reinterpret_cast<char const *>(node->buf), node->n_bytes};
```
bigerl, line 9:
```suggestion
    return std::string_view{static_cast<char const *>(node->buf), node->n_bytes};
```

`node->buf` is `void const *`, right?

private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -71,8 +71,16 @@ nonstd::expected<Node, SerdStatus> IStreamQuadIterator::Impl::get_bnode(std::str
 }
 
 nonstd::expected<IRI, SerdStatus> IStreamQuadIterator::Impl::get_iri(SerdNode const *node) noexcept {
+    auto const iri = [this, node]() noexcept {
+        auto s = node_into_string_view(node);
```
bigerl:
```suggestion
        auto const s = node_into_string_view(node);
```

bigerl (changes requested):
Thx for the PR. Looks good and hopefully does a bit. 

I suggested at two spots to use static thread local string buffers. We might also maintaining buffers elsewhere and pass them to the functions (or have them accessible as a field). That way we could get the code more-or less amortized allocation free.

src/rdf4cpp/rdf/IRIFactory.cpp
liss-h: using bit_ceil instead

private/rdf4cpp/rdf/parser/IStreamQuadIteratorSerdImpl.cpp
liss-h: No its actually `uint8_t`

bigerl: oh. ok, then never mind.

src/rdf4cpp/rdf/IRIFactory.hpp
```diff
@@ -87,6 +90,8 @@ class IRIFactory {
     [[nodiscard]] static nonstd::expected<IRI, IRIFactoryError> create_and_validate(std::string_view iri, storage::node::NodeStorage &storage = storage::node::NodeStorage::default_instance()) noexcept;
 
 private:
+    [[nodiscard]] std::string to_absolute(std::string_view rel) const noexcept;
```
bigerl: add docstring

src/rdf4cpp/rdf/IRIFactory.cpp
```diff
@@ -158,19 +168,8 @@ std::string IRIFactory::merge_paths(IRIView::AllParts const &base, std::string_v
...
-    }
-    if (query.has_value()) {
-        l += query->size() + 1;
-    }
-    if (frag.has_value()) {
-        l += frag->size() + 1;
-    }
-
     std::string str;
-    str.reserve(l);
```
bigerl, line 150: add docstring explaining the formula.

And why does bit_ceil solve the problem?

src/rdf4cpp/rdf/IRIFactory.hpp
```diff
@@ -99,6 +90,8 @@ struct IRIFactory {
...
```
bigerl: please double check if some methods here can be made static. (have to go now otherwise I would have checked myself)

bigerl (changes requested):

src/rdf4cpp/rdf/IRIFactory.hpp
liss-h: I've converted all functions where it is possible to internal linkage

src/rdf4cpp/rdf/IRIFactory.cpp
liss-h: It avoids the branches and preserves power of two growth semantics. This is our best possible change to not realloc if we don't want branches and don't wanto to randomly guess the size of the additional parts

bigerl (approved):

bigerl (approved):
looks good. Algorithm seems equivalent. And given that it passes the test cases, this should be fine.

PR #238, opened 2024-01-17
Quad serialization

### LINKED ISSUE #234: Implement serializers for various rdf formats
The following formats are required:
- ntriples
- turtle
- nquads
- trig

The interface should look something like this:
```c++
bool Quad::serialize_ntriples(void *const buffer, Cursor *const cursor, FlushFunc const flush) {
    ... // actual impl here, inspiration can be gained from e.g. Literal::serialize
}

bool Quad::serialize_turtle(TurtleState *const state, void *const buffer, Cursor *const cursor, FlushFunc const flush) {
    ... // actual impl here, inspiration can be gained from e.g. Literal::serialize
}

template<BufWriter W>
bool Quad::serialize_ntriples(W &w) {
    return serialize_ntriples(&w.buffer(), &w.cursor(), &W::flush);
}

template<BufWriter W>
bool Quad::serialize_turtle(TurtleState &state, W &w) {
    return serialize_turtle(&state, &w.buffer(), &w.cursor(), &W::flush);
}
```

Example interface usage (n-triples):
```c++
writer::StringWriter ser{256};

for (Quad const &q : quads) {
    q.serialize_ntriples(ser);
}

std::cout << ser.finalize() << std::endl;
```

Example interface usage (turtle):
```c++
writer::StringWriter ser{256};
writer::TurtleState st{};

st.begin(ser); // writes prefixes of known datatypes (in our case this would be `xsd`, `rdf` and `owl`)

for (Quad const &q : quads) {
    q.serialize_turtle(st, ser);
}

std::cout << ser.finalize() << std::endl;
```

For turtle and trig it can be assumed that all triples are batched by subject. e.g.:
```
subj1 .. ..
subj1 .. ..
subj2 .. ..
subj2 .. ..
```

(new)
The `serialize` functions of `Dataset` and `Graph` should be updated to use the new interfaces. The way `Graph` and `Dataset` are implemented currently however, they will only be able to properly support `ntriples` serialization (because the order is not guaranteed). Sidenote I think the current implementation for `Dataset` is also wrong (because the graph is at the start, I think it should be at the end of the quad).

mcb5637:
close #234


src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -664,7 +665,7 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
     }                                                  \
     RDF4CPP_DETAIL_TRY_WRITE_STR("\"");
 
-bool Literal::serialize(void *const buffer, writer::Cursor *cursor, writer::FlushFunc const flush) const noexcept {
+bool Literal::serialize(void *const buffer, writer::Cursor *cursor, writer::FlushFunc const flush, const writer::SerializationState* state) const noexcept {
```
liss-h: Please create two versions of this function. (They can share the same impl switched by a bool)

- keep `Literal::serialize` interface as it was
- add `Literal::serialize_short` (better name would be good) that serializes for formats like turtle, i.e. shortens the datatype IRI 

Note that datatype shortening can be done without any additional information from the outside for fixed id datatypes (which is what we want).

Have a look at `datatypes::registry::reserved_datatype_ids` you can (at compiletime) create a `array<string>` (index by LiteralType) out of it that stores the shortened urls for those datatypes.

You can get the literal type to index the array out of the backend handle

src/rdf4cpp/rdf/IRI.cpp
```diff
@@ -100,13 +101,10 @@ IRI::operator datatypes::registry::DatatypeIDView() const noexcept {
     }
 }
 
-bool IRI::serialize(void *const buffer, writer::Cursor *cursor, writer::FlushFunc const flush) const noexcept {
+bool IRI::serialize(void *const buffer, writer::Cursor *cursor, writer::FlushFunc const flush, const writer::SerializationState* state) const noexcept {
```
liss-h: Just to reiterate what we discussed in the meeting and have it here: Currently we don't want to shorten arbitrary IRIs. So you can restore the interface to its previous version

liss-h (changes requested):
Partial review of some basics. Mainly reiterating what we discussed and giving some more context so you don't have to search around for solutions

src/rdf4cpp/rdf/Literal.cpp
liss-h: *from the outside meaning without an additional function param

liss-h:
Another idea: maybe it makes sense to add the IRI that corresponds to the `a` predicate to the IRIs that are always present in node-storages. Simply because that would allow you to just compare the nodeid of the IRIs you receive to the one of the `a` predicate.

So this would be a integer compare instead of a string compare for each IRI that gets serialized, which is probably a lot faster

mcb5637:
@Clueliss  do you need the exact interface from the Issue for FFI or are the templates here enough?

src/rdf4cpp/rdf/datatypes/registry/FixedIdMappings.hpp
```diff
@@ -13,6 +13,7 @@ namespace rdf4cpp::rdf::datatypes::registry {
 using storage::node::identifier::LiteralType;
 
 inline constexpr util::ConstexprString default_graph_iri{""};
+inline constexpr util::ConstexprString rdf_type{"http://www.w3.org/1999/02/22-rdf-syntax-ns#type"};
```
liss-h, line 16: I see why you have added this here (probably because `default_graph_iri` is also here, no idea why I added it here), but thinking about it, it really does not make sense to have them in the `reserved_datatype_ids` map. Since they are both not datatypes and therefore just occupy `LiteralType` values without ever being used as such.

I need to think a little about how to best approach this, because backwards abi compatibility and such

src/rdf4cpp/rdf/writer/Serialization.cpp
```diff
@@ -0,0 +1,117 @@
+#include "Serialization.hpp"
+
+namespace rdf4cpp::rdf::writer {
+
+class TypeIRIBufferBuilder {
```
liss-h: On first glance this looks a little more complicated than it needs to be for the functionality. I think building the array can just be a single for loop inside a `consteval` function. I might be wrong, have not implemented it to try it out

src/rdf4cpp/rdf/writer/Serialization.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+template<OutputFormat F>
+bool serialize(const Quad &s, void *const buffer, Cursor &cursor, FlushFunc const flush, SerializationState *state) {
+    auto write_node = &Node::serialize_short_form;
+    if constexpr (!format_has_prefix<F>)
+        write_node = &Node::serialize;
+    auto write_predicate = [&](const Node &p) {
+        if constexpr (format_has_prefix<F>) {
+            static constexpr size_t rdf_type = datatypes::registry::reserved_datatype_ids[datatypes::registry::rdf_type].to_underlying();
+            if (p.is_iri() && p.backend_handle().node_id().value() == rdf_type)
```
liss-h: That is not the correct way to compare this.
The correct way is `iri_node_id_to_literal_type(p.backend_handle().node_id()) == rdf_type`.

That this works the way it does is an implementation detail.

src/rdf4cpp/rdf/writer/Serialization.hpp
```diff
@@ -0,0 +1,178 @@
...
+    }
+    return true;
+}
+
+template<OutputFormat F>
+bool serialize(const Quad &s, void *const buffer, Cursor &cursor, FlushFunc const flush, SerializationState *state) {
+    auto write_node = &Node::serialize_short_form;
+    if constexpr (!format_has_prefix<F>)
+        write_node = &Node::serialize;
+    auto write_predicate = [&](const Node &p) {
```
liss-h: Predicates are always IRIs.

src/rdf4cpp/rdf/writer/Serialization.hpp
```diff
@@ -0,0 +1,178 @@
...
+        state->active_predicate = Node::make_null();
+        state->active_subject = Node::make_null();
+        state->active_graph = Node::make_null();
+    }
+    return true;
+}
+
+template<OutputFormat F>
+bool serialize(const Quad &s, void *const buffer, Cursor &cursor, FlushFunc const flush, SerializationState *state) {
+    auto write_node = &Node::serialize_short_form;
```
liss-h: Use static dispatch instead of dynamic. I.e.

```c++
struct ShortFormNodeSerializer {
    static bool serialize(Node term, void *const buffer, Cursor *cursor, FlushFunc const flush) noexcept {
        return term.serialize_short_form(buffer, cursor, flush);
    }
};

struct LongFormNodeSerializer {
    static bool serialize(Node term, void *const buffer, Cursor *cursor, FlushFunc const flush) noexcept {
        return term.serialize(buffer, cursor, flush);
    }
};

[..]

using NodeSer = std::conditional_t<format_has_prefix<F>, ShortFormNodeSerializer, LongFormNodeSerializer>;

NodeSer::serialize(...) // this was write_node(...) before.

```

src/rdf4cpp/rdf/writer/Serialization.hpp
```diff
@@ -0,0 +1,178 @@
...
+        if constexpr (format_has_prefix<F>) {
+            static constexpr size_t rdf_type = datatypes::registry::reserved_datatype_ids[datatypes::registry::rdf_type].to_underlying();
+            if (p.is_iri() && p.backend_handle().node_id().value() == rdf_type)
+                return write_str("a", buffer, &cursor, flush);
+        }
+        return std::invoke(write_node, p, buffer, &cursor, flush);
+    };
+    if constexpr (format_has_prefix<F>) {
+        if (state == nullptr)
+            return false;
```
liss-h: That is not the purpose of the return value. The purpose is strictly: "did the writer succeed in writing/flushing"

src/rdf4cpp/rdf/writer/Serialization.hpp
```diff
@@ -0,0 +1,178 @@
...
+    if constexpr (!format_has_prefix<F>)
+        if (!write_str(" .\n", buffer, &cursor, flush))
+            return false;
+    return true;
+}
+
+bool write_type_iri(datatypes::registry::LiteralType t, void *buffer, writer::Cursor *cursor, writer::FlushFunc flush, bool short_form);
+
+template<OutputFormat F>
+class QuadSerializer {
```
liss-h: This class can be removed

src/rdf4cpp/rdf/writer/SerializationState.hpp
```diff
@@ -0,0 +1,29 @@
...
+    Turtle,
+    NQuads,
+    TriG
+};
+
+class SerializationState {
+    Node active_graph, active_subject, active_predicate;
+
+    template<OutputFormat F>
+    friend bool flush_state(void *buffer, Cursor &cursor, FlushFunc flush, SerializationState *state);
```
liss-h: just make the members public (i.e. make this a struct)

src/rdf4cpp/rdf/writer/SerializationState.hpp
```diff
@@ -0,0 +1,29 @@
...
+namespace rdf4cpp::rdf::writer {
+enum class OutputFormat {
+    NTriples,
+    Turtle,
+    NQuads,
+    TriG
+};
+
+class SerializationState {
+    Node active_graph, active_subject, active_predicate;
```
liss-h:
```
Node active_graph;
Node active_subject;
IRI active_predicate
```

src/rdf4cpp/rdf/writer/WriterBasics.hpp
```diff
@@ -0,0 +1,8 @@
+#ifndef RDF4CPP_WRITERBASICS_HPP
```
liss-h: This file is unused I think

src/rdf4cpp/rdf/Node.hpp
```diff
@@ -91,6 +91,24 @@ class Node {
...
+     *
+     * @param buffer pointer to arbitrary buffer structure
+     * @param cursor cursor into buffer
+     * @param flush function to flush the contents of buffer and update cursor to point to the new free space
+     * @return true if serialization was successful, false if a call to flush was not able to make room
+     *
+     * For specific usage examples have a look at StringWriter and
+     * how the template Node::serialize<BufWriter> uses Serializers and tests/bench_SerDe.cpp#serialize.
+     */
+    bool serialize_short_form(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const noexcept;
```
liss-h, line 111: The template version is missing `template<writer::BufWriter W>`

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -777,10 +777,21 @@ class Literal : public Node {
...
      * See Node::serialize
      */
-    bool serialize(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const noexcept;
+    inline bool serialize(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const noexcept {
+        return serialize(buffer, cursor, flush, false);
+    }
+    /**
+     * See Node::serialize_short_form
+     */
+    inline bool serialize_short_form(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const noexcept {
```
liss-h: Forgot to add the template version.
Sidenote: Functions with their body inside a struct are always implicitly `inline`

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -664,7 +666,7 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
     }                                                  \
     RDF4CPP_DETAIL_TRY_WRITE_STR("\"");
 
-bool Literal::serialize(void *const buffer, writer::Cursor *cursor, writer::FlushFunc const flush) const noexcept {
+bool Literal::serialize(void *const buffer, writer::Cursor *cursor, writer::FlushFunc const flush, bool short_form) const noexcept {
```
liss-h: This should be a template `template<bool short_form>`

liss-h (changes requested):
Ok, apart from the specific things I requested in the comments:

The interface provided here is not what was requested in the issue, and some interfaces are also just missing (granted they were not explicitly requested, but they just make the interface adhere to whats already present). 

I have updated the issue to have some more examples, to make things more clear. Also fixed some typos. 

Again, you may share the implementation of the `Quad::serialize_*` functions, but that they are shared is a detail that should be hidden away inside a `.cpp` file.

I think I get your intention with the interface change, you wanted to make this more high level/easier to use. But thats not the point of this API, it is deliberately low-level.

The high-level interfaces (that users will probably interact with) are in `Dataset::serialize(W &w)` and `Graph::serialize(W &w)`, those are basically fool proof and conveniently serialize a whole graph/dataset (that they currently cannot support turtle serialization is a problem for later).

Now that I think about it, maybe it makes sense to add convenience interfaces to those which just parse a whole file and put the contents in them.. But thats also nothing for this PR. (edit: created an issue #242)

src/rdf4cpp/rdf/writer/Serialization.cpp
mcb5637: I tried that at first. But for various constexpr reasons, this seems to be the best c++20 implementation.

src/rdf4cpp/rdf/Literal.cpp
mcb5637: If i make this a template, it needs to go into the header. With all the makro stuff going on, that might be a bad idea.

src/rdf4cpp/rdf/writer/Serialization.cpp
liss-h: Implemented it. Not quite a for loop but simple for-loop-like recursion does the trick

src/rdf4cpp/rdf/Literal.cpp
liss-h: Doesn't need to go in the header, all instantiations will be in the cpp file if you put the other serialize functions in the cpp file as well instead of defining them in the header

src/rdf4cpp/rdf/writer/Serialization.hpp
liss-h: I noticed that there are problems further down in the code. I.e. in `Quad` it is also not `IRI` for some reason. Conflicted on whether it makes sense to fix this here. Probably leave it as is, but `assert` that the predicate is IRI instead of checking it with `if`

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -635,6 +637,66 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
...
+        }
+
+        return make_type_iri_buf<N, Ix + 1, Len>(ret);
+    }
+}
+
+static constexpr auto type_iri_buffer = make_type_iri_buf<1 << storage::node::identifier::LiteralType::width, 0, datatypes::registry::reserved_datatype_ids.size()>();
+
+// check at compile time, if all datatype ids have a valid mapping. if not, at throws, which breaks the static_assert
+static consteval bool validate_iri_buffer() {
```
liss-h: This doesn't really do much in terms of testing. It would be better to write a real test (in the tests directory). Like, serialize a literal as short form and check if the short form is correct. You don't need to do that for everything just some selected ones.

src/rdf4cpp/rdf/Graph.cpp
```diff
@@ -78,16 +79,36 @@ bool Graph::serialize(void *const buffer, writer::Cursor *cursor, writer::FlushF
...
-        RDF4CPP_DETAIL_TRY_SERIALIZE(solution[0]);
-        RDF4CPP_DETAIL_TRY_WRITE_STR(" ");
+        Quad q{solution[0], solution[1], solution[2]};
+        if (!q.serialize_ntriples(buffer, cursor, flush))
+            return false;
+    }
+
+    return true;
+}
+bool Graph::serialize_turtle(void *const buffer, writer::Cursor *cursor, writer::FlushFunc const flush) const noexcept {
```
liss-h, line 89: Question: Does this work or will it just basically always be normal ntriples?

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -635,6 +637,66 @@ Literal Literal::as_language_tag_eq(Literal const &other, Node::NodeStorage &nod
...
+        if (type_iri_buffer.at(d.second.to_underlying()).size() == 1)  // also check, if something is just :
+            return false;
+    }
+    return true;
+}
+static_assert(validate_iri_buffer());
+
+// will only get called with fixed ids
+template<bool short_form>
+bool write_type_iri(datatypes::registry::LiteralType t, void *const buffer, writer::Cursor *cursor, writer::FlushFunc const flush) {
```
liss-h, line 668: For safety add an assert that `t` is not the dynamic type

src/rdf4cpp/rdf/Node.cpp
```diff
@@ -154,6 +174,9 @@ std::weak_ordering Node::operator<=>(const Node &other) const noexcept {
 }
 
 bool Node::operator==(const Node &other) const noexcept {
+    if (!this->is_literal() && !other.is_literal() && this->backend_handle().node_storage_id() == other.backend_handle().node_storage_id()) {
```
liss-h, line 177: Does this not work for literals? Could you explain why, I think I am missing somthing

src/rdf4cpp/rdf/writer/SerializationState.cpp
```diff
@@ -0,0 +1,40 @@
+#include "SerializationState.hpp"
+
+namespace rdf4cpp::rdf::writer {
+
+class PrefixDataBuilder {
```
liss-h: I think this adds some unecessary complexity. I would prefer if you would just write out all prefixes manually using a chain `[..]TRY_WRITE()` in a for loop (inside `SerializationState::begin`.

This will only happen exactly once per however many billion triples so the minor perf benefit will never show.

liss-h (changes requested):
Thanks for the changes 👍 I've added some more comments, will go over it again a final time when they are addressed

src/rdf4cpp/rdf/Node.cpp
mcb5637: Because Literals of different datatypes can be equals, which gives us different backend handles even if they are in the same node storage.

liss-h: Right, forgot that. Could you add a comment explaining that?

src/rdf4cpp/rdf/Quad.hpp
```diff
@@ -65,6 +68,28 @@ class Quad : public query::QuadPattern {
...
+    bool serialize_nquad(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const;
+    bool serialize_turtle(writer::SerializationState& state, void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const;
+    bool serialize_trig(writer::SerializationState& state, void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const;
+
+    template<writer::BufWriter W>
+    bool serialize_ntriples(W &w) const noexcept {
+        return serialize_ntriples(&w.buffer(), &w.cursor(), &W::flush);
+    }
+    template<writer::BufWriter W>
+    bool serialize_nquad(W &w) const noexcept {
```
liss-h: I think there is a typo here. Probably should be `serialize_nquads`

src/rdf4cpp/rdf/Graph.cpp
mcb5637: It works, you get prefixed type IRIs in any case, and if the order in which Triples are serialized is good, you get lists as well.
(There is even a test case for it)

liss-h: Ok yeah, thats what I meant. I would assume that the order is always bad. But I guess you still get some benefits even if its bad

src/rdf4cpp/rdf/datatypes/registry/FixedIdMappings.hpp
liss-h: Regarding this issue: I talked with @bigerl and we agreed that it probably does not make sense to fix the underlying problem in this PR because it would be a slightly larger fix and the problem has not much to do with this PR.

Could you please add a TODO (and create an issue) that reminds us that we need to refactor this at some point?

liss-h (approved):
PR looks good 👍
I've also already tried integrating it into our upstream consuming repos and encountered no issues.

Can be merged once the one TODO is in place

src/rdf4cpp/rdf/datatypes/registry/FixedIdMappings.hpp
liss-h: Quickly created them myself in the interest of getting this merged

PR #240, opened 2024-01-26
Feature: RAND with user provided random number generator

liss-h:
Adds the following additional overload to `Literal::generate_random_double`:
```c++
template<typename Rng>
[[nodiscard]] static Literal Literal::generate_random_double(Rng &rng, NodeStorage &node_storage = NodeStorage::default_instance());
```

### DIFF
--- a/src/rdf4cpp/rdf/Literal.hpp
+++ b/src/rdf4cpp/rdf/Literal.hpp
@@ -5,6 +5,7 @@
 #include <any>
 #include <optional>
 #include <ostream>
+#include <random>
 #include <rdf4cpp/rdf/Node.hpp>
 #include <rdf4cpp/rdf/datatypes/LiteralDatatype.hpp>
 #include <rdf4cpp/rdf/datatypes/owl.hpp>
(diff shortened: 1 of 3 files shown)


src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -329,10 +330,28 @@ class Literal : public Node {
...
      * @see https://www.w3.org/TR/sparql11-query/#idp2130040
      */
     [[nodiscard]] static Literal generate_random_double(NodeStorage &node_storage = NodeStorage::default_instance());
 
+    /**
+     * Generates a random double in the range [0.0, 1.0).
+     * The values are generated using the given random number generator
+     *
+     * @param rng random number generator
+     * @return random double in [0.0, 1.0]
```
nkaralis: this should be [0.0, 1.0)

src/rdf4cpp/rdf/Literal.hpp
```diff
@@ -329,10 +330,28 @@ class Literal : public Node {
...
+     *
+     * @see https://www.w3.org/TR/sparql11-query/#idp2130040
+     */
+    template<typename Rng>
+    [[nodiscard]] static Literal generate_random_double(Rng &rng, NodeStorage &node_storage = NodeStorage::default_instance()) {
+        // uniform_real_distribution does not have any state, therefore we can construct a new one for each call
+        std::uniform_real_distribution<typename datatypes::xsd::Double::cpp_type> dist{0.0, 1.0};
+        return Literal::make_typed_from_value<datatypes::xsd::Double>(dist(rng), node_storage);
```
nkaralis, line 352: to generate multiple and different values, does the random engine need to be manipulated by the caller of `generate_random_double`?

nkaralis (commented):
A typo needs fixing and I left a question.

src/rdf4cpp/rdf/Literal.hpp
liss-h: No, you can just call it multiple times
Example:
```c++
std::default_random_engine rng{/*seed*/};
auto r1 = Literal::generate_random_double(rng);
auto r2 = Literal::generate_random_double(rng);

// r1 != r2 (unlikely to be equal)
```

nkaralis: Sorry, I got confused with the test case. Did not notice that you initialize the engine in each for loop.

nkaralis:
Can be merged

src/rdf4cpp/rdf/Literal.hpp
liss-h: Its meant as a sanity check that the `uniform_real_distribution` really does not have state and doesn't influence the generated values

liss-h: because the function call operator is not const for some reason

PR #245, opened 2024-02-26
Feature: node storage optimization

liss-h:
This PR attempts to optimize node-storage size and access pattern.

Size is reduced by:
  - using a custom string type that does not store the capacity (minimal improvement, but storing the capacity is not necessary)
  - using a `std::vector` instead of `sparse_map` for the forward lookup direction (not sure how much this does)

Access pattern is improved by:
  - increased data locality achieved by flattening the forward access structure (lookup by id now takes 3 indirections instead of 5)

### DIFF
--- a/.github/workflows/run_test_and_examples.yml
+++ b/.github/workflows/run_test_and_examples.yml
@@ -9,8 +9,6 @@ jobs:
             compiler: gcc-13
 
           - os: ubuntu-22.04
-            compiler: clang-15
-          - os: ubuntu-22.04
             compiler: clang-16
           - os: ubuntu-22.04
             compiler: clang-17
--- a/src/rdf4cpp/rdf/storage/node/identifier/NodeID.hpp
+++ b/src/rdf4cpp/rdf/storage/node/identifier/NodeID.hpp
@@ -79,8 +79,6 @@ public:
 
     explicit operator uint64_t() const noexcept { return value_; }
 
-    //    void value(uint64_t val) { assert(val < (1UL << 48)); value_ = val; }
-
     constexpr std::strong_ordering operator<=>(NodeID const &other) const noexcept { return value_ <=> other.value_; }
 
     constexpr bool operator==(NodeID const &other) const noexcept { return value_ == other.value_; }
--- a/src/rdf4cpp/rdf/storage/node/identifier/NodeID.hpp
+++ b/src/rdf4cpp/rdf/storage/node/identifier/NodeID.hpp
@@ -76,6 +76,9 @@ public:
      */
     [[nodiscard]] constexpr LiteralType literal_type() const noexcept { return LiteralType::from_underlying(literal_.literal_type); }
     [[nodiscard]] constexpr uint64_t value() const noexcept { return value_; }
+
+    explicit operator uint64_t() const noexcept { return value_; }
+
     //    void value(uint64_t val) { assert(val < (1UL << 48)); value_ = val; }
 
     constexpr std::strong_ordering operator<=>(NodeID const &other) const noexcept { return value_ <=> other.value_; }
--- a/src/rdf4cpp/rdf/storage/node/reference_node_storage/SyncReferenceNodeStorageBackend.cpp
+++ b/src/rdf4cpp/rdf/storage/node/reference_node_storage/SyncReferenceNodeStorageBackend.cpp
@@ -4,13 +4,13 @@
 namespace rdf4cpp::rdf::storage::node::reference_node_storage {
 
 SyncReferenceNodeStorageBackend::SyncReferenceNodeStorageBackend() noexcept {
-    iri_storage_.mapping.reserve(NodeID::min_iri_id);
-    bnode_storage_.mapping.reserve(NodeID::min_bnode_id);
-    variable_storage_.mapping.reserve(NodeID::min_variable_id);
-    fallback_literal_storage_.mapping.reserve(NodeID::min_literal_id);
+    iri_storage_.mapping.reserve(identifier::NodeID::min_iri_id);
+    bnode_storage_.mapping.reserve(identifier::NodeID::min_bnode_id);
+    variable_storage_.mapping.reserve(identifier::NodeID::min_variable_id);
+    fallback_literal_storage_.mapping.reserve(identifier::NodeID::min_literal_id);
 
     specialization_detail::tuple_for_each(specialized_literal_storage_, [](auto &storage) {
-        storage.mapping.reserve(NodeID::min_literal_id);
+        storage.mapping.reserve(identifier::NodeID::min_literal_id);
     });
 
     // some iri's like xsd:string are there by default
@@ -66,12 +66,12 @@ bool SyncReferenceNodeStorageBackend::has_specialized_storage_for(identifier::Li
  * @return the NodeID for the looked up Node Backend. Result is the null-id if there was no matching Node Backend.
  */
 template<bool create_if_not_present, typename Storage>
-static identifier::NodeID lookup_or_insert_impl(typename Storage::BackendView const &view,
+static identifier::NodeID lookup_or_insert_impl(typename Storage::backend_view_type const &view,
                                                 Storage &storage) noexcept {
 
     {
         std::shared_lock lock{storage.mutex};
-        if (auto const id = storage.mapping.lookup_id(view); id != typename Storage::BackendId{}) {
+        if (auto const id = storage.mapping.lookup_id(view); id != typename Storage::backend_id_type{}) {
             return Storage::to_node_id(id, view);
         }
     }
@@ -82,7 +82,7 @@ static identifier::NodeID lookup_or_insert_impl(typename Storage::BackendView co
         std::unique_lock lock{storage.mutex};
 
         // check again, might have changed between unlocking of shared_lock and locking of unique_lock
-        if (auto const id = storage.mapping.lookup_id(view); id != typename Storage::BackendId{}) {
+        if (auto const id = storage.mapping.lookup_id(view); id != typename Storage::backend_id_type{}) {
             return Storage::to_node_id(id, view);
         }
 
@@ -149,7 +149,7 @@ identifier::NodeID SyncReferenceNodeStorageBackend::find_id(view::VariableBacken
 }
 
 template<typename Storage>
-static typename Storage::BackendView find_backend_view(Storage &storage, identifier::NodeID const id) {
+static typename Storage::backend_view_type find_backend_view(Storage &storage, identifier::NodeID const id) {
     std::shared_lock<std::shared_mutex> shared_lock{storage.mutex};
 
     if (auto view = storage.mapping.lookup_value(Storage::to_backend_id(id)); view.has_value()) {
--- a/src/rdf4cpp/rdf/storage/node/reference_node_storage/SyncReferenceNodeStorageBackend.cpp
+++ b/src/rdf4cpp/rdf/storage/node/reference_node_storage/SyncReferenceNodeStorageBackend.cpp
@@ -4,25 +4,27 @@
 namespace rdf4cpp::rdf::storage::node::reference_node_storage {
 
 SyncReferenceNodeStorageBackend::SyncReferenceNodeStorageBackend() noexcept {
-    // set correct initial value for atomics
-    for (auto &id : next_specialized_literal_ids_) {
-        id = NodeID::min_literal_id.value;
-    }
+    iri_storage_.mapping.reserve(NodeID::min_iri_id);
+    bnode_storage_.mapping.reserve(NodeID::min_bnode_id);
+    variable_storage_.mapping.reserve(NodeID::min_variable_id);
+    fallback_literal_storage_.mapping.reserve(NodeID::min_literal_id);
+
+    specialization_detail::tuple_fold(specialized_literal_storage_, 0, [](auto acc, auto &storage) noexcept {
+        storage.mapping.reserve(NodeID::min_literal_id);
+        return acc;
+    });
 
     // some iri's like xsd:string are there by default
     for (const auto &[iri, literal_type] : datatypes::registry::reserved_datatype_ids) {
         auto const id = literal_type.to_underlying();
-
-        auto const [it, inserted] = iri_storage_.id2data.emplace(id, std::make_unique<IRIBackend>(view::IRIBackendView{.identifier = iri}));
-        assert(inserted);
-        iri_storage_.data2id.emplace(it->second.get(), id);
+        iri_storage_.mapping.insert_assume_not_present_at(view::IRIBackendView{.identifier = iri}, identifier::NodeID{id});
     }
 }
 
 template<typename Backend>
 static size_t storage_size(SyncNodeTypeStorage<Backend> const &storage) noexcept {
     std::shared_lock<std::shared_mutex> l{storage.mutex};
-    return storage.id2data.size();
+    return storage.mapping.size();
 }
 
 size_t SyncReferenceNodeStorageBackend::size() const noexcept {
@@ -45,19 +47,16 @@ bool SyncReferenceNodeStorageBackend::has_specialized_storage_for(identifier::Li
  * @tparam create_if_not_present enables code for creating non-existing Node Backends
  * @param view contains the data of the requested Node Backend
  * @param storage the storage where the Node Backend is looked up
- * @param next_id_func function to generate the next ID which is assigned in case a new Node Backend is created
  * @return the NodeID for the looked up Node Backend. Result is the null-id if there was no matching Node Backend.
  */
-template<bool create_if_not_present, typename Storage, typename NextIDFunc = void *>
-    requires (!create_if_not_present || std::is_nothrow_invocable_r_v<identifier::NodeID, NextIDFunc>)
+template<bool create_if_not_present, typename Storage>
 static identifier::NodeID lookup_or_insert_impl(typename Storage::BackendView const &view,
-                                                Storage &storage,
-                                                NextIDFunc next_id_func = nullptr) noexcept {
+                                                Storage &storage) noexcept {
 
     {
         std::shared_lock lock{storage.mutex};
-        if (auto const it = storage.data2id.find(view); it != storage.data2id.end()) {
-            return it->second;
+        if (auto const id = storage.mapping.lookup_id(view); id != typename Storage::BackendId{}) {
+            return Storage::to_node_id(id, view);
         }
     }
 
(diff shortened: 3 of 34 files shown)


liss-h:
TODO:
  - [x] fix the serialization test case (it has the triples in the wrong order now because different order in storage)

.github/workflows/run_test_and_examples.yml
```diff
@@ -8,8 +8,6 @@ jobs:
           - os: ubuntu-22.04
             compiler: gcc-13
 
-          - os: ubuntu-22.04
-            compiler: clang-15
```
bigerl, line 12: why was clang-15 removed?

src/rdf4cpp/rdf/storage/node/identifier/NodeID.hpp
```diff
@@ -76,6 +76,9 @@ class __attribute__((__packed__)) NodeID {
      */
     [[nodiscard]] constexpr LiteralType literal_type() const noexcept { return LiteralType::from_underlying(literal_.literal_type); }
     [[nodiscard]] constexpr uint64_t value() const noexcept { return value_; }
+
+    explicit operator uint64_t() const noexcept { return value_; }
+
     //    void value(uint64_t val) { assert(val < (1UL << 48)); value_ = val; }
```
bigerl: remove the comment

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
```diff
@@ -0,0 +1,188 @@
...
+        size_t hash; // hash of the Value being looked up
+        id_type id;  // id into forward_
+    };
+
+    struct backward_key_equal {
+        using is_transparent = void;
+
+        using forward_const_pointer = typename std::allocator_traits<allocator_type>::template rebind_traits<forward_type>::const_pointer;
+
+        forward_const_pointer forward_ptr; // pointer to forward_ to be able to check for equality between View and Value
```
bigerl:
```suggestion
        forward_const_pointer forward_ptr; ///< pointer to forward_ to be able to check for equality between View and Value
```

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
```diff
@@ -0,0 +1,188 @@
...
+    backward_type backward_; // backward_[view] stores Id into forward_
+    [[no_unique_address]] allocator_type alloc_;
+
+    [[nodiscard]] static constexpr size_type to_index(id_type id) noexcept {
+        return static_cast<size_type>(id) - 1;
+    }
+
+    [[nodiscard]] static constexpr id_type to_id(size_type ix) noexcept {
+        return static_cast<id_type>(ix + 1);
+    }
```
bigerl: Please add a docstring to these

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
```diff
@@ -0,0 +1,188 @@
...
+    BiDirFlatMap &operator=(BiDirFlatMap const &) = delete;
+    BiDirFlatMap &operator=(BiDirFlatMap &&) = delete;
+
+    [[nodiscard]] size_type size() const noexcept {
+        return forward_.size();
+    }
+
+    void shrink_to_fit() {
+        forward_.shrink_to_fit();
+        // TODO backward_?
```
bigerl: rehash(0)

bigerl:
```suggestion
        backward_.rehash(0);
```

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
```diff
@@ -0,0 +1,188 @@
...
+        auto it = backward_.find(view);
+        if (it == backward_.end()) {
+            return Id{};
+        }
+
+        return it->id;
+    }
+
+    void reserve(id_type const min_id) {
+        forward_.resize(to_index(min_id));
```
bigerl: this might delete entries. only `min_id >= forward.size()` should be valid.

bigerl: also do `bachward_.reserve(min_id);`

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
```diff
@@ -0,0 +1,188 @@
...
+    [[nodiscard]] static constexpr size_type to_index(id_type id) noexcept {
+        return static_cast<size_type>(id) - 1;
+    }
+
+    [[nodiscard]] static constexpr id_type to_id(size_type ix) noexcept {
+        return static_cast<id_type>(ix + 1);
+    }
+
+public:
+    explicit BiDirFlatMap(hasher const &hash = hasher{},
```
bigerl, line 111: Please add docstrings to the public methods and ctors

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/ConstString.hpp
```diff
@@ -0,0 +1,103 @@
...
+        // prevent copy constructors from being accidentally called
+        BasicConstString(BasicConstString const &) = delete;
+        BasicConstString &operator=(BasicConstString const &) = delete;
+
+        BasicConstString(BasicConstString &&other) noexcept : data_{std::exchange(other.data_, nullptr)},
+                                                              size_{std::exchange(other.size_, 0)},
+                                                              alloc_{std::move(other.alloc_)} {
+        }
+
+        BasicConstString &operator=(BasicConstString &&other) noexcept(std::allocator_traits<Allocator>::propagate_on_container_move_assignment::value
```
bigerl, line 49: nicely done

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/ConstString.hpp
```diff
@@ -0,0 +1,103 @@
+#ifndef RDF4CPP_REFERENCENODESTORAGE_CONSTSTRING_HPP
+#define RDF4CPP_REFERENCENODESTORAGE_CONSTSTRING_HPP
+
+#include <string>
+
+namespace rdf4cpp::rdf::storage::node::reference_node_storage::detail {
+
+    template<typename Char, typename Traits = std::char_traits<Char>, typename Allocator = std::allocator<Char>>
+    struct BasicConstString {
```
bigerl, line 13: please add at least a docstring to the class

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/SpecializationDetail.hpp
```diff
@@ -20,23 +20,42 @@ constexpr Acc tuple_fold_impl(std::index_sequence<Ixs...>, Tuple const &tuple, A
...
+    return tuple_fold_impl(std::make_index_sequence<std::tuple_size_v<std::remove_cvref_t<Tuple>>>{}, tuple, std::forward<Acc>(init), std::forward<FoldF>(f));
+}
+
+template<typename Tuple, typename F>
+constexpr void tuple_for_each(Tuple &&tuple, F &&f) {
+    return tuple_for_each_impl(std::make_index_sequence<std::tuple_size_v<std::remove_cvref_t<Tuple>>>{}, std::forward<Tuple>(tuple), std::forward<F>(f));
+}
+
+template<typename Tuple, typename F>
+constexpr void tuple_type_for_each(F &&f) {
```
bigerl, line 49: we could think about moving the tuple functions to dice template library

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/SpecializationDetail.hpp
```diff
@@ -20,23 +20,42 @@ constexpr Acc tuple_fold_impl(std::index_sequence<Ixs...>, Tuple const &tuple, A
...
+    return tuple_type_for_each_impl<Tuple>(std::make_index_sequence<std::tuple_size_v<Tuple>>{}, std::forward<F>(f));
 }
 
 template<typename Tuple>
 static consteval std::array<bool, 1 << identifier::LiteralType::width> make_storage_specialization_lut() noexcept {
```
bigerl, line 63: Please add a docstring.

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/UnsyncNodeTypeStorage.hpp
```diff
@@ -22,15 +23,12 @@ struct UnsyncNodeTypeStorage {
     struct DefaultBackendTypeEqual {
         using is_transparent = void;
 
-        bool operator()(Backend const *lhs, Backend const *rhs) const noexcept {
```
bigerl, line 25: did you remove this because it is no longer required?

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/UnsyncNodeTypeStorage.hpp
```diff
@@ -68,15 +63,25 @@ struct UnsyncNodeTypeStorage {
...
-    };
+    }
+
+    static BackendId to_backend_id(identifier::NodeID const id) noexcept {
+        if constexpr (requires { Backend::to_backend_id(id); }) {
+            return Backend::to_backend_id(id);
+        } else {
+            return id;
+        }
+    }
```
bigerl: add docstrings

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    void reserve(id_type const min_id) {
+        forward_.resize(to_index(min_id));
+    }
+
+    [[nodiscard]] id_type insert_assume_not_present(view_type const &view) {
+        assert(lookup_id(view) == Id{});
+
+        auto const assigned_id = to_id(forward_.size());
```
bigerl: This is a general design issue: if we create n + 1 entries, remove the first n entries and add another `n` entries we will end up with 2n + 1 slots used. 

This is not easy to solve. The clean solution would be managing holes (as in files). The easy solution are tombstones. I would suggest a vector of `size_t`s where the n-th bit tells if the n-th slot is free.

src/rdf4cpp/rdf/storage/node/reference_node_storage/SyncReferenceNodeStorageBackend.cpp
```diff
@@ -4,25 +4,26 @@
 namespace rdf4cpp::rdf::storage::node::reference_node_storage {
 
 SyncReferenceNodeStorageBackend::SyncReferenceNodeStorageBackend() noexcept {
```
bigerl, line 6: side note: how about a common parent class for Sync/Unsync Ref Node Storage?

tests/serializer/tests_Serialization.cpp
```diff
@@ -18,35 +18,43 @@ TEST_CASE("Literal short type") {
     CHECK(ser.finalize()=="\"2042-05-04\"^^xsd:date\ntrue\n\"4.0E0\"^^xsd:float\n\"4\"^^xsd:unsignedByte\n4");
 }
 
-template<writer::OutputFormat F>
-bool serialize(const Quad& q, void *buffer, writer::Cursor &cursor, writer::FlushFunc flush, writer::SerializationState* state) {
```
bigerl: these changes seem totally unrelated

bigerl (changes requested):
thx for the PR. I like the proposed method because it makes things much cleaner. 

There is one major issue though with the design. See my comment in BiDirFlatMap.

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
liss-h: We had almost the same issue before (i.e. ids not being reused) just that it didn't occupy additional space. Didn't really think about that.

So I see two options:
1. keep the increasing id behaviour as is, apply an offset to the vector index so that we do not need to store the empty range at the start
2. Solve both problems by reusing ids

src/rdf4cpp/rdf/storage/node/reference_node_storage/SyncReferenceNodeStorageBackend.cpp
liss-h: I tried that in the original PR that introduced `UnsyncReferenceNodeStorageBackend`, but I stopped at some point because I decided that the additional effort was not worth it

tests/serializer/tests_Serialization.cpp
liss-h: 😅 I guess that happens if you work on two PRs at once. Those changes were meant for the other one

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/SpecializationDetail.hpp
liss-h: Yeah, probably makes sense

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
liss-h: Is the zero just a placeholder value that tells it to clean up?

.github/workflows/run_test_and_examples.yml
liss-h: Supporting it was additional effort because it apparently doesn't generate automatic constructors properly. See https://github.com/rdf4cpp/rdf4cpp/actions/runs/8063550056/job/22025596346

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
bigerl: yes, see https://en.cppreference.com/w/cpp/container/unordered_map/rehash

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
bigerl: I would reuse IDs. Let's have a short call about how to solve this tomorrow.

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/SpecializationDetail.hpp
liss-h: I will create an issue in template-lib and do it at some point later

tests/serializer/tests_Serialization.cpp
liss-h: I will just keep it in here, my intention is to merge the other PR first anyway

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
liss-h: added check

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/SpecializationDetail.hpp
liss-h: https://github.com/dice-group/dice-template-library/issues/33

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
liss-h: Implemented

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/BiDirFlatMap.hpp
liss-h: Left out ctors because there isn't really anything to document

liss-h:
Coverage is broken somehow, no idea what thats about

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/IndexFreeList.hpp
```diff
@@ -0,0 +1,143 @@
...
+        auto const last_occupied = std::find_if(occupied_bitmap_.rbegin(), occupied_bitmap_.rend(), [](auto const bitmap) noexcept {
+            return bitmap != 0;
+        });
+
+        auto const unused_size = std::distance(occupied_bitmap_.rbegin(), last_occupied);
+        occupied_bitmap_.resize(occupied_bitmap_.size() - unused_size);
+    }
+
+    /**
+     * Marks all slots up-to (not-including) the given index as occupied
```
bigerl, line 104: the implementation does not cover the case ix = 70 (>63) but there is a true at 130 (>127).

what is missing is a fill for slots before slot_ix if the condition in line 106 is not met. 

I think this is not practically relevant for our use-case but also very unexpected behavior. So, I would feel safer if it would be fixed.

bigerl (changes requested):
Changes look good. There is just one little corner case I found.

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/IndexFreeList.hpp
liss-h: Fixed this and another bug related to `occupy_until` (that also doesn't occur with our usecase, but can't hurt to fix it)

src/rdf4cpp/rdf/storage/node/reference_node_storage/detail/IndexFreeList.hpp
```diff
@@ -102,14 +104,19 @@ struct IndexFreeList {
...
+            occupied_bitmap_.resize(slot_ix + 1, all_ones);
+            occupied_bitmap_[slot_ix] = (1ul << offset) - 1;
+        } else {
+            std::fill_n(occupied_bitmap_.begin(), slot_ix, all_ones);
+            occupied_bitmap_[slot_ix] |= (1ul << offset) - 1;
         }
 
-        occupied_bitmap_[slot_ix] = (1ul << offset) - 1;
-        next_free_ix_ = ix;
+        next_free_ix_ = search_next_free(slot_ix, offset);
```
bigerl, line 119: good catch

bigerl (approved):

PR #246, opened 2024-02-27
Feature: faster serialization for non nformat

liss-h:
Adds additional interfaces for the purpose of increasing serialization speed in non-nformat use-cases.

### DIFF
--- a/src/rdf4cpp/rdf/Node.hpp
+++ b/src/rdf4cpp/rdf/Node.hpp
@@ -76,22 +76,22 @@ public:
     /**
      * Serialize the string representation of the given node in N-format as defined by <a href="https://www.w3.org/TR/n-triples/">N-Triples</a> and <a href="https://www.w3.org/TR/n-quads/">N-Quads</a>.
      *
-     * @param parts writer parts
+     * @param writer Typically, you pass in an instance of a writer::BufWriter (Concept) instance. Any writer::BufWriter instance is implicitly convertible to writer::BufWriterParts.
      * @return true if serialization was successful, false if a call to flush was not able to make room
      *
      * For specific usage examples have a look at tests/bench_SerDe.cpp#serialize.
      */
-    bool serialize(writer::BufWriterParts parts) const noexcept;
+    bool serialize(writer::BufWriterParts writer) const noexcept;
 
     /**
      * Serialize the string representation of the given node in N-format as defined by <a href="https://www.w3.org/TR/rdf12-turtle/">Turtle</a> and <a href="https://www.w3.org/TR/rdf12-trig/">TriG</a>.
      *
-     * @param parts writer parts
+     * @param writer writer parts
      * @return true if serialization was successful, false if a call to flush was not able to make room
      *
      * For specific usage examples have a look at tests/bench_SerDe.cpp#serialize.
      */
-    bool serialize_short_form(writer::BufWriterParts parts) const noexcept;
+    bool serialize_short_form(writer::BufWriterParts writer) const noexcept;
 
     /**
      * Returns a string representation of the given node in N-format as defined by <a href="https://www.w3.org/TR/n-triples/">N-Triples</a> and <a href="https://www.w3.org/TR/n-quads/">N-Quads</a>.
--- a/src/rdf4cpp/rdf/Node.hpp
+++ b/src/rdf4cpp/rdf/Node.hpp
@@ -76,60 +76,22 @@ public:
     /**
      * Serialize the string representation of the given node in N-format as defined by <a href="https://www.w3.org/TR/n-triples/">N-Triples</a> and <a href="https://www.w3.org/TR/n-quads/">N-Quads</a>.
      *
-     * Takes as input an arbitrary, flushable/extendable buffer type (e.g. a std::string), a cursor
-     * into the buffer (that describes its base address and length of the free segment) and a function to flush the buffer (i.e. to either flush to io or extend the buffer).
-     * During operation will update the cursor to reflect the current start of the spare capacity and remaining length.
-     * When the remaining length reaches zero, flush is called using the provided buffer and cursor. Its task is to somehow make more buffer
-     * room and then adjust the cursor to point to the new spare room.
-     *
-     * @param buffer pointer to arbitrary buffer structure
-     * @param cursor cursor into buffer
-     * @param flush function to flush the contents of buffer and update cursor to point to the new free space
+     * @param parts writer parts
      * @return true if serialization was successful, false if a call to flush was not able to make room
      *
-     * For specific usage examples have a look at StringWriter and
-     * how the template Node::serialize<BufWriter> uses Serializers and tests/bench_SerDe.cpp#serialize.
+     * For specific usage examples have a look at tests/bench_SerDe.cpp#serialize.
      */
-    bool serialize(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const noexcept;
-    /**
-     * Serialize the string representation of the given node in N-format as defined by <a href="https://www.w3.org/TR/rdf12-turtle/">Turtle</a> and <a href="https://www.w3.org/TR/rdf12-trig/">TriG</a>.
-     *
-     * Takes as input an arbitrary, flushable/extendable buffer type (e.g. a std::string), a cursor
-     * into the buffer (that describes its base address and length of the free segment) and a function to flush the buffer (i.e. to either flush to io or extend the buffer).
-     * During operation will update the cursor to reflect the current start of the spare capacity and remaining length.
-     * When the remaining length reaches zero, flush is called using the provided buffer and cursor. Its task is to somehow make more buffer
-     * room and then adjust the cursor to point to the new spare room.
-     *
-     * @param buffer pointer to arbitrary buffer structure
-     * @param cursor cursor into buffer
-     * @param flush function to flush the contents of buffer and update cursor to point to the new free space
-     * @return true if serialization was successful, false if a call to flush was not able to make room
-     *
-     * For specific usage examples have a look at StringWriter and
-     * how the template Node::serialize<BufWriter> uses Serializers and tests/bench_SerDe.cpp#serialize.
-     */
-    bool serialize_short_form(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const noexcept;
+    bool serialize(writer::BufWriterParts parts) const noexcept;
 
     /**
-     * Serialize the string representation of the given node in N-format as defined by <a href="https://www.w3.org/TR/n-triples/">N-Triples</a> and <a href="https://www.w3.org/TR/n-quads/">N-Quads</a>.
-     *
-     * @param w a serializer
-     * @return true if serialization was successful, false if a call to W::flush was not able to make room
-     */
-    template<writer::BufWriter W>
-    bool serialize(W &w) const noexcept {
-        return serialize(&w.buffer(), &w.cursor(), &W::flush);
-    }
-    /**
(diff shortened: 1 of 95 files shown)


src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -45,7 +45,7 @@ I from_chars(std::string_view s) {
  * @param value the value to be serialized
```
bigerl, line 46: update doctstring

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -120,16 +121,16 @@ constexpr size_t log10ceil(T const value) noexcept {
...
```
bigerl, line 110: update doctstring

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -192,7 +194,7 @@ std::string to_chars_canonical(F const value) noexcept {
...
```
bigerl, line 185: update doctstring

src/rdf4cpp/rdf/writer/BufWriter.hpp
```diff
@@ -157,36 +152,65 @@ struct BufWriterBase {
     [[nodiscard]] constexpr Cursor &cursor() noexcept { return cursor_; }
     [[nodiscard]] constexpr Buffer &buffer() noexcept { return buffer_; }
 
+    [[nodiscard]] constexpr Cursor const &cursor() const noexcept { return cursor_; }
+    [[nodiscard]] constexpr Buffer const &buffer() const noexcept { return buffer_; }
+
```
bigerl, line 186: Is there any Writer that doesn't use BufWriterBase? Otherwise, we could also make it a requirement to use it as Base and add write_str as static method of BufWriterBase.

src/rdf4cpp/rdf/writer/BufWriter.hpp
```diff
@@ -157,36 +152,65 @@ struct BufWriterBase {
...
     static void flush(void *buffer, Cursor *cursor, size_t additional_cap) noexcept {
         CRTP::flush_impl(*static_cast<typename CRTP::Buffer *>(buffer), *cursor, additional_cap);
     }
```
bigerl, line 189: I would suggest to add a declaration of finalize() here as well and document it. Important aspect to be mentioned: It must be called only once (at least for some implementations).

src/rdf4cpp/rdf/query/Variable.cpp
```diff
@@ -63,7 +63,7 @@ std::string_view Variable::name() const {
     return this->handle_.variable_backend().name;
 }
 
-bool Variable::serialize(void *const buffer, writer::Cursor *cursor, writer::FlushFunc const flush) const noexcept {
+bool Variable::serialize(void *const buffer, writer::Cursor *const cursor, writer::FlushFunc const flush) const noexcept {
```
bigerl: update interface in hpp as well

src/rdf4cpp/rdf/Quad.hpp
```diff
@@ -56,23 +56,14 @@ class Quad : public query::QuadPattern {
...
-                *(it++) = item.to_node_storage(node_storage);
-        return qu;
-    }
+    [[nodiscard]] Quad to_node_storage(storage::node::NodeStorage &node_storage) const noexcept;
 
-    bool serialize_ntriples(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const;
-    bool serialize_nquads(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const;
-    bool serialize_turtle(writer::SerializationState& state, void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const;
-    bool serialize_trig(writer::SerializationState& state, void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const;
+    bool serialize_ntriples(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const noexcept;
```
bigerl: *const missing in the signature compared to definition

src/rdf4cpp/rdf/writer/BufWriter.hpp
bigerl, line 1: A genral interface thing: it is very common that a `BufWriter` is used to extract ` void *const buffer, Cursor *cursor, FlushFunc const flush`. That is enough because typically, not the full functionality is required for `write_str`. Similar interface are everywhere throughout the code. 

For the user, providing buffer,cursor,flush is cumbersome. I would suggest that we introduce a type that has that essential info as fields and is implicitly convertible from any BufWriter implementation. That way user's can just pass in their BufWriter. 

This can completely replace the overloads like
```
template<BufWriter W>
bool write_str(std::string_view str, W &w) {
    return write_str(str, &w.buffer(), &w.cursor(), &W::flush);
}
```

and allows us to only provide one simple interface like:

```
bool serialize(BufferWriterProxy w) { // ugly name I know; maybe you have a better suggestion :)
    ...
}
```

in user-facing types. it is not templated so implementation details can completely go into cpp files.

bigerl (changes requested):
Thx for the PR. Things look good from what I can see. 
One thing I noticed, though. After going through the BufWriter architecture again for this review I found it felt a bit hard to follow. I have left some suggestions for improvements. Please start with the comments in BufWriter.

src/rdf4cpp/rdf/Quad.hpp
liss-h: the outer most `const` does nothing on declarations

src/rdf4cpp/rdf/writer/BufWriter.hpp
liss-h: Every bufwriter that is not implemented here (in rdf4cpp) does not use it

src/rdf4cpp/rdf/writer/BufWriter.hpp
liss-h: ~~Users are never really supposed to use the 3 param version, it simply exists for FFI purposes, adding the proxy defeats the purpose~~

liss-h: I guess we could make it constructable from the three params as well

liss-h: name suggestion: `BufWriterParts`

src/rdf4cpp/rdf/writer/BufWriter.hpp
liss-h: If we add a decl here we will get an IDE warning on every type inheriting from this (because the implementation in the derived will hide the previous one from the base)

src/rdf4cpp/rdf/Quad.hpp
bigerl: it is compatible, yes. It's more that it makes me uneasy that the signatures are different.

src/rdf4cpp/rdf/writer/BufWriter.hpp
bigerl: How about CRTP for finalize as well?

src/rdf4cpp/rdf/Quad.hpp
liss-h: Here are some arguments against using it in declarations:
1. It doesn't give the user any additional promises or any useful information at all
2. It doesn't help the compiler to optimize the code
3. It exposes an implementation detail that the users don't want to know and shouldn't know
4. It clutters the API surface with keywords that do nothing but make the function signature longer and harder to read
5. Declarations and definitions are already not symmetric in many other ways
   - `explicit` only appears in the declaration
   - `static` only appears in the declaration
   - `[[nodiscard]]` only appears in the declaration
   - `[[maybe_unused]]` only appears in defintions

bigerl: point 5. convinced me :)

src/rdf4cpp/rdf/writer/BufWriter.hpp
liss-h: I'm not sure what you are trying to achieve, but the `BufWriter` concept already tells you that you need to implement it and that you have to call it

bigerl: OK. That's true. 
I am also not so sure I must admit. My main concern was that users are not seeing the documentation of finalize() in their IDE. But maybe a link in the class docstring of the concrete buffer Writer implementations to the BufWriter concept would do.

src/rdf4cpp/rdf/Node.hpp
```diff
@@ -76,60 +76,22 @@ class Node {
...
-     * @param flush function to flush the contents of buffer and update cursor to point to the new free space
+     * @param parts writer parts
      * @return true if serialization was successful, false if a call to flush was not able to make room
      *
-     * For specific usage examples have a look at StringWriter and
-     * how the template Node::serialize<BufWriter> uses Serializers and tests/bench_SerDe.cpp#serialize.
+     * For specific usage examples have a look at tests/bench_SerDe.cpp#serialize.
      */
-    bool serialize(void *buffer, writer::Cursor *cursor, writer::FlushFunc flush) const noexcept;
+    bool serialize(writer::BufWriterParts parts) const noexcept;
```
bigerl: I would call the parameter something like writer, buf_writer, writer_p, buf_writer_parts, writer_parts. (oder: most prefered -> least preferred).
It should focus more on what a user typically passes in then what it is converted to. After all the BufWriter implementations are mostly for construction/defining behaviour and are internally not used directly.

whatever you apply, please apply it everywhere

src/rdf4cpp/rdf/Node.hpp
```diff
@@ -76,60 +76,22 @@ class Node {
...
-     * Takes as input an arbitrary, flushable/extendable buffer type (e.g. a std::string), a cursor
-     * into the buffer (that describes its base address and length of the free segment) and a function to flush the buffer (i.e. to either flush to io or extend the buffer).
-     * During operation will update the cursor to reflect the current start of the spare capacity and remaining length.
-     * When the remaining length reaches zero, flush is called using the provided buffer and cursor. Its task is to somehow make more buffer
-     * room and then adjust the cursor to point to the new spare room.
-     *
-     * @param buffer pointer to arbitrary buffer structure
-     * @param cursor cursor into buffer
-     * @param flush function to flush the contents of buffer and update cursor to point to the new free space
+     * @param parts writer parts
```
bigerl:
```suggestion
     * @param writer Typically, you pass in an instance of a writer::BufWriter (Concept) instance. Any writer::BufWriter instance is implicitly convertible to writer::BufWriterParts.
```

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -56,21 +59,7 @@ bool to_chars_canonical(I const value, void *buffer, writer::Cursor *cursor, wri
...
- * Serializes an integer into its _simplified_ string representation which is e.g. used in
- * casting integers to strings.
- * @note for integers this is identical to the canonical representation
- *
- * @see https://www.w3.org/TR/xpath-functions/#casting-to-string
- *
- * @param value the value to be serialized
- */
-template<std::integral I>
-std::string to_chars_simplified(I const value) noexcept {
```
bigerl, line 71: where did `to_chars_simplified` go?

bigerl (changes requested):

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
liss-h: Was unused

bigerl (approved):

Issue #239, opened 2024-01-25
Literal serialization performance improvement: Writer support for `Literal::lexical_form`


liss-h:
Currently, requesting the lexical form of a literal might cause a heap allocation, if the literal is in typed storage instead of lexical storage (because typed storage does not store the lexical form and a `std::string` with the lexical form needs to be materialized). 

This should be fixed by adding the following functions to the interface of Literal:
```c++
bool Literal::serialize_lexical_form(void *const buffer, Cursor &cursor, FlushFunc const flush) {
    ... 
}

template<BufWriter W>
bool Literal::serialize_lexical_form(W &w) {
    return serialize_lexical_form(&w.buffer(), &w.cursor(), &W::flush);
}
```

To actually get rid of the heap allocation however, this requires explicit support in the `DatatypeRegistry`, (i.e. adding an additional function pointer with signature `bool serialize(std::any, void *const buffer, Cursor &cursor, FlushFunc const flush)` to the registry entries. This function can then be efficiently implemented for each datatype and called from `Literal::serialize_lexical_form`.

The other typical node accessor function (like `Literal::language_tag()` or `IRI::identififer()` probably do not need this treatment because they do not need to be materialized, such functions can be added however to contribute to a uniform interface.

This is again an effort to improve Literal serialization performance.

Usecase: serializing to formats that are not N-format-like

liss-h:
Closed by #246

PR #253, opened 2024-03-12
Fix: some literal ops

liss-h:
Fixes some bugs introduced by #239

### DIFF
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -548,7 +548,18 @@ auto Literal::serialize_lexical_form_impl(C &&consume) const noexcept {
     }
 
     return handle_.literal_backend().visit(
-            [&consume](storage::node::view::LexicalFormLiteralBackendView const &lexical_backend) noexcept {
+            [this, &consume](storage::node::view::LexicalFormLiteralBackendView const &lexical_backend) noexcept {
+                if constexpr (simplified) {
+                    auto const *entry = datatypes::registry::DatatypeRegistry::get_entry(this->datatype_id());
+
+                    if (entry != nullptr) {
+                        auto const value = entry->factory_fptr(lexical_backend.lexical_form);
+                        return std::invoke(std::forward<C>(consume), value, entry->serialize_simplified_string_fptr);
+                    }
+                } else {
+                    (void) this;
+                }
+
                 return std::invoke(std::forward<C>(consume), lexical_backend.lexical_form);
             },
             [&consume](storage::node::view::ValueLiteralBackendView const &value_backend) noexcept {
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -2266,7 +2266,7 @@ Literal Literal::hash_with(char const *alg, NodeStorage &node_storage) const {
     std::span<std::byte const> const bytes{reinterpret_cast<std::byte const *>(hash_buffer), len};
 
     auto const lex = writer::StringWriter::oneshot([bytes](auto &w) {
-        return datatypes::xsd::HexBinary::cpp_type::serialize(bytes, w);
+        return datatypes::xsd::HexBinary::cpp_type::serialize_lowercase(bytes, w);
     });
 
     return Literal::make_simple(lex, node_storage);
--- a/src/rdf4cpp/rdf/Literal.cpp
+++ b/src/rdf4cpp/rdf/Literal.cpp
@@ -557,7 +557,7 @@ auto Literal::serialize_lexical_form_impl(C &&consume) const noexcept {
                         return std::invoke(std::forward<C>(consume), value, entry->serialize_simplified_string_fptr);
                     }
                 } else {
-                    (void) this;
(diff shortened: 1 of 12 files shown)


liss-h:
Coverage still broken, ignore

src/rdf4cpp/rdf/datatypes/xsd/HexBinary.cpp
```diff
@@ -55,24 +65,24 @@ HexBinaryRepr HexBinaryRepr::from_encoded(std::string_view const hex_encoded) {
...
-
-        auto const higher = encode_decode_detail::hex_decode(hex_encoded[ix - 1]);
-        auto const lower = encode_decode_detail::hex_decode(hex_encoded[ix]);
-
-        buf.push_back(static_cast<std::byte>((higher << 4) | lower));
-    }
-
     if (rest > 0) {
-        // handle last byte that was not represented by two digits (if there is one)
+        // handle first byte that was not represented by two digits (if there is one)
```
bigerl:
```suggestion
        // handle the case where the most significant byte is represented by only one digit
```

src/rdf4cpp/rdf/datatypes/xsd/HexBinary.hpp
bigerl, line 1: I think we should keep the internal representation equal between base64Binary and hexBinary. Both share the value space definition:
> The [·value space·](https://www.w3.org/TR/xmlschema-2/#dt-value-space) of base64Binary/hexBinary is the set of finite-length sequences of binary octets.

It does not specify if the octets store at position 0 LSB or MSB and it does not specify if the sequence goes from LS octet to MS octet or the other way around. 

I think though, that for the sake of keeping the interface easy to understand we should stick with little-endian for access and internal representation. So LS octet first and within the octet LSB is at position 0. Thin includes constructors and access functions.

I expect that we will use HexBinaries (or derived types) to store blobs like files in the future. There it would also be rather unusual to read hexBinary back-to-front.

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -548,7 +548,18 @@ auto Literal::serialize_lexical_form_impl(C &&consume) const noexcept {
     }
 
     return handle_.literal_backend().visit(
-            [&consume](storage::node::view::LexicalFormLiteralBackendView const &lexical_backend) noexcept {
+            [this, &consume](storage::node::view::LexicalFormLiteralBackendView const &lexical_backend) noexcept {
+                if constexpr (simplified) {
+                    auto const *entry = datatypes::registry::DatatypeRegistry::get_entry(this->datatype_id());
```
bigerl, line 553: good catch

src/rdf4cpp/rdf/Literal.cpp
```diff
@@ -548,7 +548,18 @@ auto Literal::serialize_lexical_form_impl(C &&consume) const noexcept {
...
+
+                    if (entry != nullptr) {
+                        auto const value = entry->factory_fptr(lexical_backend.lexical_form);
+                        return std::invoke(std::forward<C>(consume), value, entry->serialize_simplified_string_fptr);
+                    }
+                } else {
+                    (void) this;
+                }
```
bigerl: what is the else branch for? Compiler warning? If so, please add a comment.

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -189,7 +189,7 @@ bool to_chars_canonical(F const value, writer::BufWriterParts const writer) noex
 template<std::floating_point F>
 bool to_chars_simplified(F const value, writer::BufWriterParts const writer) noexcept {
     if (value == 0) {
-        return std::signbit(value) ? "-0" : "0";
+        return std::signbit(value) ? writer::write_str("-0", writer) : writer::write_str("0", writer);
```
bigerl, line 192: oops

bigerl (changes requested):
thx for the PR. Now, the tests should actually tell if things are wrong. 

I am against the change though that stores xsd:hexBinary as BigEndian. I explain my considerations in a comment below.

src/rdf4cpp/rdf/datatypes/xsd/HexBinary.hpp
liss-h: So what's the plan for the literal hash functions then? Should I just reimplements the same thing?

liss-h: Found a solution. It makes for a little awkward interface but works

src/rdf4cpp/rdf/datatypes/registry/util/CharConvExt.hpp
liss-h: yeah, implicit conversions are great..

src/rdf4cpp/rdf/datatypes/xsd/HexBinary.hpp
bigerl: I like the interface very much as it is now. Especially, the explicit `serialize_hash` for the hash algorithms.

bigerl (approved):
Looks good. thx :)

PR #255, opened 2024-03-18
external format and tidy file

mcb5637:
uses clang-format/tidy file from https://github.com/dice-group/tentris-cpp-coding-guidelines


liss-h (approved):
Looks good to me

PR #260, opened 2024-03-25
xsd:decimal inlining

### LINKED ISSUE #111: Future improvement: xsd:decimal inlining
Inlining `xsd:decimal` is currently not inlineable because of difficulty truncating the used type to 42bits without loss of precision.

Explored solutions:
- converting boost decimal float -> string -> native float => slow and does not work for many values
- gcc's `std::decimal` => incomplete (ex. no string conversion support)
- https://github.com/GaryHughes/stddecimal => promising but either:
    1.  would need to find a fast way to convert to/from boost decimal float
    2. would need to alter inlining api to be able to perform operations on a seperate "inlined-datatype" eg. `decimal32` for `xsd:decimal` if the precision allows it

Another possibility: explore boost decimal float allocator support => if it properly supports fancy pointers it would be possible to store the value directly in the backend (not the string repr) (#110)

mcb5637:
close #111 
(xsd:decimal has been changed since that issue was written, but still missed inlining)

### DIFF
--- a/src/rdf4cpp/datatypes/xsd/Decimal.hpp
+++ b/src/rdf4cpp/datatypes/xsd/Decimal.hpp
@@ -98,8 +98,8 @@ struct Decimal : registry::LiteralDatatypeImpl<registry::xsd_decimal,
                                                registry::capabilities::Numeric,
                                                registry::capabilities::Comparable,
                                                registry::capabilities::Promotable,
-                                                   registry::capabilities::FixedId,
-                                                   registry::capabilities::Inlineable> {};
+                                               registry::capabilities::FixedId,
+                                               registry::capabilities::Inlineable> {};
 
 }  // namespace rdf4cpp::datatypes::xsd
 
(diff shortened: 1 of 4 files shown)


liss-h (changes requested):
The implementation looks good, but please revert the formatting changes, we want to apply them all in one go (later)

src/rdf4cpp/datatypes/xsd/Decimal.hpp
```diff
@@ -91,7 +98,8 @@ struct Decimal : registry::LiteralDatatypeImpl<registry::xsd_decimal,
                                                registry::capabilities::Numeric,
                                                registry::capabilities::Comparable,
                                                registry::capabilities::Promotable,
-                                               registry::capabilities::FixedId> {};
+                                                   registry::capabilities::FixedId,
```
liss-h: indentation

liss-h (changes requested):
one formatting thing

liss-h (approved):

PR #264, opened 2024-03-27
Fix: misc fixes for node storage rework

liss-h:


src/rdf4cpp/Literal.cpp
```diff
@@ -409,12 +409,12 @@ Literal Literal::try_get_in_node_storage(storage::DynNodeStoragePtr node_storage
...
 
                     return node_storage.find_id(storage::view::ValueLiteralBackendView{.datatype = dt_id,
-                                                                                             .value = std::move(value)});
+                                                                                         .value = std::move(value)});
                 }
 
                 // Default case.
                 // This node storage doesn't have specialized storage for the value and the new one also doesn't
-                auto const dtype_iri_view = node_storage.find_iri_backend(lexical_backend.datatype_id);
+                auto const dtype_iri_view = handle_.storage().find_iri_backend(lexical_backend.datatype_id);
```
bigerl: Good catch. Took me a bit though, to understand why this is correct. 

It would make the code easier to read if there was a something like 
```
auto& source_node_storage = handle_.storage();
```

src/rdf4cpp/storage/identifier/NodeBackendID.hpp
```diff
@@ -6,7 +6,7 @@
 
 namespace rdf4cpp::storage::identifier {
 
-struct NodeBackendID {
+struct alignas(uint64_t) NodeBackendID {
```
bigerl, line 9: Why is the alignas necessary? I would have expected that it is already aligned like that.

bigerl (changes requested):
thx for the PR. Please find my comments below.

src/rdf4cpp/storage/identifier/NodeBackendID.hpp
liss-h: I guess its because of the bitfields, the alignment was 1 before

bigerl (approved):

PR #265, opened 2024-04-02
Node ctor validation

### LINKED ISSUE #261: Node constructor validation
`BlankNode`, `Variable` and `IRI` constructors should validate the passed in values.

Additionally, add `BlankNode::make_unchecked` and friends to be able to construct nodes without these checks (probably still add asserts in those tho).

mcb5637:
close #261

### DIFF
--- a/src/rdf4cpp/BlankNode.cpp
+++ b/src/rdf4cpp/BlankNode.cpp
@@ -144,8 +144,9 @@ std::string_view BlankNode::validate_bnode_name(std::string_view v) {
     }
     auto lastchar = *it;
     ++it;
+    static constexpr auto pn_matcher = PNCharsMatcher | ASCIIPatternMatcher(".-");
     while (it != r.end()) {
-        if (!PNCharsMatcher.match(*it))
+        if (!pn_matcher.match(*it))
         {
             throw std::invalid_argument(std::format("invalid blank node label {}", v));
         }
--- a/src/rdf4cpp/BlankNode.cpp
+++ b/src/rdf4cpp/BlankNode.cpp
@@ -3,6 +3,7 @@
 
 #include <rdf4cpp/util/CharMatcher.hpp>
 #include <rdf4cpp/writer/TryWrite.hpp>
+#include <rdf4cpp/ParsingError.hpp>
 #include <uni_algo/all.h>
 
 namespace rdf4cpp {
@@ -10,9 +11,7 @@ BlankNode::BlankNode() noexcept : Node{storage::identifier::NodeBackendHandle{{}
 }
 
 BlankNode::BlankNode(std::string_view identifier, storage::DynNodeStoragePtr node_storage)
-    : Node{storage::identifier::NodeBackendHandle{node_storage.find_or_make_id(storage::view::BNodeBackendView{.identifier = validate_bnode_name(identifier), .scope = std::nullopt}),
-                                                  storage::identifier::RDFNodeType::BNode,
-                                                  node_storage}} {
+    : BlankNode{make_unchecked(validate_bnode_name(identifier), node_storage)} {
 }
 
 BlankNode::BlankNode(storage::identifier::NodeBackendHandle handle) noexcept : Node{handle} {
@@ -141,10 +140,10 @@ std::string_view BlankNode::validate_bnode_name(std::string_view v) {
     auto r = v | una::views::utf8;
     auto it = r.begin();
     if (it == r.end()) {
-        throw std::invalid_argument("invalid blank node label (empty string)");
+        throw ParsingError("invalid blank node label (empty string)");
     }
     if (!first_matcher.match(*it)) {
-        throw std::invalid_argument(std::format("invalid blank node label {}", v));
+        throw ParsingError(std::format("invalid blank node label {}", v));
     }
     auto lastchar = *it;
     ++it;
@@ -152,13 +151,13 @@ std::string_view BlankNode::validate_bnode_name(std::string_view v) {
     while (it != r.end()) {
         if (!pn_matcher.match(*it))
         {
-            throw std::invalid_argument(std::format("invalid blank node label {}", v));
+            throw ParsingError(std::format("invalid blank node label {}", v));
         }
         lastchar = *it;
         ++it;
     }
     if (lastchar == '.') {
-        throw std::invalid_argument(std::format("invalid blank node label {}", v));
+        throw ParsingError(std::format("invalid blank node label {}", v));
     }
     return v;
 }
--- a/src/rdf4cpp/BlankNode.cpp
+++ b/src/rdf4cpp/BlankNode.cpp
@@ -1,14 +1,16 @@
 #include "BlankNode.hpp"
 #include <rdf4cpp/bnode_mngt/NodeScope.hpp>
 
+#include <rdf4cpp/util/CharMatcher.hpp>
 #include <rdf4cpp/writer/TryWrite.hpp>
+#include <uni_algo/all.h>
 
 namespace rdf4cpp {
 BlankNode::BlankNode() noexcept : Node{storage::identifier::NodeBackendHandle{{}, storage::identifier::RDFNodeType::BNode, {}}} {
 }
 
 BlankNode::BlankNode(std::string_view identifier, storage::DynNodeStoragePtr node_storage)
-    : Node{storage::identifier::NodeBackendHandle{node_storage.find_or_make_id(storage::view::BNodeBackendView{.identifier = identifier, .scope = std::nullopt}),
+    : Node{storage::identifier::NodeBackendHandle{node_storage.find_or_make_id(storage::view::BNodeBackendView{.identifier = validate_bnode_name(identifier), .scope = std::nullopt}),
                                                   storage::identifier::RDFNodeType::BNode,
                                                   node_storage}} {
 }
@@ -22,6 +24,12 @@ BlankNode BlankNode::make_null() noexcept {
 
 BlankNode BlankNode::make(std::string_view identifier, storage::DynNodeStoragePtr node_storage) {
     return BlankNode{identifier, node_storage};
+}
+
+BlankNode BlankNode::make_unchecked(std::string_view identifier, storage::DynNodeStoragePtr node_storage) {
+    return BlankNode{storage::identifier::NodeBackendHandle{node_storage.find_or_make_id(storage::view::BNodeBackendView{.identifier = identifier, .scope = std::nullopt}),
+                                                            storage::identifier::RDFNodeType::BNode,
+                                                            node_storage}};
 }
 
 BlankNode BlankNode::to_node_storage(storage::DynNodeStoragePtr node_storage) const {
@@ -121,6 +129,33 @@ TriBool BlankNode::union_eq(BlankNode const &other) const noexcept {
     assert(other_label.has_value());
 
     return *this_label == *other_label;
+}
+
+std::string_view BlankNode::validate_bnode_name(std::string_view v) {
+    using namespace util::char_matcher_detail;
+    static constexpr auto first_matcher = ASCIINumMatcher{} | PNCharsBaseMatcher;
+    auto r = v | una::views::utf8;
+    auto it = r.begin();
+    if (it == r.end()) {
+        throw std::invalid_argument("invalid blank node label (empty string)");
+    }
+    if (!first_matcher.match(*it)) {
+        throw std::invalid_argument(std::format("invalid blank node label {}", v));
+    }
+    auto lastchar = *it;
+    ++it;
+    while (it != r.end()) {
+        if (!PNCharsMatcher.match(*it))
+        {
+            throw std::invalid_argument(std::format("invalid blank node label {}", v));
+        }
+        lastchar = *it;
+        ++it;
+    }
+    if (lastchar == '.') {
+        throw std::invalid_argument(std::format("invalid blank node label {}", v));
+    }
+    return v;
 }
 
 inline namespace shorthands {
--- a/src/rdf4cpp/IRI.hpp
+++ b/src/rdf4cpp/IRI.hpp
@@ -38,7 +38,7 @@ public:
      * Constructs an IRI object from a IRI string
      * @param iri IRI string
      * @param node_storage optional custom node_storage used to store the IRI
-     * @throw std::invalid_argument if iri is invalid
+     * @throw rdf4cpp::ParsingError if iri is invalid
      */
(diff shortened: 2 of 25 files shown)


src/rdf4cpp/util/CharMatcher.hpp
```diff
@@ -1,27 +1,91 @@
...
+};
 
 template<typename T>
-concept CharMatcher = requires(const T a, int c) {{a.match(c)} -> std::same_as<bool>; };
+concept CharMatcher = requires(T const a, int c) {
+    {
+        a.match(c)
+    } -> std::same_as<bool>;
+    {
+        T::SIMD_RANGE_NUM
```
liss-h: wrong constant style

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,155 @@
...
+        if (HWY_UNLIKELY(idx == count))
+            return;
+
+        size_t const remaining = count - idx;
+        HWY_DASSERT(0 != remaining && remaining < N);
+        hwy::HWY_NAMESPACE::Vec<D> const v = LoadNOr(no, d, in + idx, remaining);
+        func(d, v);
+    }
+
+    std::optional<bool> try_match_simd_impl(std::string_view data, std::array<CharRange, 3> const &ranges, std::string_view single) {
```
liss-h: This function needs more documentation what the idea of operation is

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,155 @@
...
+        // 5 - logic
+        // ? - singles
+        // 16 overall
+        bool found_unicode = false;
+        bool r = true;
+        using D = hwy::HWY_NAMESPACE::ScalableTag<int8_t>;  //NOLINT
+        using V = hwy::HWY_NAMESPACE::VFromD<D>;             //NOLINT
+        D d;
+        bool r0a = ranges[0].first < ranges[0].last;
+        V zero = r0a ? hwy::HWY_NAMESPACE::Set(d, static_cast<int8_t>(ranges[0].first)) : hwy::HWY_NAMESPACE::Set(d, static_cast<int8_t>(single.at(0)));
```
liss-h: Is it possible to alias the `hwy::HWY_NAMESPACE`? This adds a lot of noise

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,155 @@
...
+        D d;
+        V zero = hwy::HWY_NAMESPACE::Set(d, static_cast<int8_t>(0));
+        V m0 = hwy::HWY_NAMESPACE::Set(d, static_cast<int8_t>(match[0]));
+        V m1 = hwy::HWY_NAMESPACE::Set(d, static_cast<int8_t>(match[1]));
+        V m2 = hwy::HWY_NAMESPACE::Set(d, static_cast<int8_t>(match[2]));
+        V m3 = hwy::HWY_NAMESPACE::Set(d, static_cast<int8_t>(match[3]));
+
+        Foreach(d, reinterpret_cast<int8_t const *>(data.data()), data.size(), zero, [&](auto d, auto in_vec) HWY_ATTR {
+            auto m = in_vec == m0;
+            m = hwy::HWY_NAMESPACE::Or(m, in_vec == m1);
```
liss-h: documentation here as well

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,155 @@
...
+        // register usage:
+        // 1 - data
+        // 6 - ranges
+        // 5 - logic
+        // ? - singles
+        // 16 overall
+        bool found_unicode = false;
+        bool r = true;
+        using D = hwy::HWY_NAMESPACE::ScalableTag<int8_t>;  //NOLINT
+        using V = hwy::HWY_NAMESPACE::VFromD<D>;             //NOLINT
```
liss-h: A very short comment what D and V are would be helpful, I know they say what they are in the documentation but having it here would be helpful

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,155 @@
...
+        size_t si_num = single.size();
+        if (si_num > si.size()) {
+            assert(false);
+            return std::nullopt;
+        }
+        for (size_t i = 0; i < si_num; ++i) {
+            si[i] = hwy::HWY_NAMESPACE::Set(d, static_cast<int8_t>(single[i]));
+        }
+
+        Foreach(d, reinterpret_cast<int8_t const *>(data.data()), data.size(), zero, [&](auto d, auto in_vec) HWY_ATTR {
```
liss-h, line 81: Is foreach just a badly named `fold`? I'm confused what the `zero` param does

liss-h (changes requested):
Thx for the PR.

I think the SIMD stuff needs a lot more documentation.
At least to me the library makes understanding the SIMD stuff harder than the raw C instruction equivalents, but it is what it is. At least we get dispatch.

So I hope that adding some comments and reducing noise the library introduces helps. I will have a deeper look then

src/rdf4cpp/BlankNode.cpp
```diff
@@ -1,14 +1,16 @@
...
 #include <rdf4cpp/writer/TryWrite.hpp>
+#include <uni_algo/all.h>
 
 namespace rdf4cpp {
 BlankNode::BlankNode() noexcept : Node{storage::identifier::NodeBackendHandle{{}, storage::identifier::RDFNodeType::BNode, {}}} {
 }
 
 BlankNode::BlankNode(std::string_view identifier, storage::DynNodeStoragePtr node_storage)
-    : Node{storage::identifier::NodeBackendHandle{node_storage.find_or_make_id(storage::view::BNodeBackendView{.identifier = identifier, .scope = std::nullopt}),
+    : Node{storage::identifier::NodeBackendHandle{node_storage.find_or_make_id(storage::view::BNodeBackendView{.identifier = validate_bnode_name(identifier), .scope = std::nullopt}),
```
liss-h: this could use `make_unchecked`

src/rdf4cpp/util/CharMatcher.cpp
mcb5637: zero is the placeholder that gets used, if there is no full vector of data left

mcb5637:
in preparation for #267 , also added ParsingError and throw it instead of runtime_error

src/rdf4cpp/IRI.hpp
```diff
@@ -36,6 +38,7 @@ struct IRI : Node {
      * Constructs an IRI object from a IRI string
      * @param iri IRI string
      * @param node_storage optional custom node_storage used to store the IRI
+     * @throw std::invalid_argument if iri is invalid
```
liss-h: needs to be updated to `ParsingError`

src/rdf4cpp/IRI.hpp
```diff
@@ -45,12 +48,20 @@ struct IRI : Node {
     [[nodiscard]] static IRI make_null() noexcept;
 
     /**
-     * Constructs an IRI object from a IRI string
+     * Constructs an IRI object from a IRI string.
      * @param iri IRI string
      * @param node_storage optional custom node_storage used to store the IRI
+     * @throw std::invalid_argument if iri is invalid
```
liss-h: same here

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <stdexcept>
+
+
+HWY_BEFORE_NAMESPACE();  // at file scope
+namespace rdf4cpp::util::char_matcher_detail::HWY_NAMESPACE {
+    // Calls `func(d, v)` for each input vector; out of bound lanes with index i >=
+    // `count` are instead taken from `no[i % Lanes(d)]`.
+    // if func returns false, exits early
+    // from hwy/contrib/algo/transform-inl.h
+    template<class D, class Func, typename T = hwy::HWY_NAMESPACE::TFromD<D>>
```
liss-h: typename

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <hwy/highway.h>
+#include <stdexcept>
+
+
+HWY_BEFORE_NAMESPACE();  // at file scope
+namespace rdf4cpp::util::char_matcher_detail::HWY_NAMESPACE {
+    // Calls `func(d, v)` for each input vector; out of bound lanes with index i >=
+    // `count` are instead taken from `no[i % Lanes(d)]`.
+    // if func returns false, exits early
+    // from hwy/contrib/algo/transform-inl.h
```
liss-h: Why did you copy the code here? Is this a seperate library?

src/rdf4cpp/util/CharMatcher.hpp
```diff
@@ -1,27 +1,105 @@
...
+    explicit constexpr ASCIIPatternMatcher(char const (&str)[n]) noexcept
+        : pattern(str) {
+    }
 
     [[nodiscard]] constexpr bool match(int c) const noexcept {
-        auto ch = static_cast<char>(c);
-        if (c != static_cast<int>(ch))  // not asciii
-            return false;
-        return pattern.find(ch) != std::string_view::npos;
+    auto ch = static_cast<char>(c);
```
liss-h: indentation

src/rdf4cpp/util/CharMatcher.hpp
```diff
@@ -1,27 +1,105 @@
...
+ * @return
+ */
+bool contains_any(std::string_view data, std::array<char, 4> match);
+
+/**
+ * for test cases only!
+ * iterates over each available highway target and calls func specifically with only that target selected.
+ * @param func
+ */
+void test_simd_foreach_supported(void(*func)(std::string_view target_name));
```
liss-h: Put this into the test cases if its only used there

src/rdf4cpp/util/CharMatcher.hpp
```diff
@@ -1,27 +1,105 @@
...
+    char first = '\0';
+    char last = '\0';
+};
 
 template<typename T>
-concept CharMatcher = requires(const T a, int c) {{a.match(c)} -> std::same_as<bool>; };
+concept CharMatcher = requires(T const a, int c) {
+    {
+        a.match(c)
+    } -> std::same_as<bool>;
```
liss-h: `convertible_to`

src/rdf4cpp/util/CharMatcher.hpp
```diff
@@ -1,27 +1,105 @@
...
+    {
+        T::simd_range_num
+    } -> std::same_as<const size_t &>;
```
liss-h: `convertible_to<size_t>`

src/rdf4cpp/util/CharMatcher.hpp
```diff
@@ -1,27 +1,105 @@
...
+    {
+        T::fail_if_unicode
+    } -> std::same_as<const bool &>;
```
liss-h: `convertible_to<bool>`

src/rdf4cpp/util/CharMatcher.hpp
```diff
@@ -1,27 +1,105 @@
...
+    {
+        a.simd_ranges()
+    } -> std::same_as<std::array<CharRange, T::simd_range_num>>;
+    {
+        a.simd_singles()
+    };  // -> std::same_as<std::string_view>;
```
liss-h: Why is this commented out?

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,211 @@
...
+        return try_match_simd_impl(data, ranges, single);
+    }
+    std::optional<bool> try_match_simd_impl_3_21(std::string_view data, std::array<CharRange, 3> const &ranges, datatypes::registry::util::ConstexprString<21> const &single) {
+        return try_match_simd_impl(data, ranges, single);
+    }
+    std::optional<bool> try_match_simd_impl_1_1(std::string_view data, std::array<CharRange, 1> const &ranges, datatypes::registry::util::ConstexprString<1> const &single) {
+        return try_match_simd_impl(data, ranges, single);
+    }
+
+    bool contains_any_impl(std::string_view data, std::array<char, 4> match) {
```
liss-h: why is `match` not simply `string_view`?

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,211 @@
...
+        bool r = false;
+
+        using D = ScalableTag<int8_t>;  //NOLINT  tag type, selects the used vector type
+        using V = Vec<D>;             //NOLINT vector type
+        D const d;
+
+        // elements from this vector are used, if data is not a multiple of Lanes(d)
+        // should not influence final result
+        // => comparisons need to evaluate as false for the logic to work
+        V const zero = Set(d, static_cast<int8_t>(0));
```
liss-h, line 139: So, we need to `assert` that `match` does not contain `\0`

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+        Foreach(d, reinterpret_cast<int8_t const *>(data.data()), data.size(), zero, [&](auto d, auto in_vec) HWY_ATTR {
+            // if bit 7 is set, the char belongs to a unicode sequence
+            // => std::nullopt
+            if (!AllTrue(d, HighestSetBitIndex(in_vec) < unicode_bit_index)) {
+                found_unicode = true;
+                return false;
+            }
+
+            // highway doc: on x86 < and > are 1 instruction for signed ints (3 for unsigned)
```
liss-h: this comment would be more useful higher up where the ranges are set-up

liss-h (changes requested):
Thanks for the changes, much more readable now. I've left some additional comments.

The thing with the templates is really ugly, but I couldn't find a way to make it work with the export macro either. I'm pondering if it makes sense to use a `static thread_local std::vector` for the required buffers (or maybe `flex_array`) to avoid the template all together, but idk

src/rdf4cpp/util/CharMatcher.cpp
mcb5637: the highway one does not allow for early return. added a comment explaining it.

src/rdf4cpp/util/CharMatcher.cpp
mcb5637: because string_view is arbitrary length.
and the code is more efficient, if i know the length at compile time.

liss-h (approved):

PR #266, opened 2024-04-03
Fix: Node serialization segfault

mcb5637:
something like `std::cout << IRI{};` would segfault, because serialization was missing null checks.
affected are `IRI`, `BlankNode`, `Variable` and `Node`

### DIFF
--- a/src/rdf4cpp/BlankNode.cpp
+++ b/src/rdf4cpp/BlankNode.cpp
@@ -57,8 +57,7 @@ std::string_view BlankNode::identifier() const noexcept { return handle_.bnode_b
 
 bool BlankNode::serialize(writer::BufWriterParts const writer) const noexcept {
     if (null()) {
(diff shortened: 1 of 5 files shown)


src/rdf4cpp/BlankNode.cpp
```diff
@@ -56,6 +56,11 @@ BlankNode BlankNode::find(std::string_view identifier, storage::DynNodeStoragePt
 std::string_view BlankNode::identifier() const noexcept { return handle_.bnode_backend().identifier; }
 
 bool BlankNode::serialize(writer::BufWriterParts const writer) const noexcept {
+    if (null()) {
```
liss-h, line 59: I think the version from `Node` is better (i.e. `return writer::write_str("null", writer)`)

liss-h (changes requested):

liss-h (approved):

PR #271, opened 2024-04-24
Fix: decimal inlining

liss-h:
Fixes 2 bugs with decimal inlining:
1. negative values were decoded incorrectly
2. decimals were not normalized, leading to different inlined representations for the same number

I basically rewrote the code from scratch, as a bonus I think its easier to understand now

### DIFF
--- a/tests/datatype/tests_Decimal.cpp
+++ b/tests/datatype/tests_Decimal.cpp
@@ -161,6 +161,22 @@ TEST_CASE("decimal inlining sanity check") {
 
         auto const l2 = Literal::make_typed<Decimal>("-1.0");
         CHECK_EQ(l2.value<Decimal>(), -1);
+
+        auto const l3 = Literal::make_typed<Decimal>("1.984612364642233");
+        CHECK(!l3.is_inlined());
+        CHECK_EQ(l3.value<Decimal>(), Decimal::cpp_type{"1.984612364642233"});
+
+        auto const l4 = Literal::make_typed<Decimal>("-1.984612364642233");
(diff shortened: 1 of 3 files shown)


tests/datatype/tests_Decimal.cpp
```diff
@@ -151,11 +151,56 @@ TEST_CASE("Datatype Decimal buffer overread UB") {
...
+            CHECK(l2.value<Decimal>() == Decimal::cpp_type{-very_big_value, 1U});
+
+            // right at the inlining limit
+            auto const l3 = Literal::make_typed_from_value<Decimal>(Decimal::cpp_type(1L << 32, 0));
+            CHECK(!l3.is_inlined());
+            CHECK(l3.value<Decimal>() == Decimal::cpp_type(1L << 32, 0));
+
+            auto const l4 = Literal::make_typed_from_value<Decimal>(Decimal::cpp_type(-1L << 32, 0));
+            CHECK(!l4.is_inlined());
+            CHECK(l4.value<Decimal>() == Decimal::cpp_type(-1L << 32, 0));
```
bigerl: Please also check just below the threshold where it is just still inlined. 
Ideally, this should include: positive/negative integer as well as positive and negative decimal number of the kind 1.9841...

bigerl (changes requested):
Thx for the PR. Code actually is easier to understand. Still, with exponents, mantissa and normalisation, it takes a moment to understand what's going on. 

I think we should add some more test cases. Otherwise, this looks good to me.

bigerl (approved):

PR #276, opened 2024-05-23
Fix: reduce requirements for node storages

liss-h:
Reduces the requirements for the `NodeStorage` concept down to the bare minimum that is necessary. This has the benefit that node storages have to implement less features to qualify as `NodeStorage`s.

### DIFF
--- a/src/rdf4cpp/storage/NodeStorage.cpp
+++ b/src/rdf4cpp/storage/NodeStorage.cpp
@@ -1,16 +1,14 @@
-#include "NodeStorage.hpp"
-
+#include <rdf4cpp/storage/NodeStorage.hpp>
 #include <rdf4cpp/storage/reference_node_storage/SyncReferenceNodeStorage.hpp>
 
 namespace rdf4cpp::storage {
 
-#ifndef DOXYGEN_PARSER
-reference_node_storage::SyncReferenceNodeStorage default_node_storage_holder_;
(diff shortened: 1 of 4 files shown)


src/rdf4cpp/storage/NodeStorage.cpp
```diff
@@ -1,16 +1,14 @@
...
+
 #ifndef DOXYGEN_PARSER
-reference_node_storage::SyncReferenceNodeStorage default_node_storage_holder_;
-DynNodeStoragePtr default_node_storage{default_node_storage_holder_};
+DynNodeStoragePtr default_node_storage{reference_node_storage::default_instance};
 #endif // DOXYGEN_PARSER
 
-void reset_default_node_storage() noexcept {
-    default_node_storage = default_node_storage_holder_;
-}
```
nkaralis, line 14: Is this function not needed anymore?

nkaralis (commented):
Thank you for the PR. I only have one question.

src/rdf4cpp/storage/NodeStorage.cpp
liss-h: To achieve the same effect, you can now do
```c++
rdf4cpp::storage::default_node_storage = rdf4cpp::storage::reference_node_storage::default_instance;
```

PR #277, opened 2024-05-24
datatype parsing error

### LINKED ISSUE #267: Add parse_error type for datatypes
Currently datatypes just throw `std::runtime_error` on failing to parse.
This makes it basically impossible to catch just rdf4cpp parse errors.
We should have a dedicated exception type for parse errors.

mcb5637:
close #267 
also renamed `rdf4cpp::ParsingError` to `rdf4cpp:InvalidNode`, because it might get confused with  `rdf4cpp::parser::ParsingError`.


src/rdf4cpp/BigDecimal.hpp
```diff
@@ -123,8 +124,7 @@ struct BigDecimal {
      * parses a BigDecimal from a string_view.
      * may include a leading sign and one decimal point ., everything else needs to be numeric.
      * @param value
-     * @throw std::invalid_argument if a invalid char is found
-     * @throw std::overflow_error on exceeding the types numeric limits
+     * @throw ParsingError if a invalid char is found, or on exceeding the types numeric limits
```
liss-h: InvalidNode

src/rdf4cpp/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -68,7 +70,7 @@ bool to_chars_canonical(I const value, writer::BufWriterParts const writer) noex
  * @tparam F the result floating point type
  * @param s the string to be parsed
  * @return the resulting value
- * @throws std::runtime_error if the string cannot be parsed
+ * @throws ParsingError if the string cannot be parsed
```
liss-h: InvalidNode

src/rdf4cpp/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -17,7 +19,7 @@ namespace rdf4cpp::datatypes::registry::util {
  * @tparam I the resulting integral type
  * @param s the string to be parsed
  * @return the resulting value
- * @throws std::runtime_error if the string cannot be parsed
+ * @throws ParsingError if the string cannot be parsed
```
liss-h: InvalidNode

src/rdf4cpp/datatypes/owl/Real.cpp
```diff
@@ -8,7 +9,7 @@ capabilities::Default<owl_real>::cpp_type capabilities::Default<owl_real>::from_
     try {
         return cpp_type{s};
     } catch (std::runtime_error const &e) {
-        throw std::runtime_error{std::string{"owl:real parsing error:"} + e.what()};
+        throw InvalidNode{std::string{"owl:real parsing error:"} + e.what()};
```
liss-h: missing space after `[..] error:`

src/rdf4cpp/datatypes/owl/Rational.cpp
```diff
@@ -19,18 +20,18 @@ template<>
...
         // owl:rational does not allow hex, boost does
-        throw std::runtime_error{"owl:rational parsing error: invalid character in string"};
+        throw InvalidNode{"owl:rational parsing error: invalid character in string"};
     }
 
     try {
         return cpp_type{s};
     } catch (std::runtime_error const &e) {
-        throw std::runtime_error{std::string{"owl:rational parsing error:"} + e.what()};
+        throw InvalidNode{std::string{"owl:rational parsing error:"} + e.what()};
```
liss-h: missing space after `[..] error:`

src/rdf4cpp/InvalidNode.hpp
```diff
@@ -0,0 +1,17 @@
+#ifndef RDF4CPP_INVALIDNODE_HPP
+#define RDF4CPP_INVALIDNODE_HPP
+
+#include <stdexcept>
+
+namespace rdf4cpp {
+    struct InvalidNode : std::runtime_error {
+        explicit inline InvalidNode(std::string const &msg)
```
liss-h: `inline` is implicit (because def is in struct)

liss-h (changes requested):
Thanks for the PR, I have two general points:

1. I think it makes sense to seperate the parsing error and this one. I was just thinking if it makes sense to still somehow make them related, or maybe add a general error type to rdf4cpp that all others must inherit from (so that users can just catch `rdf4cpp::Error` and it will catch everything from rdf4cpp. Opionion or other ideas?

2. In the parser, we should probably now catch `InvalidNode` instead of `runtime_error` on node construction.

liss-h (changes requested):
You didn't address the first review fully, see header comment

mcb5637:
Using a common base class for InvalidNode and ParsingError might not be the best idea, because ParsingError does not inherit from std::exception and is not thrown.

liss-h:
> Using a common base class for InvalidNode and ParsingError might not be the best idea, because ParsingError does not inherit from std::exception and is not thrown.

good point

PR #278, opened 2024-05-24
regex replace fixes

### LINKED ISSUE #132: Bug: Regex replace behaviour when matching the empty string
https://www.w3.org/TR/sparql11-query/#func-replace states:
```
The expression fn:replace("abracadabra", ".*?", "$1") raises an error, 
because the pattern matches the zero-length string
```

This is currently not the case as the used regex implementation (RE2) does not behave this way.
The current result is the empty string instead.

### LINKED ISSUE #141: Bug: Regex replace stray backslashes break replacement
For an yet unknown reason this currently passes:
```c++
Regex r{"[0-9]"};

std::string s = "Hello 1 World";
r.make_replacer(R"(Hello \hgfhf World)").regex_replace(s);

CHECK(s == "Hello Hello  World");
```

I.e. It doesn't seem to matter what comes after the backslash if it isn't another backslash or a number the rest of the replacement just gets ignored. This seems to be a problem with RE2 and needs to be investigated further.

mcb5637:
close #141 
close #132 

```
In the absence of the q flag, a dynamic error is raised [[err:FORX0004](https://www.w3.org/TR/xpath-functions/#ERRFORX0004)] if the value of $replacement contains a backslash (\) character that is not part of a \\ pair, unless it is immediately followed by a dollar sign ($) character.
```

### DIFF
--- a/private/rdf4cpp/regex/RegexReplacerImpl.cpp
+++ b/private/rdf4cpp/regex/RegexReplacerImpl.cpp
@@ -76,6 +76,10 @@ RegexReplacer::Impl::Impl(Regex::Impl const &regex, std::string_view const rewri
                                                                                       rewrite{regex.flags.contains(RegexFlag::Literal)
                                                                                                       ? rewrite
                                                                                                       : detail::translate_rewrite(rewrite)} {
+    std::string err{};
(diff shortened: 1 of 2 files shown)


liss-h (approved):
Thanks, that was an unexpectedly straight forward fix

PR #281, opened 2024-05-29
Fix: simplify bnode management

liss-h:
When #130 introduced blank node management, we misunderstood our own requirements of the feature, which resulted in a massively overcomplicated solution. Now that we've actually had a look a how query processing needs to happen, we've decided to drop the excess complexity in favour of a much simplified approach.

Summary of the changes:
- Drop the inclusion of the node scope in the blank node backend. This was never required, we misunderstood what the spec was trying to say.
  - related: drop `BlankNode::merge_eq` and `BlankNode::union_eq`, these are now handled at parsing-time (see `tests_BlankNodeIdManagement.cpp` subcase `file parsing`)
- The previous change allowed us to drop all code that was previously required to make blank nodes and their scopes persistable together
  - note: a persistable scope can still be implemented with the current mechanism
- Drop inheritance for blank node management related interfaces in favor of `Dyn[X]Ptr` (this achives interface uniformity between node storages and blank node management)
- Drop unecessary sub interfaces in bnode management (e.g. `INodeFactory`)
- Reduce requirements for basically all interfaces
- Remove the `SkolemIRIGenerator`. I think this one is a little to specific for each usecase, so it does not make sense for us to provide it. It is also very easy to implement now (I've done so in an example) so its not a problem that we do not provide an implemention anymore.


private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -36,28 +36,12 @@ nonstd::expected<Node, SerdStatus> IStreamQuadIterator::Impl::get_bnode(std::str
...
-    try {
-        assert(this->state->blank_node_generator != nullptr);
-        assert(this->state->blank_node_scope_manager != nullptr);
 
-        return this->state->blank_node_scope_manager->scope(graph_str).get_or_generate_node(node_str,
-                                                                                            *this->state->blank_node_generator,
-                                                                                            this->state->node_storage);
+        return this->state->blank_node_scope_manager.scope(graph_str).get_or_generate_node(node_str, this->state->node_storage);
     } catch (std::runtime_error const &e) {
         // TODO: check when actual blank node validation implemented
```
nkaralis: is this TODO still up-to-date?

src/rdf4cpp/Dataset.hpp
```diff
@@ -151,14 +151,18 @@ struct Dataset {
...
+                       ErrF &&errf = {}) requires std::invocable<decltype(errf), parser::ParsingError> {
 
-        parser::ParsingState state{.node_storage = node_storage_};
-        parser::RDFFileParser parser{file_path, flags, &state};
+        if (state != nullptr && state->node_storage != node_storage_) {
+            throw std::invalid_argument{"NodeStorage of the parsing state must be the same as NodeStorage of the Dataset"};
+        }
 
-        for (auto const &quad : parser) {
+        parser::IStreamQuadIterator parser{rdf_file, flags, state};
```
nkaralis, line 163: what is the reason of directly using IStreamQuadIterator over RDFFileParser?

src/rdf4cpp/bnode_mngt/reference_backends/generator/IncreasingIdGenerator.cpp
```diff
@@ -11,20 +11,21 @@ namespace generator_detail {
...
-size_t IncreasingIdGenerator::max_generated_id_size() const noexcept {
-    return generator_detail::max_generated_id_size + this->prefix.size();
-}
+BlankNode IncreasingIdGenerator::generate(storage::DynNodeStoragePtr node_storage) noexcept {
+    static thread_local std::string buf;
 
-char *IncreasingIdGenerator::generate_to_buf(char *buf) {
-    buf = std::copy(this->prefix.begin(), this->prefix.end(), buf);
+    buf.clear();
+    buf.resize(prefix_.size() + generator_detail::max_generated_id_size);
```
nkaralis, line 23: is this to avoid constructing a new string every time this function is called?

src/rdf4cpp/bnode_mngt/reference_backends/generator/IncreasingIdGenerator.cpp
```diff
@@ -11,20 +11,21 @@ namespace generator_detail {
...
 
-    std::to_chars_result const res = std::to_chars(buf, buf + generator_detail::max_generated_id_size, this->counter.fetch_add(1, std::memory_order_relaxed));
+    auto write_it = std::copy(prefix_.begin(), prefix_.end(), buf.data());
+    std::to_chars_result const res = std::to_chars(write_it, write_it + generator_detail::max_generated_id_size, counter_.fetch_add(1, std::memory_order_relaxed));
     assert(res.ec == std::errc{});
-    return res.ptr;
+
+    return BlankNode::make(std::string_view{buf.data(), static_cast<size_t>(res.ptr - buf.data())}, node_storage);
```
nkaralis: in the `RandomIdGenerator`, you use `make_unchecked`. Could it be used here as well?

src/rdf4cpp/bnode_mngt/reference_backends/scope/ReferenceNodeScope.hpp
```diff
@@ -55,15 +29,55 @@ struct ReferenceNodeScope : INodeScope {
...
+    ReferenceNodeScope(GenArgs &&...gen_args) noexcept(std::is_nothrow_constructible_v<G, decltype(std::forward<GenArgs>(gen_args))...>)
+        : gen_{std::forward<GenArgs>(gen_args)...} {
+    }
+
+    [[nodiscard]] G const &generator() const noexcept {
+        return gen_;
+    }
+
+    [[nodiscard]] Node try_get_node(std::string_view label) const noexcept {
+        if (auto it = label_to_handle_.find(label); it != label_to_handle_.end()) {
```
nkaralis, line 44: do we need to get a shared lock on the mutex here?

src/rdf4cpp/bnode_mngt/reference_backends/scope/ReferenceNodeScope.hpp
```diff
@@ -55,15 +29,55 @@ struct ReferenceNodeScope : INodeScope {
...
-    ReferenceNodeScope() = default;
-    ReferenceNodeScope(ReferenceNodeScope &&other) noexcept;
+    template<typename ...GenArgs>
+    ReferenceNodeScope(GenArgs &&...gen_args) noexcept(std::is_nothrow_constructible_v<G, decltype(std::forward<GenArgs>(gen_args))...>)
+        : gen_{std::forward<GenArgs>(gen_args)...} {
+    }
+
+    [[nodiscard]] G const &generator() const noexcept {
+        return gen_;
+    }
```
nkaralis, line 40: is there a use case where we need to access the underlying generator?

src/rdf4cpp/bnode_mngt/reference_backends/scope_manager/MergeNodeScopeManager.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <rdf4cpp/bnode_mngt/NodeScopeManager.hpp>
+#include <rdf4cpp/bnode_mngt/reference_backends/scope/ReferenceNodeScope.hpp>
+#include <dice/sparse-map/sparse_map.hpp>
+
+#include <dice/hash.hpp>
+
+namespace rdf4cpp::bnode_mngt {
+
+template<NodeScope S = ReferenceNodeScope<>> requires std::is_default_constructible_v<S>
```
nkaralis, line 12: this defaults to the `RandomIdGenerator`, right?

src/rdf4cpp/parser/ParsingState.hpp
```diff
@@ -51,10 +41,15 @@ struct ParsingState {
     /**
      * The node storage to put the parsed quads into
      */
-    node_storage_type node_storage = storage::default_node_storage;
+    storage::DynNodeStoragePtr node_storage = storage::default_node_storage;
```
nkaralis, line 42: can we use the alias defined above (`node_storage_type`)?

nkaralis (commented):
Thank you for the PR. I left some comments/questions.

src/rdf4cpp/Dataset.hpp
liss-h: Is more general (because you can use any istream) and was needed for the tests

src/rdf4cpp/bnode_mngt/reference_backends/scope/ReferenceNodeScope.hpp
liss-h: not sure tbh, just added it because it might be useful

src/rdf4cpp/bnode_mngt/reference_backends/generator/IncreasingIdGenerator.cpp
liss-h: No, because the prefix might be invalid.. but it might make sense to move that check to the constructor

src/rdf4cpp/parser/ParsingState.hpp
liss-h: was planning to remove that one. Apparently forgot to do that

private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
liss-h: no, will remove

src/rdf4cpp/bnode_mngt/reference_backends/generator/IncreasingIdGenerator.cpp
liss-h: yes, this way we only have one heap allocation per thread instead of one per generated node

src/rdf4cpp/bnode_mngt/reference_backends/generator/IncreasingIdGenerator.cpp
liss-h: Moved the check to the ctor

nkaralis (approved):

liss-h:
Pasting the relevant standardese for future reference here (compiled by @nkaralis):
```
SPARQL Query Standard
    * An application writer should not expect blank node labels in a query to refer to a particular blank node in the data.

    * Blank nodes in graph patterns act as variables, not as references to specific blank nodes in the data being queried.

    * The RDF dataset includes an RDF merge of the named graphs in the default graph, re-labeling blank nodes to keep them distinct.

    * In an RDF merge, blank nodes in the merged graph are not shared with blank nodes from the graphs being merged.

    * The actions required to construct the dataset are not determined by the dataset description alone. If an IRI is given twice in a dataset description, either by using two FROM clauses, or a  FROM clause and a FROM NAMED clause, then it does not assume that exactly one or exactly two attempts are made to obtain an RDF graph associated with the IRI. Therefore, no assumptions can be made about blank node identity in triples obtained from the two occurrences in the dataset description. In general, no assumptions can be made about the equivalence of the graphs.

    * CONSTRUCT Template: A template can create an RDF graph containing blank nodes. The blank node labels are scoped to the template for each solution. If the same label occurs twice in a template, then there will be one blank node created for each query solution, but there will be different blank nodes for triples generated by different query solutions.

    * BNODE() SPARQL Function: The BNODE function constructs a blank node that is distinct from all blank nodes in the dataset being queried and distinct from all blank nodes created by calls to this constructor for other query solutions. If the no argument form is used, every call results in a distinct blank node. If the form with a simple literal is used, every call results in distinct blank nodes for different simple literals, and the same blank node for calls with the same simple literal within expressions for one solution mapping. This functionality is compatible with the treatment of blank nodes in SPARQL CONSTRUCT templates.

    * Blank node labels are scoped to the SPARQL Request String in which they occur. Different uses of the same blank node label in a request string refer to the same blank node. Fresh blank nodes are generated for each request; blank nodes can not be referenced by label across requests.

SPARQL Update Standard
    * Variables in QuadDatas are disallowed in INSERT DATA requests (see Notes 8 in the grammar). That is, the INSERT DATA statement only allows to insert ground triples. Blank nodes in QuadDatas are assumed to be disjoint from the blank nodes in the Graph Store, i.e., will be inserted with "fresh" blank nodes.

    * Blank nodes that appear in an INSERT clause operate similarly to blank nodes in the template of a CONSTRUCT query, i.e., they are re-instantiated for any solution of the WHERE clause

RDF Concepts and Abstract Syntax
    * Blank node identifiers are local identifiers that are used in some concrete RDF syntaxes or RDF store implementations. They are always locally scoped to the file or RDF store, and are not persistent or portable identifiers for blank nodes.

    * SPARQL 1.1 also defines the concept of an RDF Dataset. The definition of an RDF Dataset in SPARQL 1.1 and this specification differ slightly in that this specification allows RDF Graphs to be identified using either an IRI or a blank node. SPARQL 1.1 Query Language only allows RDF Graphs to be identified using an IRI. Existing SPARQL implementations might not allow blank nodes to be used to identify RDF Graphs for some time, so their use can cause interoperability problems. Skolemizing blank nodes used as graph names can be used to overcome these interoperability problems.
```

PR #282, opened 2024-06-03
upgrade to conan 2

mcb5637:
TODO:
- [x] point to template-library release


liss-h:
I think it makes sense to try this further up the chain, before merging. When everything works there we can merge this

liss-h (approved):

Issue #243, opened 2024-02-09
Clean seperation for `reserved_datatype_ids`


liss-h:
As rediscovered in #238, `datatypes::registry::reserved_datatype_ids` currently contains IRIs that are not in fact related to datatypes. Namely, `rdf_type` and `default_graph_iri`. These should be moved to a separate structure so that they do not take up values of `LiteralType`.

These changes might require some other changes in the `IRI NodeID` <-> `LiteralType` conversions.
And might also require the increase of the minium Literal ID. This change will definitely break binary compatibility for persistent node storages.

nkaralis:
What's the priority of this one @liss-h?

liss-h:
currently, low

PR #285, opened 2024-06-20
Fix: avoid default node storage access in `write_quad`

liss-h:
Avoid default node storage access in `write_quad`

### DIFF
--- a/private/rdf4cpp/writer/WriteQuad.hpp
+++ b/private/rdf4cpp/writer/WriteQuad.hpp
@@ -41,7 +41,7 @@ bool write_pred(Node const pred, writer::BufWriterParts const writer) {
 
 [[nodiscard]] inline bool is_default_graph(Node const &graph) noexcept {
     auto const g = graph.as_iri();
-    return !g.null() && g.is_default_graph();
+    return g.is_default_graph();
 }
 
 template<writer::OutputFormat F, typename Q>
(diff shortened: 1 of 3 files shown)


private/rdf4cpp/writer/WriteQuad.hpp
```diff
@@ -39,6 +39,11 @@ bool write_pred(Node const pred, writer::BufWriterParts const writer) {
         return false;                       \
     }
 
+[[nodiscard]] inline bool is_default_graph(Node const &graph) noexcept {
+    auto const g = graph.as_iri();
+    return !g.null() && g.is_default_graph();
```
bigerl: why check that it is not the null node?

src/rdf4cpp/IRI.cpp
```diff
@@ -140,8 +140,17 @@ bool IRI::is_iri() const noexcept { return true; }
 
 
 IRI IRI::default_graph(storage::DynNodeStoragePtr node_storage) {
-    return IRI::make_unchecked("", node_storage);
+    auto const id = datatypes::registry::reserved_datatype_ids[datatypes::registry::default_graph_iri];
```
bigerl, line 143: Note to myself: this is strange because of https://github.com/rdf4cpp/rdf4cpp/issues/243

bigerl (approved):
Thx for the PR. Looks good. I have a little remark below but I am sure that is non-blocking

private/rdf4cpp/writer/WriteQuad.hpp
liss-h: Good point, might as well directly call the function

bigerl: patched

PR #292, opened 2024-07-12
prevent predefined IRIs from being erased

### LINKED ISSUE #201: Protect predefined IRIs from being erased from NodeStorage
It is possible to acidentally erase the predefined IRIs from the node storage, which causes bugs. There needs to be a check in the node storage that prevents this

mcb5637:
close #201 
TODO:

- [x] can the check get simplified?


liss-h:
> can the check get simplified?

Yes:
```c++
if (identifier::iri_node_id_to_literal_type(id).is_fixed()) {
    return false;
}
```

liss-h (approved):

PR #295, opened 2024-07-24
Feature: more serialization control

### LINKED ISSUE #294: `Literal::serialize_short_form` should not shorten prefixes
We should introduce a third literal serialization function for that and let `serialize_short_form` only shorten integer, decimal, double, ... but never use prefixes. This is useful for sparql TSV results because they are allowed to contain shortened literals but cannot define prefixes.

Examples for clarification:
```c++
1_xsd_integer.serialize_short_form() -> 1  // good
1.0_xsd_double.serialize_short_form() -> 1.0E0 // good
-1_xsd_negative_integer.serialize_short_form() -> "-1"^^xsd:negativeInteger // bad
```

Suggested functions:
```c++
Literal::serialize()
// 1_xsd_integer -> "1"^^<http://...url...>
// -1_xsd_negative_integer -> "-1"^^<http://...url...>

Literal::serialize_short_form()
// 1_xsd_integer -> 1
// -1_xsd_negative_integer -> "-1"^^<http://...url...>

Literal::serialize_short_form_with_prefixes()
// 1_xsd_integer -> 1
// -1_xsd_negative_integer -> "-1"^^xsd:negativeInteger
```

liss-h:
Fix #294

### DIFF
--- a/src/rdf4cpp/Node.hpp
+++ b/src/rdf4cpp/Node.hpp
@@ -18,6 +18,57 @@ namespace query {
 struct Variable;
 } // namespace rdf4cpp
 
+/**
+ * Options to control the serialization of Nodes/Literals
+ */
+struct NodeSerializationOpts {
+    bool use_short_form; ///< use the short form of the literals, for example: if this option is true "1"^^xsd:integer serializes to 1
+    bool use_prefixes;   ///< use prefixes for known datatypes, for example: if this option is true "1"^^xsd:byte serializes to "1"^^xsd:byte
+
+    /**
+     * Do not use short forms and do not use prefixes. This is used in N-Triples for example.
+     *
+     * Examples:
+     *     "1"^^xsd:integer serializes to "1"^^<http://www.w3.org/2001/XMLSchema#integer>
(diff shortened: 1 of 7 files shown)


src/rdf4cpp/Node.hpp
```diff
@@ -18,6 +18,57 @@ namespace query {
 struct Variable;
 } // namespace rdf4cpp
 
+/**
+ * Options to control the serialization of Nodes/Literals
+ */
+struct NodeSerializationOpts {
+    bool use_short_form; ///< use the short form of the literals, for example: if this option is true "1"^^xsd:integer serializes to 1
+    bool use_prefixes;   ///< use prefixes for known datatypes, for example: if this option is true "1"^^xsd:byte serializes to "1"^^xsd:byte
```
bigerl: state the exact rule here and list the types that are used by default (xsd, rdf).

bigerl (changes requested):
Thx for the PR. Things look good. We only need to make more clear what `use_refixed` does exactly (see comment below).

PR #297, opened 2024-07-29
Fix: serd buffer overread

liss-h:
Updates serd to resolve a buffer overread when reading rdf files without a newline at the end.

While this issue is not yet resolved in serd. I've implemented a mitigation in our serd fork that should hopefully fix the issue (https://github.com/dice-group/serd/commit/ea27f19617d039a9b37c87e799b18a7fac101389)

### DIFF
--- a/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -336,7 +336,10 @@ std::optional<nonstd::expected<IStreamQuadIterator::ok_type, IStreamQuadIterator
             // handle error from last time
             if (this->last_error_requires_skip) {
                 this->last_error_requires_skip = false;
-                serd_reader_skip_until_byte(this->reader, '\n');
+                if (serd_reader_skip_until_byte(this->reader, '\n') != SERD_SUCCESS) {
+                    // EOF reached
+                    this->end_flag = true;
+                }
             }
             return nonstd::make_unexpected(*std::exchange(this->last_error, std::nullopt));
         } else if (this->end_flag) {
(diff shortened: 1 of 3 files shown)


liss-h:
~~Did not help yet, waiting for https://gitlab.com/drobilla/serd/-/issues/30~~

private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -336,7 +336,10 @@ std::optional<nonstd::expected<IStreamQuadIterator::ok_type, IStreamQuadIterator
             // handle error from last time
             if (this->last_error_requires_skip) {
                 this->last_error_requires_skip = false;
-                serd_reader_skip_until_byte(this->reader, '\n');
+                if (serd_reader_skip_until_byte(this->reader, '\n') != SERD_SUCCESS) {
```
bigerl, line 339: Good catch. This was actually a bug on our side.

tests/parser/tests_IStreamQuadIterator.cpp
```diff
@@ -609,4 +610,55 @@ TEST_SUITE("IStreamQuadIterator") {
...
+                }
+            }
+        };
+
+        SUBCASE("less than one chunk") {
+            run_overread(100);
+        }
+
+        SUBCASE("exactly one chunk") {
+            run_overread(4096);
```
bigerl, line 649: Where did you find serd's chunk size? If they have a constant, it would be good to use that.

bigerl (approved):
Thx for the quick fix. If you know where the chunk size is defined, please use the constant. Otherwise, this is good to go.

tests/parser/tests_IStreamQuadIterator.cpp
liss-h: We define the chunk size in the implementation of IStreamQuadIterator

PR #299, opened 2024-08-05
Fix: serd bug by updating serd

liss-h:
Continuation of #297 but this time pointing to a proper fix instead of a mitigation


bigerl (approved):
Thx for the PR. Hash is correct, tests pass. So, this is ready for merging.

PR #300, opened 2024-08-07
iterators for IRIFactory

nkaralis:


tests/nodes/tests_IRIFactory.cpp
```diff
@@ -208,6 +208,28 @@ TEST_CASE("relative prefix") {
     fact.clear_prefix("pre");
     CHECK(fact.from_prefix("pre", "bar").error() == IRIFactoryError::UnknownPrefix);
 }
+
+TEST_CASE("IRIFactor iterators") {
```
liss-h: that is a very weird way to write this test. Why not collect the prefixes from the factory into a `std::set<std::pair<std::string, std::string>>` and check for range equality with the expected result?

liss-h (changes requested):
Thx. Apart from the test this looks fine

tests/nodes/tests_IRIFactory.cpp
liss-h: Sorry, that wasn't quite what I meant. I've implemented what I meant, faster than communicating 😅

liss-h (approved):

PR #303, opened 2024-08-29
Fix: crash on invalid lang str

liss-h:
Fixes a bug where applications would unexpectedly crash when they encounted a lang string in "typed"-form `"AAA"^^rdf:langString`.

The issue was the wrong exception type being raised, which was subsequently not caught.


bigerl (approved):
Thx for the fix. Changes look good. I also didn't find any other uncaught exceptions that might get past the exception handling of the IStreamQuadIterator.

PR #304, opened 2024-09-04
copy license files

### LINKED ISSUE #78: Copy license files properly with conan
License files from all dependencies should be collected an provided with the built binaries. We probably need to add the names of the respective libraries to the license file name to avoid file name clashes among the license files (They will mostly be named `LICENSE` or `LICENSE.md` I suppose)

mcb5637:
close #78 

why double copy? `At the moment, this information should only be used in generate() and validate() methods. ` (https://docs.conan.io/2/reference/conanfile/methods/generate.html)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,3 +124,4 @@ environment_run.sh.env
 # End of https://www.toptal.com/developers/gitignore/api/c++,conan,ninja,cmake,clion+all
 /CMakeUserPresets.json
 /conan_provider.cmake
+/licenses/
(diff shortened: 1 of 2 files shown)


liss-h (changes requested):
Hi, sorry I talked with alex and after thinking about it it does not make sense to copy the license files here, because this is not our complete dependency graph (and it might change through overrides).

So this probably does not make sense here. But please double check that the `serd` license file is copied properly into the conan package (and that is has a good name, something like LICENSE_serd or something)

liss-h:
superseded by #325

PR #305, opened 2024-09-05
Fix: allow to_node_storage & try_get_in_node_storage on null nodes

liss-h:
This is very easy to get wrong and resulted in a crash before this PR


bigerl (approved):
Good catch. Thx for the PR. Changes are trivial and look good.

PR #306, opened 2024-09-09
Fix: bnode and variable validation

liss-h:
title

### DIFF
--- a/src/rdf4cpp/BlankNode.cpp
+++ b/src/rdf4cpp/BlankNode.cpp
@@ -91,7 +91,7 @@ std::ostream &operator<<(std::ostream &os, BlankNode const &bnode) {
 
 void BlankNode::validate(std::string_view v) {
     using namespace util::char_matcher_detail;
-    static constexpr auto first_matcher = ASCIINumMatcher{} | PNCharsBaseMatcher;
+    static constexpr auto first_matcher = ASCIINumMatcher{} | PNCharsUMatcher;
     auto r = v | una::views::utf8;
     auto it = r.begin();
     if (it == r.end()) {
@@ -102,7 +102,7 @@ void BlankNode::validate(std::string_view v) {
     }
     auto lastchar = *it;
     ++it;
-    static constexpr auto pn_matcher = PNCharsMatcher | ASCIIPatternMatcher(".-");
+    static constexpr auto pn_matcher = PNCharsMatcher | ASCIIPatternMatcher{"."};
     while (it != r.end()) {
         if (!pn_matcher.match(*it))
         {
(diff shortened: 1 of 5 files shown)


src/rdf4cpp/BlankNode.cpp
```diff
@@ -102,7 +102,7 @@ void BlankNode::validate(std::string_view v) {
     }
     auto lastchar = *it;
     ++it;
-    static constexpr auto pn_matcher = PNCharsMatcher | ASCIIPatternMatcher(".-");
```
bigerl, line 105: Why did this not work? From what I can see it should have accepted `-` through the ASCIIPatternMatcher

src/rdf4cpp/query/Variable.cpp
```diff
@@ -120,9 +124,10 @@ void Variable::validate(std::string_view n) {
         throw InvalidNode(std::format("invalid blank node label {}", n));
     }
     ++it;
+
+    static constexpr auto matcher = ASCIINumMatcher{} | PNCharsUMatcher | PNChars_UniMatcher{};
```
bigerl: I first assumed it `-` through PNChars_UniMatcher (because PNChars would match it) but apparently it doesn't. So should be fine like this.

src/rdf4cpp/util/CharMatcher.hpp
```diff
@@ -298,6 +297,11 @@ struct PNCharsBase_UniMatcher {
  */
 constexpr auto PNCharsBaseMatcher = ASCIIAlphaMatcher{} | PNCharsBase_UniMatcher{};
 
+/**
+ * matches PN_CHARS_U of the Turtle/SPARQL specificiation
+ */
+constexpr auto PNCharsUMatcher = ASCIIPatternMatcher{"_"} | PNCharsBaseMatcher;
+
 /**
  * matches PN_CHARS of the Turtle/SPARQL specification, without ASCII num and PN_CHARS_BASE.
```
bigerl: - not only without PN_CHARS_BASE but even without PN_CHARS_U
- and without -.
- So for the docstring, something like: unicode ranges defined directly in PN_CHARS would work.

Also, I still find the name a bit confusing. How about: PNChars_UniPart_Matcher?

bigerl (changes requested):
Thx for the PR. Code looks good. Some minor comments and queries below.

src/rdf4cpp/BlankNode.cpp
liss-h: The bug was caused `Variable::make_anonymous` which previously did not match bnode names (it matched var names)

bigerl (approved):

PR #312, opened 2024-09-26
Fix: node comparisons

liss-h:
# Background
In SPARQL there are two types or orderings/comparisons: `FILTER`s and `ORDER BY` clauses.
The main difference is that for `ORDER BY` we sometimes need to order `Literals` that would usually not be comparable or are equal. (e.g. so that values of different types show up as single blocks in the solutions). Additionally, in `ORDER BY` IRIs and bnodes also have an ordering whereas the following `FILTER` would error: `<iri> <= <iri2>`.

This PR adds functions to use either behaviour. That was already possible before but required some wrapper code with some special cases.

For `FILTER` semantics use: `Node::{eq,lt,as_eq,...}`, `Literal::operator{==,<=>}`. Note due to the behaviour difference of `==`/`!=` and `<`/`<=`/`>`/`>=` in SPARQL we cannot provide a `compare` function for Node

For `ORDER BY` semantics use: `Node::order`, `Node::{order_eq,order_lt,as_order_eq,...}`, `Node::operator{<=>,==}`

As you have probably noticed, `operator<=>` for `Literal` and `Node` do different things.
`Literal`s follow `FILTER` semantics, whereas `Node`s follow `ORDER BY` semantics.
The reason for this is a compromise consisting of two reasons:
- `Node` needs to use `ORDER BY` semantics because you cannot use `std::set<Node>` properly otherwise (it would not work)
- `Literal` needs to use `FILTER` semantics because otherwise the following condition would not be true: `1_xsd_int == 1_xsd_integer`. This would confuse users. If you still want to put `Literal`s in a `std::set` use the `rdf4cpp::LiteralOrderByLess` comparator.

### Table of some the semantic differences when comparing op 1 and 2
This is not new, it was just not easily accessible previously.

| op 1            | op 2          | `FILTER`  | `ORDER BY` |
| --------------- | ------------- | --------- | ---------- |
| `1_xsd_integer` | `1_xsd_int`   | equal     | less       |
| `Literal{}`     | `1_xsd_int`   | error     | less       |
| `"iri"_iri`     | `1_xsd_int`   | error     | less       |
| `"iri1"_iri`    | `"iri2"_iri`  | not-equal | less       |
| `"bn1"_bnode`   | `"bn2"_bnode` | not-equal | less       |
| `null`          | `null`        | error     | equal      |
| `null`          | `"iri"_iri`   | error     | less       |

# Changes
- rename `*::compare_with_extensions` to `*::order` to better reflect its purpose
- Add `Node::{order,eq,order_eq,as_eq,...}` for feature parity with `Literal`

# TODO
- [x] Double check how comparisons with unbound/null nodes are supposed to behave

### DIFF
--- a/src/rdf4cpp/Node.cpp
+++ b/src/rdf4cpp/Node.cpp
@@ -120,9 +120,55 @@ bool Node::is_inlined() const noexcept {
     return handle_.is_inlined();
 }
 
-std::weak_ordering Node::operator<=>(const Node &other) const noexcept {
-    if (this->handle_ == other.handle_){
-        return std::strong_ordering::equivalent;
+template<typename T>
+[[nodiscard]] static std::partial_ordering eq_else_unordered(T const &a, T const &b) noexcept {
+    if (a == b) {
+        return std::partial_ordering::equivalent;
+    }
+
+    return std::partial_ordering::unordered;
+}
+
+std::partial_ordering Node::compare(Node const &other) const noexcept{
+    if (handle_ == other.handle_) {
+        return std::partial_ordering::equivalent;
+    }
+
+    if (handle_.type() != other.handle_.type()) {
+        return std::partial_ordering::unordered;
+    }
+
+    // unbound
+    if (null()) {
+        return std::weak_ordering::less;
+    } else if (other.null()) {
+        return std::weak_ordering::greater;
+    }
+
+    using storage::identifier::RDFNodeType;
+    switch (handle_.type()) {
+        case RDFNodeType::Literal: {
+            return Literal{handle_}.compare(Literal{other.handle_});
+        }
+        case RDFNodeType::IRI: {
+            return eq_else_unordered(handle_.iri_backend(), other.handle_.iri_backend());
+        }
+        case RDFNodeType::BNode: {
+            return eq_else_unordered(handle_.bnode_backend(), other.handle_.bnode_backend());
+        }
+        case RDFNodeType::Variable: {
+            return eq_else_unordered(handle_.variable_backend(), other.handle_.variable_backend());
+        }
+        default: {
+            assert(false); // unreachable
+            return std::partial_ordering::unordered;
+        }
+    }
+}
+
+std::weak_ordering Node::order(Node const &other) const noexcept {
+    if (this->handle_ == other.handle_) {
+        return std::weak_ordering::equivalent;
     }
 
     if (this->null() && other.null()) {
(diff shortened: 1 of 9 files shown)


src/rdf4cpp/Node.cpp
```diff
@@ -120,24 +120,82 @@ bool Node::is_inlined() const noexcept {
...
-        return std::strong_ordering::greater;
+        return std::weak_ordering::less;
+    }
+    if (other.null()) {
+        return std::weak_ordering::greater;
+    }
+
+    if (this->handle_ == other.handle_) {
+        return std::weak_ordering::equivalent;
     }
```
bigerl: I would reorder it a bit

```suggestion
    if (this->handle_ == other.handle_) {
        return std::weak_ordering::equivalent;
    }
    if (this->handle_.null() || other.handle_.null()) {
        return this->handle_.null()
            ? std::weak_ordering::less
            : std::weak_ordering::greater;
    }
```

tests/nodes/tests_Node.cpp
```diff
@@ -71,16 +71,16 @@ TEST_SUITE("comparisions") {
         IRI iri3 = IRI{"http://www.example.org/oest"};
         Literal lit1 = Literal::make_typed_from_value<String>("testlit1");
         Literal lit2 = Literal::make_typed_from_value<String>("testlit2");
-        CHECK(iri2 < iri1);
```
bigerl, line 74: rename 
```
TEST_SUITE("comparisions") {
```

to
```
TEST_SUITE("comparisons") {
```

tests/nodes/tests_Node.cpp
```diff
@@ -107,41 +107,41 @@ TEST_SUITE("comparisions") {
...
             // reason: decimal has fixed id and any fixed id type is always less than a dynamic one
-            CHECK(Literal::make_typed_from_value<Incomparable>(1).compare_with_extensions(Literal::make_typed_from_value<Decimal>(rdf4cpp::BigDecimal(10.0))) == std::weak_ordering::greater);
+            CHECK(Literal::make_typed_from_value<Incomparable>(1).order(Literal::make_typed_from_value<Decimal>(rdf4cpp::BigDecimal(10.0))) == std::weak_ordering::greater);
         }
 
         SUBCASE("nulls") {
-            CHECK(Literal{}.compare_with_extensions(Literal{}) == std::weak_ordering::equivalent);
+            CHECK(Literal{}.order(Literal{}) == std::partial_ordering::equivalent);
 
             // null <=> other (expecting null < any other a)
```
bigerl:
```suggestion
            // null <=> other (expecting null < any other literal)
```

bigerl (changes requested):
Thx for the PR. Changes look good. I only pointed out minor things, most not even originating from the PR.

src/rdf4cpp/Node.cpp
liss-h: Thats very subly wrong, it does not order null nodes of different types correctly

bigerl:
```c++
std::weak_ordering Node::order(Node const &other) const noexcept {
    // first, check for null nodes
    // "Apart from BOUND, COALESCE, NOT EXISTS and EXISTS, all functions and operators operate on RDF Terms and will produce a type error if any arguments are unbound."
    // - https://www.w3.org/TR/sparql11-query/#evaluation
    if (this->null() && other.null()) {
        return this->handle_.type() <=> other.handle_.type();
    }

    if (this->handle_ == other.handle_) {
        return std::weak_ordering::equivalent;
    }
    if (this->handle_.null() || other.handle_.null()) {
        return this->handle_.null()
            ? std::weak_ordering::less
            : std::weak_ordering::greater;
    }
```

bigerl (approved):

PR #314, opened 2024-10-02
Fix: boost issue by version bump

liss-h:
Boost versions below 1.86.0 had a bug where casting a `boost::multiprecision::cpp_int` to a double would sometimes throw an exception inside boost (in a noexcept context) which crashed the program.

Bumping the version resolves this.


nkaralis (approved):
Thank you for the PR.

PR #315, opened 2024-10-02
Fix: add workaround for older boost version

liss-h:
Follow up to #314 

It turns out that we cannot use more recent boost versions in our other library (because of an incompatibility in another dependency).

Therefore I added a workaround, so that it still works with older versions. I will keep the version bump in the conanfile, since that is the proper fix.

Please double check that I did not miss anything

### DIFF
--- a/src/rdf4cpp/BigDecimal.hpp
+++ b/src/rdf4cpp/BigDecimal.hpp
@@ -779,7 +779,7 @@ public:
      * @return
      */
     [[nodiscard]] explicit operator double() const noexcept {
-        double const v = cast_unscaled_to_double_safe(unscaled_value) * std::pow(base, -exponent);
+        double const v = cast_unscaled_to_double_safe(unscaled_value) * std::pow(static_cast<double>(base), -static_cast<double>(exponent));
         if (!std::isnan(v) && !std::isinf(v))
             return v;
         // even Javas BigDecimal has no better solution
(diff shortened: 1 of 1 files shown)


src/rdf4cpp/BigDecimal.hpp
```diff
@@ -435,6 +435,25 @@ struct BigDecimal {
...
     }
 
+    static double cast_unscaled_to_double_safe(UnscaledValue_t const &value) noexcept {
+        // There is a bug in boost versions <1.86.0
+        // that can cause a crash in the cast `static_cast<double>(cpp_int)`
+        // https://github.com/boostorg/multiprecision/issues/553
+
+#if BOOST_VERSION < 108600
+        // workaround from https://github.com/scylladb/scylladb/commit/51d09e6a2a407ea9b9ad380ba30e5558a25bb8be#diff-81e6758bbedbe566a51706c5af1ea68be225c36feb68983c3de0a69138f01264R73
+        static auto const min = UnscaledValue_t{std::numeric_limits<double>::lowest()};
```
nkaralis, line 445: the workaround in the linked fix is for `utils::multiprecision_int`. The default value for `UnscaledValue_t` is `cpp_int`. Could there be an issue if we use something else (do we actually use another type?)

nkaralis (commented):
Thanks for the PR. I have one question.

src/rdf4cpp/BigDecimal.hpp
liss-h: I didn't double check it but I was assuming that `utils::multiprecision_int` is `boost::multiprecision::cpp_int`. 

At least this workaround works the way I've written it (tried it with an older boost ver)

nkaralis (approved):

PR #319, opened 2024-10-17
Fix: clang 19 compile

liss-h:
As is always the case with clang, it was a template-template parameter with a concept in the inner template.. seems to get worse in every version

### DIFF
--- a/.github/workflows/run_test_and_examples.yml
+++ b/.github/workflows/run_test_and_examples.yml
@@ -9,9 +9,11 @@ jobs:
             compiler: gcc-13
 
           - os: ubuntu-22.04
-            compiler: clang-16
-          - os: ubuntu-22.04
             compiler: clang-17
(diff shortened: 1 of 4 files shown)


.github/workflows/run_test_and_examples.yml
```diff
@@ -8,10 +8,12 @@ jobs:
...
             compiler: gcc-13
 
-          - os: ubuntu-22.04
-            compiler: clang-16
           - os: ubuntu-22.04
             compiler: clang-17
+          - os: ubuntu-22.04
+            compiler: clang-18
+          - os: ubuntu-22.04
+            compiler: clang-19
```
bigerl, line 16: Please also change the build requirements in the readme.

bigerl (changes requested):
Thx for quick fix. Changes look good. Just the readme left.

PR #322, opened 2024-10-28
Single null node representation

### LINKED ISSUE #321: NULL nodes
Currently, the id of the NULL node depends on the node's type (Literal, BlankNode, IRI).
We need a representation of the NULL node, which is independent of the node type.

liss-h:
Close #321 

## Design
- null node's backend id is all zeros. This happens to be the blank node repr, but that doesn't matter (see following points)
- `Node::is_{literal,iri,..}` now return `TriBool`. If the node is null: `TriBool::Err` is returned, otherwise the function reports the type as before.
- `{Literal,IRI,BlankNode,Variable}::is_{literal,iri,..}` are now deleted, as it does not make sense to call them if you already know the concrete type of the node.

## POBR
This is **not** a POBR breaking change **under the assumption** that null-nodes are never persisted.

### DIFF
--- a/src/rdf4cpp/Node.cpp
+++ b/src/rdf4cpp/Node.cpp
@@ -103,17 +103,17 @@ Node::operator std::string() const noexcept {
     }
 }
 
-bool Node::is_literal() const noexcept {
-    return !handle_.null() && handle_.is_literal();
+TriBool Node::is_literal() const noexcept {
+    return handle_.is_literal();
 }
-bool Node::is_variable() const noexcept {
-    return !handle_.null() && handle_.is_variable();
+TriBool Node::is_variable() const noexcept {
+    return handle_.is_variable();
 }
-bool Node::is_blank_node() const noexcept {
-    return !handle_.null() && handle_.is_blank_node();
+TriBool Node::is_blank_node() const noexcept {
+    return handle_.is_blank_node();
 }
-bool Node::is_iri() const noexcept {
-    return !handle_.null() && handle_.is_iri();
+TriBool Node::is_iri() const noexcept {
+    return handle_.is_iri();
 }
 
 bool Node::is_inlined() const noexcept {
--- a/src/rdf4cpp/Node.cpp
+++ b/src/rdf4cpp/Node.cpp
@@ -104,16 +104,16 @@ Node::operator std::string() const noexcept {
 }
 
 bool Node::is_literal() const noexcept {
-    return handle_.is_literal();
+    return !handle_.null() && handle_.is_literal();
 }
 bool Node::is_variable() const noexcept {
-    return handle_.is_variable();
+    return !handle_.null() && handle_.is_variable();
 }
 bool Node::is_blank_node() const noexcept {
-    return handle_.is_blank_node();
+    return !handle_.null() && handle_.is_blank_node();
 }
 bool Node::is_iri() const noexcept {
-    return handle_.is_iri();
+    return !handle_.null() && handle_.is_iri();
 }
 
 bool Node::is_inlined() const noexcept {
--- a/src/rdf4cpp/Node.hpp
+++ b/src/rdf4cpp/Node.hpp
@@ -157,28 +157,28 @@ public:
     friend std::ostream &operator<<(std::ostream &os, const Node &node);
 
     /**
-     * Checks weather the node is a Literal. If yes, it is safe to convert it with `auto literal = (Literal) rdf_node;`
-     * @return if this is a Literal
+     * Checks whether the node is a Literal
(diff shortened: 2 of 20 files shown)


liss-h:
It might make sense to also modify the comparison functions. I'm pretty sure it will also work like this but it would be cleaner

src/rdf4cpp/Node.cpp
```diff
@@ -104,16 +104,16 @@ Node::operator std::string() const noexcept {
...
+    return !handle_.null() && handle_.is_variable();
 }
 bool Node::is_blank_node() const noexcept {
-    return handle_.is_blank_node();
+    return !handle_.null() && handle_.is_blank_node();
 }
 bool Node::is_iri() const noexcept {
-    return handle_.is_iri();
+    return !handle_.null() && handle_.is_iri();
 }
```
nkaralis: Should we also check in the dedicated functions (e.g., Literal::is_literal) if the node is null?

liss-h: I'm not sure, just because
```c++
Literal l{};
l.is_literal() // == false
```
feels kind of stupid

nkaralis: I am bit puzzled.
I am actually thinking whether the NULL node should be considered an instance of all types, since it can be joined with anything

bigerl: I like @nkaralis idea of making the null node instance of all types.  ➕ 1️⃣ 
Especially because of it making the `is_xyz` calls consistent across the type hierarchy.

bigerl (changes requested):

src/rdf4cpp/Node.cpp
nkaralis: we need to consider one thing (discussed with @liss-h): going this route contradicts SPARQL's semantics. In SPARQL, ISIRI(`null`) returns an error.

liss-h:
I have updated the design to address the concerns (description is also updated), please have another look @nkaralis

src/rdf4cpp/Node.hpp
```diff
@@ -157,28 +157,28 @@ struct Node {
     friend std::ostream &operator<<(std::ostream &os, const Node &node);
 
     /**
-     * Checks weather the node is a Literal. If yes, it is safe to convert it with `auto literal = (Literal) rdf_node;`
-     * @return if this is a Literal
+     * Checks whether the node is a Literal
+     * @return err if this is null, otherwise true iff this is a literal
```
mcb5637, line 161: typo: `if this is...`

mcb5637 (approved):
Looks good to me. Found only some typos.

src/rdf4cpp/Node.hpp
liss-h: `iff` means `if and only if`

src/rdf4cpp/storage/identifier/NodeBackendID.hpp
```diff
@@ -163,7 +163,7 @@ static_assert(alignof(NodeBackendID) == alignof(uint64_t));
  * @return the LiteralType associated with that IRI
  */
 constexpr LiteralType iri_node_id_to_literal_type(NodeBackendID const id) noexcept {
-    assert(id.is_iri());
+    assert(id.null() || id.is_iri());
```
nkaralis: I this function supposed to be called on the null node? If yes, what is it going to return?

nkaralis (commented):
Thank you for the changes. I only have one question

src/rdf4cpp/storage/identifier/NodeBackendID.hpp
liss-h: No, you are right. This was a mistake on my end

tests/query/tests_QuadPattern.cpp
```diff
@@ -190,3 +190,8 @@ TEST_CASE("QuadPattern - Check for iterators and const-iterators") {
         CHECK(*(--e_const_itr) == graph1);
     }
 }
+
+TEST_CASE("null-graph string repr") {
+    Quad const q{Node{}, IRI{"http://test.com#s"}, IRI{"http://test.com#p"}, IRI{"http://test.com#o"}};
```
nkaralis, line 195: Quad::valid would return false here, right?

liss-h: yes

liss-h: I think we need to fix this API in general, it does not make sense that we can even construct invalid quads, but thats something for another PR I think

nkaralis (approved):

PR #323, opened 2024-10-28
Feature: Variable and BlankNode inlining

liss-h:
Implements inlining for Variables and BlankNodes to avoid node storage access for variables/bnodes with short names (which is probably very common).

Variables are inlined if their name is at most 5 characters long.
BlankNodes are inlined if their name is at most 6 characters long.

This **is** a POBR breaking change.

### DIFF
--- a/src/rdf4cpp/BlankNode.cpp
+++ b/src/rdf4cpp/BlankNode.cpp
@@ -4,6 +4,8 @@
 #include <rdf4cpp/writer/TryWrite.hpp>
 #include <uni_algo/all.h>
 
+#include <cstring>
+
 namespace rdf4cpp {
 
 namespace detail_bnode_inlining {
--- a/src/rdf4cpp/BlankNode.cpp
+++ b/src/rdf4cpp/BlankNode.cpp
@@ -5,6 +5,49 @@
 #include <uni_algo/all.h>
 
 namespace rdf4cpp {
+
+namespace detail_bnode_inlining {
+    inline constexpr size_t max_inlined_name_len = storage::identifier::NodeID::width / (sizeof(char) * 8);
+
+    struct InlinedRepr {
+        char identifier[max_inlined_name_len];
+
+        [[nodiscard]] storage::view::BNodeBackendView view() const {
+            return storage::view::BNodeBackendView{.identifier = std::string_view{identifier, strnlen(identifier, max_inlined_name_len)}};
+        }
+
+        std::strong_ordering operator<=>(InlinedRepr const &other) const noexcept {
+            return view() <=> other.view();
+        }
+    };
+
+    [[nodiscard]] inline storage::identifier::NodeBackendID try_get_inlined(std::string_view const identifier) noexcept {
+        using namespace storage::identifier;
+
+        if (identifier.size() > max_inlined_name_len) {
+            return NodeBackendID{};
+        }
+
+        InlinedRepr inlined_data;
+
+        memcpy(&inlined_data.identifier, identifier.data(), identifier.size());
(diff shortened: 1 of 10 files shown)


src/rdf4cpp/BlankNode.cpp
```diff
@@ -4,7 +4,47 @@
...
+        [[nodiscard]] storage::view::BNodeBackendView view() const {
+            return storage::view::BNodeBackendView{.identifier = std::string_view{identifier, strnlen(identifier, max_inlined_name_len)}};
+        }
+
+        std::strong_ordering operator<=>(InlinedRepr const &other) const noexcept {
+            return view() <=> other.view();
+        }
+    };
+
+    [[nodiscard]] inline storage::identifier::NodeBackendID try_get_inlined(std::string_view const identifier) noexcept {
```
mcb5637: maybe rename to `try_into_inlined` to be consistent with literal datatypes

src/rdf4cpp/BlankNode.cpp
```diff
@@ -4,7 +4,47 @@
...
 
+#include <cstring>
+
 namespace rdf4cpp {
+
+namespace detail_bnode_inlining {
+    inline constexpr size_t max_inlined_name_len = storage::identifier::NodeID::width / (sizeof(char) * 8);
+
+    struct InlinedRepr {
+        char identifier[max_inlined_name_len];
```
mcb5637: maybe add ` = {}` to make clear to a reader that it gets default initialized

src/rdf4cpp/BlankNode.cpp
```diff
@@ -116,6 +189,42 @@ void BlankNode::validate(std::string_view v) {
...
 }
 
+std::strong_ordering BlankNode::order(BlankNode const &other) const noexcept {
+    if (is_inlined()) {
+        if (other.is_inlined()) {
+            return detail_bnode_inlining::from_inlined(handle_.id()) <=> detail_bnode_inlining::from_inlined(other.handle_.id());
+        }
+
+        // this is inlined but other is not, this means that the string of this must be smaller
+        // than the one of other
```
mcb5637: doesnt lexicographical order decide on the first non equal char? so `b` > `aa`

src/rdf4cpp/query/Variable.cpp
```diff
@@ -3,9 +3,53 @@
...
+#include <cstring>
+
 namespace rdf4cpp::query {
+
+namespace detail_variable_inlining {
+    inline constexpr size_t max_inlined_name_len = (storage::identifier::NodeID::width / (sizeof(char) * 8)) - 1; // -1 for anonymous tagging boolean
+
+    struct InlinedRepr {
+        char name[max_inlined_name_len];
+        bool is_anonymous;
```
mcb5637, line 18: could we store the bool in the highest bit of the last char? that should always be 0 for ASCII
(and if it is set, disable inlining for that variable)

mcb5637 (changes requested):
Overall looks good. Just some minor things and a question.
Note that every comment in BlankNode also applies to Variable.

src/rdf4cpp/BlankNode.cpp
liss-h: Ah you are right, good catch

src/rdf4cpp/query/Variable.cpp
liss-h: probably, but I don't think its worth the effort/code complexity just to get one more inlined char

mcb5637 (approved):

PR #324, opened 2024-11-06
IRIFactory prefix validity check

### LINKED ISSUE #320: PrefixeFactory: add prefix validation and eliminate allocation
https://github.com/rdf4cpp/rdf4cpp/blob/927db2d400e6d86e2f9f784eef8f336697648c11/src/rdf4cpp/IRIFactory.cpp#L230

This should validate the prefix and the expand. The purpose is that if someone iterates the prefixes of a factory and uses it to build a new turtle file or SPARQL query, it must not be possible that other sparql get's injected.

Also, the string_view should be used for lookup with find. Only, if it was not found it should be inserted. This can probably also happen with emplace without constructing a string object before. (if not already small string optimized, please double check).

mcb5637:
close #320 

add comment specifying that the parser ensures this

### DIFF
--- a/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -209,7 +209,7 @@ SerdStatus IStreamQuadIterator::Impl::on_prefix(void *voided_self, SerdNode cons
                                         .col = serd_reader_get_current_col(self->reader),
                                         .message = "Encountered prefix while parsing. hint: prefixes are not allowed in the current document. note: position may not be accurate and instead point to the end of the line."};
     } else {
-        self->state->iri_factory.assign_prefix(node_into_string_view(name), node_into_string_view(uri));
+        self->state->iri_factory.assign_prefix_unchecked(node_into_string_view(name), node_into_string_view(uri));
     }
 
     return SERD_SUCCESS;
--- a/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -209,7 +209,12 @@ SerdStatus IStreamQuadIterator::Impl::on_prefix(void *voided_self, SerdNode cons
                                         .col = serd_reader_get_current_col(self->reader),
                                         .message = "Encountered prefix while parsing. hint: prefixes are not allowed in the current document. note: position may not be accurate and instead point to the end of the line."};
     } else {
-        self->state->iri_factory.assign_prefix_unchecked(node_into_string_view(name), node_into_string_view(uri));
+        if (self->state->iri_factory.assign_prefix(node_into_string_view(name), node_into_string_view(uri)) != IRIFactoryError::Ok) {
+            self->last_error = ParsingError{.error_type = ParsingError::Type::BadSyntax,
+                                            .line = serd_reader_get_current_line(self->reader),
+                                            .col = serd_reader_get_current_col(self->reader),
+                                            .message = std::format("Invalid prefix: {}. note: position may not be accurate and instead point to the end of the line.", node_into_string_view(name))};
+        }
     }
 
     return SERD_SUCCESS;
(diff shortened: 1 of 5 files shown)


liss-h (commented):
A suggestion:
Since we already have validation infra in place here, it could maybe make sense to provide two functions

`assign_prefix` (to be used by users)
`assign_prefix_unchecked` (we will use this one)

src/rdf4cpp/IRIFactory.hpp
```diff
@@ -70,6 +70,15 @@ struct IRIFactory {
      * @param prefix
      * @param expanded
      */
+    [[nodiscard]] IRIFactoryError assign_prefix_checked(std::string_view prefix, std::string_view expanded);
```
liss-h: Naming should be the other way around `assign_prefix` and `assign_prefix_unchecked`, this is the safer approach

src/rdf4cpp/IRIFactory.hpp
```diff
@@ -70,6 +70,15 @@ struct IRIFactory {
...
+
+    /**
+     * Creates or changes a prefix.
+     * expects prefix and expanded to be valid (i.e. not contain spaces or other chars that are invalid in an IRI).
+     * this is ensured by our parser: https://github.com/dice-group/sparql-parser-base
```
liss-h: point to the other version of the function instead of the sparql parser

src/rdf4cpp/IRIFactory.cpp
```diff
@@ -226,6 +229,14 @@ nonstd::expected<IRI, IRIFactoryError> IRIFactory::create_and_validate(std::stri
     return IRI::make_unchecked(iri, node_storage);
 }
 
+IRIFactoryError IRIFactory::assign_prefix_checked(std::string_view prefix, std::string_view expanded) {
+    using namespace util::char_matcher_detail;
+    static constexpr auto pattern = PNCharsMatcher | ASCIIPatternMatcher{"."};
```
liss-h: The matcher should follow this: 
grafik

(https://www.w3.org/TR/sparql11-query/#rPN_PREFIX)

src/rdf4cpp/IRIFactory.cpp
```diff
@@ -226,6 +229,14 @@ nonstd::expected<IRI, IRIFactoryError> IRIFactory::create_and_validate(std::stri
     return IRI::make_unchecked(iri, node_storage);
 }
 
+IRIFactoryError IRIFactory::assign_prefix_checked(std::string_view prefix, std::string_view expanded) {
+    using namespace util::char_matcher_detail;
```
liss-h, line 233: please add a comment inside the function that we do not need to check the `expanded` because it will be done on IRI creation

liss-h (changes requested):

src/rdf4cpp/IRIFactory.cpp
mcb5637: The other version was faster. but i guess it does not really matter in this case.

private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -209,7 +209,7 @@ SerdStatus IStreamQuadIterator::Impl::on_prefix(void *voided_self, SerdNode cons
                                         .col = serd_reader_get_current_col(self->reader),
                                         .message = "Encountered prefix while parsing. hint: prefixes are not allowed in the current document. note: position may not be accurate and instead point to the end of the line."};
     } else {
-        self->state->iri_factory.assign_prefix(node_into_string_view(name), node_into_string_view(uri));
+        self->state->iri_factory.assign_prefix_unchecked(node_into_string_view(name), node_into_string_view(uri));
```
liss-h: I don't think we should trust serd here, its quite lax in parsing

src/rdf4cpp/IRIFactory.cpp
```diff
@@ -226,7 +229,35 @@ nonstd::expected<IRI, IRIFactoryError> IRIFactory::create_and_validate(std::stri
     return IRI::make_unchecked(iri, node_storage);
 }
 
-void IRIFactory::assign_prefix(std::string_view prefix, std::string_view expanded) {
+IRIFactoryError IRIFactory::assign_prefix(std::string_view prefix, std::string_view expanded) {
+    using namespace util::char_matcher_detail;
+    static constexpr auto first_matcher = ASCIINumMatcher{} | PNCharsUMatcher;
```
liss-h: Shouldn't this be `PNCharsBaseMatcher`? In the grammar the first char is of `PN_CHARS_BASE`

liss-h (changes requested):

liss-h (approved):

PR #325, opened 2024-11-11
fix license copy

mcb5637:
Note: easiest way to validate, if the licenses get properly copied, is to run `conan create`, get the cache path from the output and check it manually


liss-h (changes requested):
~~when I build the package the serd license is missing. Just have the rdf4cpp license~~ nvm

liss-h (approved):

PR #326, opened 2024-11-13
Feature: more fetch-or-serialize functions

liss-h:
Add `fetch_or_serialize_*` function to `Variable`, `BlankNode` and `IRI`

### DIFF
--- a/src/rdf4cpp/BlankNode.cpp
+++ b/src/rdf4cpp/BlankNode.cpp
@@ -132,6 +132,20 @@ CowString BlankNode::identifier() const noexcept {
     return CowString{CowString::borrowed, handle_.bnode_backend().identifier};
 }
 
+FetchOrSerializeResult BlankNode::fetch_or_serialize_identifier(std::string_view &out, writer::BufWriterParts writer) const noexcept {
+    if (handle_.is_inlined()) {
+        auto const inlined = detail_bnode_inlining::from_inlined(handle_.id());
+        if (!rdf4cpp::writer::write_str(inlined.view().identifier, writer)) [[unlikely]] {
+            return FetchOrSerializeResult::SerializationFailed;
+        }
+
+        return FetchOrSerializeResult::Serialized;
+    }
+
+    out = handle_.bnode_backend().identifier;
+    return FetchOrSerializeResult::Fetched;
+}
+
 bool BlankNode::serialize(writer::BufWriterParts const writer) const noexcept {
     if (null()) {
         return rdf4cpp::writer::write_str("null", writer);
--- a/src/rdf4cpp/query/Variable.cpp
+++ b/src/rdf4cpp/query/Variable.cpp
@@ -150,6 +150,20 @@ CowString Variable::name() const {
     return CowString{CowString::borrowed, handle_.variable_backend().name};
 }
 
+FetchOrSerializeResult Variable::fetch_or_serialize_name(std::string_view &out, writer::BufWriterParts writer) const noexcept {
+    if (handle_.is_inlined()) {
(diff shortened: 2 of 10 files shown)


tests/nodes/tests_Node.cpp
```diff
@@ -222,6 +222,31 @@ TEST_CASE("IRI UUID") {
...
 
+TEST_CASE("IRI fetch or serialize") {
+    auto const x = IRI::make("https://example.com#some-iri");
+    auto const expected = x.identifier();
+
+    std::string buf;
+    writer::StringWriter w{buf};
+
+    std::string_view fetched;
+    switch (x.fetch_or_serialize_identifier(fetched, w)) {
```
bigerl: No need for a switch-cases here. 

```
CHECK_EQ(x.fetch_or_serialize_identifier(fetched, w), FetchOrSerializeResult::Fetched)
```

will do because IRI cannot inline. So, it will not serialize.

src/rdf4cpp/BlankNode.cpp
```diff
@@ -132,6 +132,20 @@ CowString BlankNode::identifier() const noexcept {
     return CowString{CowString::borrowed, handle_.bnode_backend().identifier};
 }
 
+FetchOrSerializeResult BlankNode::fetch_or_serialize_identifier(std::string_view &out, writer::BufWriterParts writer) const noexcept {
+    if (handle_.is_inlined()) {
+        auto const inlined = detail_bnode_inlining::from_inlined(handle_.id());
+        if (!rdf4cpp::writer::write_str(inlined.view().identifier, writer)) [[unlikely]] {
```
bigerl, line 138: How can this happen? Can You add a test case for this path?

src/rdf4cpp/query/Variable.cpp
```diff
@@ -150,6 +150,20 @@ CowString Variable::name() const {
     return CowString{CowString::borrowed, handle_.variable_backend().name};
 }
 
+FetchOrSerializeResult Variable::fetch_or_serialize_name(std::string_view &out, writer::BufWriterParts writer) const noexcept {
+    if (handle_.is_inlined()) {
+        auto const inlined_repr = detail_variable_inlining::from_inlined(handle_.id());
+        if (!rdf4cpp::writer::write_str(inlined_repr.view().name, writer)) {
+            return FetchOrSerializeResult::SerializationFailed;
```
bigerl, line 157: Same here. How is this possible?

tests/nodes/tests_Node.cpp
```diff
@@ -360,11 +385,36 @@ TEST_CASE("variable inlining") {
         CHECK_EQ(var.backend_handle().id(), again3.backend_handle().id());
     };
 
+    auto const check_fetch_or_serialize = [&](query::Variable const &var) {
```
bigerl, line 376: `check_fetch_or_serialize` should have another parameter `FetchOrSerializeResult expected_result` to validate the outcome. Also applies to `check_fetch_or_serialize` for `BlankNode`s.

bigerl (changes requested):
Thx for the PR. I think the previous PR regarding `fetch_or_serialize` was not reviewed by me. Some of the comments to this one might also apply to the similar code for IRI.

src/rdf4cpp/BlankNode.cpp
liss-h: Can only happen if the writer has a finite amount of space or is unable to make additional space, so this has nothing to do with this PR

bigerl: would still be important to test it with every `fetch_or_serialize` function because not handling this case in any of them would make for very rare but hard to track down bugs.

bigerl (approved):
Looks good now. Thx

PR #327, opened 2024-11-14
Fix: UBSan warning in datetime tests

liss-h:
Fix
grafik

That was caused by additions that overflowed int64_t.
Please double check that this is not an issue.

Additionally, I've aligned the interface of some types more towards `std::chrono` to hopefully break less code (also of ours). This primarily consists of making fields private and providing accessors instead (like in chrono).

### DIFF
--- a/tests/datatype/tests_time_types.cpp
+++ b/tests/datatype/tests_time_types.cpp
@@ -90,8 +90,8 @@ TEST_CASE("year api") {
     CHECK(--y(5) == y(4));
     CHECK(y(5)-- == y(5));
 
-    CHECK(y::max().year == std::numeric_limits<int64_t>::max());
-    CHECK(y::min().year == std::numeric_limits<int64_t>::min());
+    CHECK(static_cast<int64_t>(y::max()) == std::numeric_limits<int64_t>::max());
+    CHECK(static_cast<int64_t>(y::min()) == std::numeric_limits<int64_t>::min());
 }
 
 TEST_CASE("YearMonth api") {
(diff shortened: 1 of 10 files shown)


tests/datatype/tests_time_types.cpp
```diff
@@ -694,22 +697,22 @@ TEST_CASE("arithmetic") {
...
     CHECK((Literal::make_typed<datatypes::xsd::Date>("2042-5-6") + Literal::make_typed<datatypes::xsd::YearMonthDuration>("P1Y")) == Literal::make_typed<datatypes::xsd::Date>("2043-5-6"));
     CHECK((Literal::make_typed<datatypes::xsd::Time>("1:2:3") + Literal::make_typed<datatypes::xsd::DayTimeDuration>("P1DT1S")) == Literal::make_typed<datatypes::xsd::Time>("1:2:4"));
-    CHECK(!(Literal::make_typed<datatypes::xsd::DateTime>("2042-5-6T1:2:3") + Literal::make_typed_from_value<datatypes::xsd::YearMonthDuration>(std::chrono::months{std::numeric_limits<int64_t>::max()})).null());
-    CHECK(!(Literal::make_typed<datatypes::xsd::DateTime>("2042-5-6T1:2:3") + Literal::make_typed_from_value<datatypes::xsd::DayTimeDuration>(std::chrono::milliseconds{std::numeric_limits<int64_t>::max()})).null());
+    CHECK((Literal::make_typed<datatypes::xsd::DateTime>("2042-5-6T1:2:3") + Literal::make_typed_from_value<datatypes::xsd::YearMonthDuration>(std::chrono::months::max())).null()); // overflow
+    CHECK(!(Literal::make_typed<datatypes::xsd::DateTime>("2042-5-6T1:2:3") + Literal::make_typed_from_value<datatypes::xsd::DayTimeDuration>(std::chrono::nanoseconds::max())).null());
```
liss-h, line 704: Note: this was also overflowing, but that is ok because the overflow happens before the `Literal` is constructed (conversion between chrono types overflows)

mcb5637 (approved):
I modified the timepoint+duration calculation a bit, so the previously overflowing calculation now works fine.
Otherwise looks good.

PR #329, opened 2024-11-20
fix mac compile

mcb5637:


src/rdf4cpp/storage/reference_node_storage/detail/BiDirFlatMap.hpp
```diff
@@ -97,7 +97,11 @@ struct BiDirFlatMap {
      * Translates the given id into an index into forward_
      */
     [[nodiscard]] static constexpr size_type to_index(id_type const id) noexcept {
-        return static_cast<size_type>(id) - 1;
+        if constexpr (std::same_as<uint64_t, size_t>) {
```
liss-h: I think it would probably be better if we just used `uint64_t` instead of `size_t` for the bitmap type

conanfile.py
```diff
@@ -35,7 +35,7 @@ def requirements(self):
         self.requires("openssl/3.3.2")
         self.requires("uni-algo/1.2.0")
         self.requires("highway/1.2.0")
-        self.requires("dice-hash/0.4.8", transitive_headers=True)
+        self.requires("dice-hash/0.4.9@dice-group/fix-mac_compile", transitive_headers=True)
```
liss-h: Just so we don't forget: point to release (when I've created it)

src/rdf4cpp/util/CharMatcher.cpp
```diff
@@ -57,50 +57,67 @@ namespace rdf4cpp::util::char_matcher_detail::HWY_NAMESPACE {
...
+#define GET_SINGLE(x) x
+#define LOAD_SINGLE(x) Set(d, x)
+#else
+        // highway doc: on arm neon <, >, <= and >= are 1 instruction for any int
+        // Set is 1 instruction, while potential load/store is 1
+
+        using single_storage = int8_t;
+#define GET_SINGLE(x) Set(d, x)
+#define LOAD_SINGLE(x) x
+#endif
```
liss-h, line 83: This is a little confusing to me, could you explain this?

liss-h (changes requested):
Thx for the PR, I have some questions and suggestions

src/rdf4cpp/util/CharMatcher.cpp
mcb5637: TLDR: its an optimization i did on x86 that does not work on arm

liss-h: Ok, I reread the comments, this time they made sense to me

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.22)
 project(rdf4cpp VERSION 0.0.49)
-set(POBR_VERSION 2)  # Persisted Object Binary Representation
+set(POBR_VERSION 3)  # Persisted Object Binary Representation
```
liss-h: Don't need a POBR version bump, the change in BiDirFlatMap does nothing on relevant architectures (which is only x86-64 atm)

liss-h: I'll revert it

PR #332, opened 2024-12-10
Fix segfault on binop with null node

### LINKED ISSUE #331: Literal subtraction crashes when one of the operands is the null node
**Describe the bug**
Literal subtraction crashes when one of the operands is the null node. This might be the case for the other numerical operations. 

**To Reproduce**
On develop:
```c++
{
            auto null_node = rdf4cpp::Node::make_null();
            auto float_lit = Literal::make_typed_from_value<datatypes::xsd::Float>(2.f);
            auto result = float_lit - null_node.as_literal(); // crashes
}
```

**Expected behavior**
I think the result should be the null node.

liss-h:
Close #331


nkaralis (approved):
Thank you for the quick fix.

PR #334, opened 2024-12-16
add werror to CI

mcb5637:


.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
```diff
@@ -108,7 +108,7 @@ jobs:
           CXX: ${{ steps.install_cc.outputs.cxx }}
           CXXFLAGS: -Wno-unused-command-line-argument -fuse-ld=mold ${{ fromJSON('["", "-fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer -fno-optimize-sibling-calls"]')[inputs.with-sanitizer] }} ${{ fromJSON('["", "--coverage"]')[inputs.with-coverage] }} ${{ fromJSON('["", "-Werror"]')[inputs.with-Werror] }} ${{ fromJSON('["", "-march=native"]')[inputs.with-AVX] }}
```
liss-h: There is still the `inputs.with-Werror` option in the scripts (is used here). I think we can remove it

liss-h (changes requested):

liss-h (approved):

PR #339, opened 2025-02-14
Fix: improve performance of binary ops

liss-h:
Improve performance of binary operations by avoiding expensive checks (involving casting) through a tagging bit for chrono types.

Additionally: includes all changes from #340:

Generalizes the chrono types into:
- `Duration` (`xsd:duration`, `xsd:yearMonthDuration`, `xsd:dayTimeDuration`)
    - operations:
        - `Duration + Duration` ⇒ (same as for numerics) cast to common type, do op
        - `Duration - Duration` ⇒ cast to common type, do op
        - `Duration / Duration` ⇒ cast to common type, do op, result: `xsd:decimal`
        - `Duration * Numeric` ⇒ `$lhs * xsd:double($rhs) -> type($lhs)`
        - `Duration / Numeric` ⇒ `$lhs / xsd:double($rhs) -> type($lhs)`
- `TimePoint` (`xsd:date`, `xsd:time`, `xsd:dateTime`, `xsd:dateTimeStamp`)
    - operations:
        - `TimePoint - TimePoint` ⇒ cast to common type, do op
        - `TimePoint + Duration` ⇒ `$lhs + xsd:duration($rhs) -> type($lhs)`
        - `TimePoint - Duration` ⇒ `$lhs - xsd:duration($rhs) -> type($lhs)`

Additionally this fixes slight inaccuracies in the type hierarchy, as many of the types that were previously in a sub-supertype relation should have been in a promotion relationship.

PR #340, opened 2025-02-18
Fix: generalize chrono binary operations

liss-h:
An advancement of #339 that takes a different approach. It generalizes the chrono types into:

- `Duration` (`xsd:duration`, `xsd:yearMonthDuration`, `xsd:dayTimeDuration`)
    - operations:
        - `Duration + Duration` ⇒ (same as for numerics) cast to common type, do op
        - `Duration - Duration` ⇒ cast to common type, do op
        - `Duration / Duration` ⇒ cast to common type, do op, result: `xsd:decimal`
        - `Duration * Numeric` ⇒ `$lhs * xsd:double($rhs) -> type($lhs)`
        - `Duration / Numeric` ⇒ `$lhs / xsd:double($rhs) -> type($lhs)`
- `TimePoint` (`xsd:date`, `xsd:time`, `xsd:dateTime`, `xsd:dateTimeStamp`)
    - operations:
        - `TimePoint - TimePoint` ⇒ cast to common type, do op
        - `TimePoint + Duration` ⇒ `$lhs + xsd:duration($rhs) -> type($lhs)`
        - `TimePoint - Duration` ⇒ `$lhs - xsd:duration($rhs) -> type($lhs)`

Additionally this fixes slight inaccuracies in the type hierarchy, as many of the types that were previously in a sub-supertype relation should have been in a promotion relationship.


src/rdf4cpp/datatypes/xsd/time/Duration.cpp
```diff
@@ -210,9 +210,46 @@ std::partial_ordering capabilities::Comparable<xsd_duration>::compare(cpp_type c
     }
     return o;
 }
+
+template<>
+nonstd::expected<capabilities::Duration<xsd_duration>::cpp_type, DynamicError>
+capabilities::Duration<xsd_duration>::duration_add(cpp_type const &lhs, cpp_type const &rhs) noexcept {
+    // TODO
```
liss-h: For `xsd:duration` the `Duration` implementation is still missing because I didn't quite know how that type was supposed to be used.

@mcb5637 could you add them?

tests/datatype/tests_time_types.cpp
```diff
@@ -722,6 +719,8 @@ TEST_CASE("arithmetic") {
     CHECK((Literal::make_typed<datatypes::xsd::DayTimeDuration>("P1DT1M") - Literal::make_typed<datatypes::xsd::DayTimeDuration>("P1DT1S")) == Literal::make_typed<datatypes::xsd::DayTimeDuration>("PT59S"));
     CHECK((Literal::make_typed<datatypes::xsd::YearMonthDuration>("P1Y") - Literal::make_typed<datatypes::xsd::YearMonthDuration>("P2Y")) == Literal::make_typed<datatypes::xsd::YearMonthDuration>("-P1Y"));
 
+    // TODO add tests for xsd:duration + xsd:duration & xsd:duration + xsd:dayTimeDuration & xsd:duration + xsd:yearMonthDuration & so on
```
liss-h: @mcb5637 could you add tests here?

liss-h:
CI is failing because of `-Werror`, this can be ignored for now

src/rdf4cpp/datatypes/registry/DatatypeConversion.hpp
```diff
@@ -17,6 +17,14 @@ namespace rdf4cpp::datatypes::registry {
 
 namespace conversion_detail {
 
+// require Property<A> <-> Property<B>
+#define RDF4CPP_DETAIL_BIDIRECTIONAL_PRESERVE_PROPERTY_RULE(Property, A, B) \
```
mcb5637, line 21: could you undef these, so they dont clutter the namespace for whoever includes the file?

src/rdf4cpp/datatypes/xsd/time/Date.cpp
```diff
@@ -69,25 +69,63 @@ rdf4cpp::TimePoint date_to_tp(YearMonthDay const &d) noexcept {
...
 
 template<>
 std::partial_ordering capabilities::Comparable<xsd_date>::compare(cpp_type const &lhs, cpp_type const &rhs) noexcept {
     return rdf4cpp::datatypes::registry::util::compare_time_points(date_to_tp(lhs.first), lhs.second, date_to_tp(rhs.first), rhs.second);
 }
+
+template<>
+nonstd::expected<capabilities::Timepoint<xsd_date>::timepoint_sub_result_cpp_type, DynamicError>
+capabilities::Timepoint<xsd_date>::timepoint_sub(cpp_type const &lhs, cpp_type const &rhs) noexcept {
+    auto const super_lhs = Promotable<xsd_date>::promote(lhs);
```
mcb5637, line 90: this is a lot of duplicated code, maybe call `capabilities::Timepoint<xsd_dateTime>::timepoint_sub` ?
(also, the try-catch is missing)

src/rdf4cpp/datatypes/xsd/time/Duration.cpp
mcb5637: I dont think its supposed to be used this way at all.
example: `{100,50}-{50,100}=={50,-50}` => thats an invalid duration.
duration/duration is even more complicated, i dont even know how to implement that.

so, implement what we can and return unsupported for everything else (or if the result is invalid)?

mcb5637 (changes requested):
Overall, looks good, some minor things and duration needs thinking about.

src/rdf4cpp/datatypes/xsd/time/Duration.cpp
liss-h: Could this operation instead return `{49, nanos_in_a_month - 50}`? So incrementing/decrementing the months when over/underflow occurs in the nanos?

src/rdf4cpp/datatypes/xsd/time/Date.cpp
liss-h: I created a utility function that all of them call now

src/rdf4cpp/datatypes/xsd/time/Duration.cpp
mcb5637: No, because months are not directly convertible to days (Feb is 28/29 days, Jan is 31). So depending on what you do in future with that duration, you need a different result to make it consistent.

liss-h: thats unfortunate. Ok I guess I'll return `Unsupported`

liss-h:
Thanks for implementing the missing functionality @mcb5637

mcb5637 (approved):
looks good.
please check the pobr workflow, otherwise this is good to go.

liss-h:
> please check the pobr workflow, otherwise this is good to go.

Since the base branch is not develop, this does not matter. In the base (in which I will merge this, the workflow will pass because I bumped the version there)

PR #344, opened 2025-03-04
Use `static_string` from dice-template-library instead of own type

liss-h:
Replace `rdf4cpp::storage::refererence_node_storage::detail::ConstString` with `dice::template_library::static_string`.


liss-h:
Not POBR breaking change because the two types have the same layout

bigerl (approved):
Thx for the PR. Integration looks good 👍🏻

PR #345, opened 2025-03-06
Feature: ostream operators for storage ids

liss-h:
Add ostream operator overloads for storage ids.
This makes debugging easier.


mcb5637 (changes requested):
Thanks for the PR. The code looks good, just missing some test cases for them.

liss-h:
While writing the tests I noticed that the ID representation was very long in some cases, so I've removed the name in front of the IDs (since that should always be easy to determine from context)

mcb5637 (approved):

PR #348, opened 2025-03-13
More features for BiDirFlatMap

liss-h:
Backport features (and fixed) from downstream.

### DIFF
--- a/src/rdf4cpp/storage/reference_node_storage/detail/BiDirFlatMap.hpp
+++ b/src/rdf4cpp/storage/reference_node_storage/detail/BiDirFlatMap.hpp
@@ -16,7 +16,12 @@ namespace rdf4cpp::storage::reference_node_storage::detail {
  * @tparam Equal equality for View and Value
  * @tparam Allocator allocator
  */
-template<typename Id, typename Value, typename View, typename Hash = std::hash<View>, typename Equal = std::equal_to<>, typename Allocator = std::allocator<Value>>
+template<typename Id, typename Value, typename View,
+         typename Hash = std::hash<View>,
+         typename Equal = std::equal_to<>,
+         typename Allocator = std::allocator<Value>,
+         template<typename, typename> typename ForwardVector = std::vector,
+         template<typename, typename, typename, typename> typename BackwardUnorderedSet = ::dice::sparse_map::sparse_set>
 struct BiDirFlatMap {
     using id_type = Id;
     using mapped_type = Value;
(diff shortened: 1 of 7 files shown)


bigerl (approved):
Thx for the PR. Looks good. Most of the changes I already reviewed elsewhere. Also the additional changes make sense. The consequent use of forward/backward and calling value Backend makes the documentation much easier to parse. 

I think we do not really have a POBR change here since the layout stays the same.

src/rdf4cpp/storage/reference_node_storage/detail/BiDirFlatMap.hpp
```diff
@@ -174,13 +259,29 @@ struct BiDirFlatMap {
         return it->id;
     }
 
+    /**
+     * Look up the id (or ids, in case backwards is a multiset) corresponding the given view
+     *
+     * @param view view of value of which to find the id
+     * @return range of ids of the value if there were any, otherwise empty range if no id was found
+     */
+    [[nodiscard]] std::ranges::range auto lookup_id_range(view_type const &view) const noexcept {
```
bigerl, line 268: Ah, just one more thing: we should use
```
template <typename Container>
concept MultiMapLike = requires(Container c, typename Container::value_type val) {
    { c.insert(val) } -> std::same_as<typename Container::iterator>;
};

template <typename Container>
constexpr bool is_multimap_v = MultiMapLike<Container>;
```
 to enable only the functions for the respective underlying map type.

bigerl (changes requested):

src/rdf4cpp/storage/reference_node_storage/detail/BiDirFlatMap.hpp
liss-h: I don't think thats necessary. Multi and non-multi maps support both kinds of functions

bigerl: This was the wrong function to comment to. 

My main concern is that `lookup_id` is called when the `backward_type` is a multi-map. This would be most probably by mistake and should not be allowed. Calling `lookup_id_range` on a non-multi-map is ok but doesn't make much sense.

liss-h: I somewhat agree with your concern, but I don't think this is a good idea, because:

- `multimap` itself also has `find`, so there must be some usecase where you may want to use that.

- trying to detect container behaviour based on an interface is inherently flawed and does not work in C++. Just because an object has a specific interface does not mean that it behaves a certain way. Constraints do not have the purpose to know what an object does, just that it can do it (mainly to improve error messages). In contrast, rust traits determine what an object does, not just that they can do it. So we would be disabling functionality based on a guess what the behaviour might be.

bigerl: ok, good points. Then let's keep things as they are.

bigerl (approved):

PR #354, opened 2025-03-24
Fix: leading zeros parsing

### LINKED ISSUE #353: valid integer value rejected
rdf4cpp rejects `"09"^^<http://www.w3.org/2001/XMLSchema#integer>` with 

```
An error occurred during query evaluation: Parse Error: Malformed Query: Illegal literal `"09"^^<http://www.w3.org/2001/XMLSchema#integer>`: http://www.w3.org/2001/XMLSchema#integer parsing error: Unexpected content found while parsing character string.. At line: 72 and position: 117
```

I've reviewed the [integer](https://www.w3.org/TR/xmlschema-2/#integer) specs. It allows 0-9. Leading zeros are only prohibited in the canonical representation. 

Please also add a regression test and double check that other datatypes are not affected.

liss-h:
Allow parsing of xsd:Integer with leading zeros

Close #353

### DIFF
--- a/tests/nodes/tests_Literal.cpp
+++ b/tests/nodes/tests_Literal.cpp
@@ -1163,26 +1163,17 @@ TEST_CASE("Literal::fetch_or_serialize_lexical_form") {
     }
 }
 
-template<datatypes::LiteralDatatype L>
-void check_leading_zeros_parse() {
-    SUBCASE(L::identifier.c_str()) {
-        auto lit = Literal::make_typed<L>("09");
-        CHECK_FALSE(lit.null());
-        CHECK_EQ(lit.template value<L>(), 9);
+TEST_CASE_TEMPLATE("leading zeros parsing", L, datatypes::xsd::Integer,
+                                               datatypes::xsd::Int,
+                                               datatypes::xsd::Float,
+                                               datatypes::xsd::Double,
+                                               datatypes::xsd::Decimal) {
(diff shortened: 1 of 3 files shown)


tests/nodes/tests_Literal.cpp
```diff
@@ -1163,3 +1163,26 @@ TEST_CASE("Literal::fetch_or_serialize_lexical_form") {
...
+        CHECK_FALSE(lit.null());
+        CHECK_EQ(lit.template value<L>(), 9);
+
+        lit = Literal::make_typed<L>("000");
+        CHECK_FALSE(lit.null());
+        CHECK_EQ(lit.template value<L>(), 0);
+    }
+}
+
+TEST_CASE("leading zeros parsing") {
```
mcb5637: this might be a perfect case for a `TEST_CASE_TEMPLATE` (example see: https://github.com/rdf4cpp/rdf4cpp/blob/086033bdbf68e7fd49a966628796c13c04e755e9/tests/nodes/tests_Node.cpp#L304 )

mcb5637 (changes requested):
Fix looks good, just one minor code quality issue

tests/nodes/tests_Literal.cpp
liss-h: Ah, right totally forgot that exists

mcb5637 (approved):

PR #356, opened 2025-03-31
Fix: SEGFAULT caused by `va_list` usage in error message generation

liss-h:
Fixes a SEGFAULT caused by invalid `va_list` usage.

### Bug Description
Serd's API is written in terms of `vprintf` compatibility (i.e. you need to generate the message via `v*printf*` and `char const *fmt, va_list args`)

Since we want to put the message in a buffer instead of printing it directly, we are using `vsnprintf`. This involves two calls to `vsnprintf`, one to figure out how big the buffer should be and another one to actually fill the buffer with the message. As it turns out the first call to `vsnprintf` consumed the `va_list`, therefore the second call to `vsnprintf` would overread into garbage. Sometimes this overread would just produce garbage messages, sometimes it would trigger a SEGFAULT.

### Fix
We just need to `va_copy` the list for the first call. Additionally, I added some more stuff to make the code more robust.

### DIFF
--- a/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -163,20 +163,56 @@ nonstd::expected<Literal, SerdStatus> IStreamQuadIterator::Impl::get_literal(Ser
     }
 }
 
+/**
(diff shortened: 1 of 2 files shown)


bigerl (approved):
Thx for the PR. Changes look good. This was a sneaky one!

PR #357, opened 2025-04-07
Fix: short form serialization of non-finite doubles

liss-h:
Fix serialization of non-finite `xsd:double` values, such as `INF`, `-INF`, `NaN` and `HUGE_VAL`.

Excerpt from turtle spec:
grafik


nkaralis (approved):
Thank you for the quick fix.

PR #358, opened 2025-04-16
Fix: mac compat of NodeBackendID cast

liss-h:
On mac `uintptr_t` and `uint64_t` are distinct types even though they have the same size. This causes issues in some of our projects

### DIFF
--- a/src/rdf4cpp/storage/identifier/NodeBackendID.hpp
+++ b/src/rdf4cpp/storage/identifier/NodeBackendID.hpp
@@ -31,7 +31,7 @@ private:
     };
 
 public:
-    NodeBackendID() noexcept = default;
+    constexpr NodeBackendID() noexcept = default;
 
     explicit constexpr NodeBackendID(underlying_type const underlying_id) noexcept : underlying_{underlying_id} {
     }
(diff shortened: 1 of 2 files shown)


src/rdf4cpp/storage/identifier/NodeBackendID.hpp
```diff
@@ -138,8 +138,9 @@ struct alignas(uint64_t) NodeBackendID {
         return underlying_;
     }
 
-    explicit constexpr operator underlying_type() const noexcept {
-        return underlying_;
+    template<std::unsigned_integral I> requires (sizeof(I) >= sizeof(underlying_type))
+    explicit constexpr operator I() const noexcept {
```
bigerl, line 142: I would add something like:
```
static_assert(std::is_same_v<decltype(static_cast<uintptr_t>(std::declval<NodeBackendID>())), uintptr_t>,
              "NodeBackendID should be convertible to uintptr_t");
static_assert(std::is_same_v<decltype(static_cast<uint64_t>(std::declval<NodeBackendID>())), uint64_t>,
              "NodeBackendID should be convertible to uint64_t");
```

after the struct. The idea is to guarantee that it is instantiatable without the test cases. 

Side note: default constructor should be constexpr as well so. Then we can get rid of the `std::declval` above.

bigerl (changes requested):
Thx for the PR. Looks good to me. I would just consider to move the testing into the hpp file as it can be done in a `static_assert`.

bigerl (approved):

PR #360, opened 2025-04-28
Feature: convenience functions for `rdf:type` and datatype IRIs

liss-h:
Adds: `IRI::rdf_type()`, `IRI::is_rdf_type()`, `IRI::datatype<>()` and `IRI::is_datatype<>()`.

Additionally, it fixes the parsing of xsd:yearMonthDuration which apparently has been buggy

### DIFF
--- a/src/rdf4cpp/IRI.cpp
+++ b/src/rdf4cpp/IRI.cpp
@@ -134,34 +134,46 @@ IRI::operator std::string() const noexcept {
 }
 
 IRI IRI::default_graph(storage::DynNodeStoragePtr node_storage) {
-    auto const id = datatypes::registry::reserved_datatype_ids[datatypes::registry::default_graph_iri];
-    return IRI{storage::identifier::NodeBackendHandle{storage::identifier::literal_type_to_iri_node_id(id),
+    namespace reg = datatypes::registry;
+    namespace sid = storage::identifier;
+
+    auto const id = reg::reserved_datatype_ids[reg::default_graph_iri];
+    return IRI{sid::NodeBackendHandle{sid::literal_type_to_iri_node_id(id),
                node_storage}};
 }
 
 TriBool IRI::is_default_graph() const noexcept {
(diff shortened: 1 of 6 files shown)


src/rdf4cpp/IRI.hpp
```diff
@@ -134,6 +134,48 @@ struct IRI : Node {
...
+
+    /**
+     * Get the IRI for rdf:type
+     * @param node_storage optional custom node_storage where the returned IRI lives
+     * @return rdf:type IRI
+     */
+    [[nodiscard]] static IRI rdf_type(storage::DynNodeStoragePtr node_storage = storage::default_node_storage);
+
+    /**
+     * @return err if this is null, otherwise true iff this IRI is rdf:type
```
bigerl, line 147: add in the brief part something like `Check if this IRI is rdf:type`. The return type is correct but hard to parse.

src/rdf4cpp/IRI.cpp
```diff
@@ -149,6 +149,22 @@ TriBool IRI::is_default_graph() const noexcept {
     return this_id == expected_id;
 }
 
+IRI IRI::rdf_type(storage::DynNodeStoragePtr node_storage) {
+    auto const id = datatypes::registry::reserved_datatype_ids[datatypes::registry::rdf_type];
+    return IRI{storage::identifier::NodeBackendHandle{storage::identifier::literal_type_to_iri_node_id(id),
+                                                      node_storage}};
```
bigerl: I find it a bit easier to read that way. 

```
using reg = namespace datatypes::registry;
using namespace storage::identifier;
return IRI{NodeBackendHandle{literal_type_to_iri_node_id(reg::reserved_datatype_ids[reg::rdf_type]), node_storage}};
```

bigerl (changes requested):
Thx for the PR. Looks good. I just have two non-logic suggestions. See below.

bigerl (approved):

PR #370, opened 2025-05-19
add anyURI datatype

### LINKED ISSUE #368: add anyURI datatype
- [ ] fixed-id datatype
- [ ] cast to bool

mcb5637:
close #368

### DIFF
--- a/tests/datatype/tests_AnyURI.cpp
+++ b/tests/datatype/tests_AnyURI.cpp
@@ -9,7 +9,7 @@ TEST_CASE("anyURI capabilities") {
     static_assert(datatypes::LiteralDatatype<datatypes::xsd::AnyURI>);
     static_assert(datatypes::LogicalLiteralDatatype<datatypes::xsd::AnyURI>);
     static_assert(!datatypes::NumericLiteralDatatype<datatypes::xsd::AnyURI>);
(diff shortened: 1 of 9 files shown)


tests/datatype/tests_AnyURI.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <doctest/doctest.h>
+#include <rdf4cpp.hpp>
+
+using namespace rdf4cpp;
+
+TEST_CASE("anyURI capabilities") {
+    static_assert(datatypes::LiteralDatatype<datatypes::xsd::AnyURI>);
+    static_assert(datatypes::LogicalLiteralDatatype<datatypes::xsd::AnyURI>);
+    static_assert(!datatypes::NumericLiteralDatatype<datatypes::xsd::AnyURI>);
+    static_assert(!datatypes::PromotableLiteralDatatype<datatypes::xsd::AnyURI>);
```
liss-h: `anyURI` is promotable to `xsd:string` (here https://www.w3.org/TR/xpath-functions/#func-signatures, pretty far down it says "3. anyURI Type Promotion: A value of type xs:anyURI can be promoted to the type xs:string.")

liss-h (changes requested):
Thanks, code looks good just one thing

liss-h (approved):

PR #371, opened 2025-05-26
add macos 15 to CI

### LINKED ISSUE #364: add macos-15 to CI
GH runners are currently broken

- so use 3rd party instead
- make sure PRs from other people must be approved by us before actions can run. This is to avoid unplanned cost.

mcb5637:
close #364 

TODO:

- [x] point dependency provider back to main

### DIFF
--- a/.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
+++ b/.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
@@ -113,11 +113,11 @@ jobs:
 
       - name: Build Tests & Examples
         working-directory: build_dir
-        run: cmake --build . --parallel 2
+        run: cmake --build . --parallel $(nproc)
 
       - name: Run tests
         working-directory: build_dir
-        run: ctest --parallel 2 --verbose
+        run: ctest --parallel $(nproc) --verbose
 
       - name: Report coverage
         env:
--- a/.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
+++ b/.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
@@ -113,11 +113,11 @@ jobs:
 
       - name: Build Tests & Examples
         working-directory: build_dir
-        run: cmake --build . --parallel $(nproc)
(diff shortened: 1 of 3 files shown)


src/rdf4cpp/util/CaseInsensitiveCharTraits.hpp
```diff
@@ -17,19 +17,19 @@ struct CiCharTraits : std::char_traits<char> {
     using state_type = std::char_traits<char>::state_type;
     using comparison_category = std::char_traits<char>::comparison_category;
 
-    static constexpr bool eq(char_type const c1, char_type const c2) noexcept {
```
liss-h, line 20: I guess these functions are not `constexpr` on macos(?)

.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
```diff
@@ -77,9 +81,15 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Get dependency provider
+        if: ${{ inputs.os != 'macos-15' }}
         uses: dice-group/cpp-conan-release-reusable-workflow/.github/actions/add_conan_provider@main
 
+      - name: Get dependency provider (macos)
```
liss-h: This should be fixed in dice-group/cpp-conan-release-reusable-workflow directly (this means detecting macos irrespective of the runner name). Maybe macos has something like `/etc/os-release`

liss-h (changes requested):
Thank you, I just have one question and one change request

src/rdf4cpp/util/CaseInsensitiveCharTraits.hpp
mcb5637: they are not constexpr on linux either, but the compiler did not complain about it.

liss-h: c++ compilers..

liss-h:
I reran a previous workflow that succeeded before. It fails now, so it guess some update of the github runners broke it

liss-h (approved):

PR #373, opened 2025-05-26
Add dataset load integration test

### LINKED ISSUE #350: Add dataset load tests
Add tests that load datasets for integration testing purposes
 
for now just: 
- load dataset from file
- check size

mcb5637:
close #350

### DIFF
--- a/tests/graph/tests_loading.cpp
+++ b/tests/graph/tests_loading.cpp
@@ -22,7 +22,7 @@ TEST_CASE("load dataset") {
     });
     CHECK(set.size() == 4);
     int i = 0;
-    for (const auto& _ : set) {
+    for ([[maybe_unused]] const auto& _ : set) {
         ++i;
     }
     CHECK(i == 4);
--- a/tests/graph/tests_loading.cpp
+++ b/tests/graph/tests_loading.cpp
@@ -1,0 +1,19 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include <doctest/doctest.h>
+
+#include <rdf4cpp.hpp>
+
+using namespace rdf4cpp;
+
+TEST_CASE("load dataset") {
+    Dataset set{};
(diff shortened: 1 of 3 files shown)


tests/graph/tests_loading.cpp
```diff
@@ -0,0 +1,19 @@
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
+#include <doctest/doctest.h>
+
+#include <rdf4cpp.hpp>
+
+using namespace rdf4cpp;
+
+TEST_CASE("load dataset") {
+    Dataset set{};
```
liss-h, line 18: For dataset we should load a nquads or trig file. It can be smaller than swdf.
But you can keep this test by changing it to `Graph`

tests/graph/tests_loading.cpp
```diff
@@ -0,0 +1,19 @@
...
+    for (const auto& v : rdf4cpp::parser::RDFFileParser{"test_swdf/swdf.nt"}) {
```
liss-h: use `load_rdf_data` function instead of parser

liss-h (changes requested):
Thanks for the PR. I think we should also test quad data and use the appropriate `load_rdf_data` functions instead of the parser directly

tests/graph/tests_loading.cpp
```diff
@@ -0,0 +1,24 @@
...
+    CHECK(set.size() == 304592);
+}
+
+TEST_CASE("load dataset") {
+    Dataset set{};
+    std::fstream file{"tests_loading_simple.trig"};
+    set.load_rdf_data(file, parser::ParsingFlags::none(), nullptr, [](parser::ParsingError e) {
+        FAIL(std::format("{}", e));
+    });
+    CHECK(set.size() == 4);
```
liss-h, line 23: Please also check that the added graph has the correct size by itself using `set.graph(...)`

liss-h (approved):

PR #374, opened 2025-06-05
Feature: limited datatypes

### LINKED ISSUE #336: Some operations on `xsd:Integer` are unecessarily expensive
For example: casting an `xsd:Integer` to string requires the following in most cases
1. un-inline the integer value into a int64_t (this is true for most integers)
2. create a `boost::multiprecision` integer out of it
3. call the to string function for the `boost::multiprecision` integer

In theory it would be more optimal if we did not construct the boost integer, instead just converting the uninlined int64_t to string.

Not sure on the implementation yet.

### LINKED ISSUE #369: add size limits for unlimited datatypes
currently, the size of an integer is only limited by the available memory, which could lead to problems.
add a fixed size limit, generously over the limits required by the specifications.

- [x] integer (including negative, positive and non-negative, non-positive)
- [x] decimal
- [ ] also document this properly
- [ ] conversion double→integer

XSD required datatype ranges: https://www.w3.org/TR/xmlschema11-2/#partial-implementation

- for Decimal with $i/10^k$, 64-bit integer is enough for $i$ and for $k$ 1 Byte would be enough. ($i < 10^{16} < 2^{63}$)
- Conversion from double to decimal needs at least 1024 bit integer to be lossless.
    - boost fixed size 128 bit integer does not throw exceptions on overflow (most likely fallback to [[__int128](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fint128.html)](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fint128.html))

RDF4CPP should implement:

- arbitrary precision integer types: `__int128` and `unsigned __int128`
- decimals: i: 128 bit, k: 64 bit
- for conversion double→integer: temp 1024 bit integer, throw if 128 bit integer is exceeded

mcb5637:
close #369 

TODO:
- [x] specialized node storages
- [x] fallback to boost if __int128 is not available
- [x] #336 (__int128 is faster to construct)
- [x] documentation


README.md
```diff
@@ -113,11 +113,34 @@ cd build_dir
...
+
+For `http://www.w3.org/2001/XMLSchema#duration` (and all derived types), both parts have a separate signed 64-bit integer
+and with it its associated limits. The seconds part of the duration supports nanosecond precision.
+
 ### Additional CMake config options:
 
 - `-DBUILD_EXAMPLES=ON/OFF [default: OFF]`: Build the examples.
 - `-DBUILD_TESTING=ON/OFF [default: OFF]`: Build the tests.
 - `-DBUILD_SHARED_LIBS=ON/OFF [default: OFF]`: Build a shared library instead of a static one.
+- `-DRDF4CPP_USE_UNLIMITED_DATATYPES=ON`: If defined, uses unlimited datatypes for integer and decimal types.
```
liss-h: I don't think we need this. Its more maintenance overhead for us and we won't use it

liss-h (changes requested):
First mini review: I didn't really look at anything, but the one thing I noticed is that you added an option to stay unlimited.

For the benefit of maintenance we should only support the limited integers.

Sorry that we weren't clear enough on that

CMakeLists.txt
```diff
@@ -178,6 +178,17 @@ target_link_libraries(rdf4cpp
         highway::highway
         )
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
liss-h: Since this is only needed on Linux ARM (if possible) please also make this conditional on the target

src/rdf4cpp/util/Int128.hpp
```diff
@@ -0,0 +1,397 @@
...
+#include <rdf4cpp/writer/BufWriter.hpp>
+#include <dice/hash.hpp>
+
+namespace rdf4cpp::util {
+#ifdef __SIZEOF_INT128__
+    using Int128 = __int128;
+#else
+    using Int128 = boost::multiprecision::checked_int128_t;
+#endif
+    using BigInt = Int128;
```
liss-h: Unecessary typdef

src/rdf4cpp/util/Int128.hpp
```diff
@@ -0,0 +1,397 @@
...
+        using cpp_int_unchecked = boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, boost::multiprecision::unchecked, Alloc>>;
+
+        template<typename T>
+        concept IntegralExt = std::integral<T> || std::same_as<T, __int128> || std::same_as<T, unsigned __int128>;
+        template<typename T>
+        concept BoostNumber = std::same_as<T, boost::multiprecision::number<typename T::backend_type, T::et>>;
+
+        template<OverflowMode m, typename T>
+        requires IntegralExt<T>
+        static constexpr bool add_checked(T const &a, T const &b, T &result) noexcept {
```
liss-h: I think it would make sense to add a `CheckedInt` specialization for boost mp types instead of putting all the functions here

src/rdf4cpp/util/Int128.hpp
```diff
@@ -0,0 +1,397 @@
...
+    using Int128 = __int128;
+#else
+    using Int128 = boost::multiprecision::checked_int128_t;
+#endif
+    using BigInt = Int128;
+
+    // string -> int128 can be found in CharConvExt
+
+#ifdef __SIZEOF_INT128__
+    inline bool to_chars_canonical(__int128 value, writer::BufWriterParts const writer) noexcept {
```
liss-h: I have a suggestion for this. Since I would assume the standard library should be pretty good at serializing integers we could utilize `std::to_chars` for this and avoid a bunch of math (maybe). Instead of dividing by 10 we divide by the largest power of 10 that fits into a 64 bit integer:

```c++
inline constexpr unsigned __int128 max_pow10 = 10'000'000'000'000'000'000ull;

std::to_chars_result to_chars(char *first, char *last, unsigned __int128 val) noexcept {
    if (val <= std::numeric_limits<uint64_t>::max()) {
        return std::to_chars(first, last, static_cast<uint64_t>(val));
    }

    std::to_chars_result upper = to_chars(first, last, val / max_pow10);
    if (upper.ec != std::errc{}) {
        return upper;
    }

    first = upper.ptr;

    std::to_chars_result lower = std::to_chars(first, last, static_cast<uint64_t>(val % max_pow10));
    return lower;
}

std::to_chars_result to_chars(char *first, char *last, __int128 val) noexcept {
    if (val >= 0) {
        return to_chars(first, last, static_cast<unsigned __int128>(val));
    }

    if (first == last) [[unlikely]] {
        return std::to_chars_result{.ptr = first, .ec = std::errc::value_too_large};
    }

    *first++ = '-';
    return to_chars(first, last, static_cast<unsigned __int128>(-val));
}
```

Let me know if I overlooked something

liss-h (changes requested):
Thanks for the PR, this is the first review. I'll review again after the changes have been made.

CMakeLists.txt
liss-h: Getting rid of this completely would also be good, I'll look into that

liss-h: I found a solution: 
The bug is no longer present in clang-20, so we simply require clang>=20 on aarch64 linux. 

If people do not have access to clang 20 on arm, they can use gcc.

src/rdf4cpp/util/Int128.hpp
mcb5637: well, `-std::numeric_limits<__int128>::min()` is a overflow, but we can work around that

src/rdf4cpp/util/Int128.hpp
mcb5637: well, `CheckedInt` and `fun_checked` are designed for different purposes.
`fun_checked` is supposed to have its return value checked immediately
while `CheckedInt` is supposed to be usable by any algorythm not aware of overflow checks

conanfile.py
```diff
@@ -58,6 +59,14 @@ def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
 
+    def generate(self):
```
liss-h: left over. Some other things in this file as well

src/rdf4cpp/datatypes/registry/util/CharConvExt.hpp
```diff
@@ -87,6 +88,82 @@ F from_chars(std::string_view s) {
...
+        auto n = std::numeric_limits<uint64_t>::digits10;
+        __int128 d = 1;
+        for (int i = 0; i < n; ++i) {
+            d = d * 10;
+        }
+        return d;
+    }();
+    static_assert(max_pow10 == 10'000'000'000'000'000'000ull);
+
+    if (s == "-170141183460469231731687303715884105728") {
```
liss-h: `[[unlikely]]`

liss-h: This check can be bypassed by adding a zero after the minus sign

src/rdf4cpp/datatypes/xsd/Decimal.cpp
```diff
@@ -70,7 +69,7 @@ nonstd::expected<capabilities::Numeric<xsd_decimal>::div_result_cpp_type, Dynami
         return nonstd::make_unexpected(DynamicError::DivideByZero);
     }
 
-    auto r = lhs.div_checked(rhs, 1000);
+    auto r = lhs.div_checked(rhs, 20);
```
liss-h: please define a constant somewhere for this. Without context I have no idea why it has to be 20

src/rdf4cpp/datatypes/xsd/integers/signed/Integer.cpp
```diff
@@ -3,32 +3,46 @@
...
 bool capabilities::Logical<xsd_integer>::effective_boolean_value(cpp_type const &value) noexcept {
     return value != 0;
 }
 
+template<>
+nonstd::expected<capabilities::Numeric<xsd_integer>::add_result_cpp_type, DynamicError> capabilities::Numeric<xsd_integer>::add(cpp_type const &lhs, cpp_type const &rhs) noexcept {
+    // https://www.w3.org/TR/xpath-functions/#op.numeric
+    // needs overflow protection
+    cpp_type r;
+    if (rdf4cpp::util::detail::add_checked<rdf4cpp::util::detail::OverflowMode::Checked>(lhs, rhs, r))
```
liss-h: missing braces (see our coding standard), also in other places in this PR

tests/datatype/tests_NumOpResults.cpp
```diff
@@ -48,10 +50,7 @@ TEST_SUITE("numeric op results") {
 
             // "On underflow, 0.0 must be returned."
             if (min != 0) {  // if min == 0 -> unlimited
-                CHECK(Decimal::sub(min, min) == zero);
-                CHECK(Decimal::div(min, two) == zero);
```
liss-h, line 52: Why did you remove these checks?

src/rdf4cpp/BigDecimal.hpp
```diff
@@ -240,7 +180,7 @@ struct BigDecimal {
         if (significand == 0) {
             return;
         }
-        while (significand & 1) {
+        while ((significand & 1) == 0) {
```
liss-h: was this incorrect before?

src/rdf4cpp/BigDecimal.hpp
```diff
@@ -250,22 +190,50 @@ struct BigDecimal {
...
+public:
+    /**
+     * converts a double to a BigDecimal.
+     * this conversion might not be exact, due to the built in limitations of doubles.
+     * if you have the possibility, use one of the other constructors.
+     * @param value
+     * @throw std::overflow_error on exceeding the types numeric limits
+     */
+    explicit BigDecimal(double value) : unscaled_value(0), exponent(0) {
+        if constexpr (!std::numeric_limits<UnscaledValue_t>::is_bounded) {
```
liss-h: tbh, I would just remove the template parameters if you need to add special cases for different instantiations

src/rdf4cpp/util/Int128.hpp
```diff
@@ -0,0 +1,425 @@
...
+        }
+        template<OverflowMode m, std::size_t MinBits, std::size_t MaxBits, boost::multiprecision::cpp_integer_type SignType, typename Alloc>
+        static constexpr bool add_checked(cpp_int_checked<MinBits, MaxBits, SignType, Alloc> const &a,
+                                          cpp_int_checked<MinBits, MaxBits, SignType, Alloc> const &b,
+                                          cpp_int_checked<MinBits, MaxBits, SignType, Alloc> &result) noexcept {
+            if (m == OverflowMode::Checked) {
+                try {
+                    result = a + b;
+                } catch (std::overflow_error const &) {
+                    return true;
```
liss-h: these return values are kind of the wrong way around for the function name. That should at least be document (maybe with a comment at the top of the file)

src/rdf4cpp/BigDecimal.hpp
```diff
@@ -727,18 +703,20 @@ struct BigDecimal {
...
-                return std::strong_ordering::less;  // t does fit into the same precision, while o does not
-            if (mul_checked<OverflowMode::Checked>(o, b, o))
-                return std::strong_ordering::less;  // t does fit into the same precision, while o does not
-        } else if (this->exponent < other.exponent) {
-            UnscaledValue_t b{base};
-            if (pow_checked<OverflowMode::Checked>(b, other.exponent - this->exponent, b))
-                return std::strong_ordering::greater;  // o does fit into the same precision, while t does not
-            if (mul_checked<OverflowMode::Checked>(t, b, t))
-                return std::strong_ordering::greater;  // o does fit into the same precision, while t does not
+        if (t != 0 && o != 0) {  // if one of the decimals is 0, compare directly (0*base^e is 0 for all e)
```
liss-h: the comment does not match the logic. The condition just says that either must not be null, both could be non-null

src/rdf4cpp/util/Int128.hpp
```diff
@@ -0,0 +1,425 @@
...
+                }
+                return false;
+            } else {
+                using Unc = cpp_int_unchecked<MinBits, MaxBits, SignType, Alloc>;
+                result = cpp_int_checked<MinBits, MaxBits, SignType, Alloc>{Unc{a} + Unc{b}};
+                return false;
+            }
+        }
+        template<OverflowMode m, std::size_t MinBits, std::size_t MaxBits, boost::multiprecision::cpp_integer_type SignType, typename Alloc>
+        static constexpr bool add_checked(cpp_int_unchecked<MinBits, MaxBits, SignType, Alloc> const &a,
```
liss-h: Are these functions with `cpp_int_unchecked` needed? I didn't find any usages with a quick search

liss-h: A quick test on the branch revealed that they are indeed all unused

src/rdf4cpp/util/Int128.hpp
```diff
@@ -0,0 +1,425 @@
+#ifndef RDF4CPP_INT128_HPP
+#define RDF4CPP_INT128_HPP
+
+#include <boost/multiprecision/cpp_int.hpp>
+#include <rdf4cpp/writer/BufWriter.hpp>
+#include <dice/hash.hpp>
+
+namespace rdf4cpp::util {
+#ifdef __SIZEOF_INT128__
+    using Int128 = __int128;
```
liss-h, line 12: `Int128` should not be in util namespace.

Just `rdf4cpp::Int128`.
Also for use in `xsd:decimal` define `using Decimal128 = BigDecimal<>` and use that instead

liss-h (changes requested):
Thanks for the changes, I have some more comments

tests/datatype/tests_NumOpResults.cpp
mcb5637: XPath (and by extension SPARQL) do not define underflow themselves. But from how they use it, (see https://www.w3.org/TR/xpath-functions/#op.numeric ) it seems, they understand it the same as it is used in IEEE floats.
And for these (as far as i can tell, i don't seem to have access to the standard) ( https://ieeemilestones.ethw.org/w/images/5/54/Handbook_Floating_Point_Chapter_3.pdf page 31) underflow occurs if the absolute of the result is not 0 but smaller than the smallest representable value.

In which case it makes a lot more sense to return 0.

TLDR: they don't test what they are supposed to test.

src/rdf4cpp/BigDecimal.hpp
mcb5637: yes, although it only affects some very rare circumstances.

src/rdf4cpp/BigDecimal.hpp
mcb5637: Its only this conversion. And both are used.

liss-h:
blocked by BETA

liss-h (changes requested):
Please add some additional parsing tests for `Int128` (mostly larger numbers, or large negative numbers) so that we can be sure that it actually works correctly

src/rdf4cpp/util/Int128.hpp
```diff
@@ -0,0 +1,338 @@
...
+
+#include <boost/multiprecision/cpp_int.hpp>
+#include <dice/hash.hpp>
+#include <rdf4cpp/writer/BufWriter.hpp>
+
+namespace rdf4cpp {
+#ifdef __SIZEOF_INT128__
+    using Int128 = __int128;
+#else
+    using Int128 = boost::multiprecision::checked_int128_t;
```
liss-h: I don't recall if we talked about this before but I don't think we need to support platforms where __int128 does not exist.

liss-h: So we can get rid of all the `#ifdef __SIZEOF_INT128__`

src/rdf4cpp/util/Int128.hpp
```diff
@@ -0,0 +1,338 @@
...
+            static constexpr bool mul_checked(T const &a, T const &b, T &result) noexcept {
+                if constexpr (m == OverflowMode::Checked) {
+                    return __builtin_mul_overflow(a, b, &result);
+                } else {
+                    result = a * b;
+                    return false;
+                }
+            }
+            template<OverflowMode m, std::size_t MinBits, std::size_t MaxBits, boost::multiprecision::cpp_integer_type SignType, typename Alloc>
+            static constexpr bool mul_checked(cpp_int_checked<MinBits, MaxBits, SignType, Alloc> const &a,
```
liss-h: Please move all this boost stuff to a separate header

liss-h (changes requested):
Hi, this review was so long ago that I barely remember anything.
I will probably have to go through everything again, but I have atleast some initial comments for you.

src/rdf4cpp/util/Int128.hpp
```diff
@@ -0,0 +1,199 @@
...
+                }
+                result = static_cast<To>(f);
+                return false;
+            }
+        }  // namespace detail
+    }  // namespace util
+}  // namespace rdf4cpp
+
+template<typename Policy>
+struct dice::hash::dice_hash_overload<Policy, __int128> {
```
liss-h: use https://github.com/dice-group/dice-hash/releases/tag/v0.4.12

liss-h:
POBR breaking change but develop already contains a bump

tests/serializer/tests_Serialize.cpp
```diff
@@ -62,7 +62,7 @@ TEST_SUITE("Serialize") {
             run_ser_test(lit, "\"simple\"");
         }
         SUBCASE("non-inlined xsd:integer") {
-            auto lit = Literal::make_typed_from_value<xsd::Integer>(xsd::Integer::cpp_type{"87960930222089"});
+            auto lit = Literal::make_typed<xsd::Integer>("87960930222089");
```
nkaralis: why is this needed? is the result different?

src/rdf4cpp/datatypes/xsd/Decimal.cpp
```diff
@@ -71,7 +71,7 @@ nonstd::expected<capabilities::Numeric<xsd_decimal>::div_result_cpp_type, Dynami
         return nonstd::make_unexpected(DynamicError::DivideByZero);
     }
 
-    auto r = lhs.div_checked(rhs, 1000);
+    auto r = lhs.div_checked(rhs, cpp_type::default_max_scale_increase);
```
nkaralis: according to the standard this should not use `RoundingMode::Floor` (truncated or rounded). please double check

src/rdf4cpp/BigDecimal.hpp
```diff
@@ -1,931 +1,898 @@
...
+                            throw InvalidNode{"http://www.w3.org/2001/XMLSchema#decimal parsing error: more than one . found"};
+                        }
+                        decimal = true;
+                        continue;
+                    }
+                    if (c < '0' || c > '9') {
+                        throw InvalidNode{"http://www.w3.org/2001/XMLSchema#decimal parsing error: non-numeric char found"};
+                    }
+                    auto n = c - '0';
+                    if (detail::mul_checked<OverflowMode::Checked>(unscaled_value, UnscaledValue_t{base}, unscaled_value)) {
```
nkaralis: considering trailing zeros and the canonical .0 here leads to overflows and representable numbers are rejected

src/rdf4cpp/BigDecimal.hpp
```diff
@@ -1,931 +1,898 @@
...
+            /**
+             * division of two BigDecimals.
+             * checks overflow and division by 0.
+             * @param other
+             * @param max_scale_increase after this many decimal places have been added, stop dividing and round
+             * @param mode rounding mode to use
+             * @return
+             */
+            [[nodiscard]] constexpr nonstd::expected<BigDecimal, DecimalError> div_checked(BigDecimal const &other, Exponent_t max_scale_increase, RoundingMode mode = RoundingMode::Floor) const noexcept {
+                if (other.unscaled_value == 0)
```
nkaralis, line 615: please check the following test cases
```
SUBCASE("scaling the dividend must not overflow when the result fits") {
    Dec const max{std::numeric_limits<rdf4cpp::Int128>::max(), 0};
    CHECK(max.div_checked(Dec{"2.5"}, 20) == Dec{"68056473384187692692674921486353642290.8"});
    CHECK(Dec{"170000000000000000000000000000000000000"}.div_checked(Dec{"0.5"}, 20) == Dec{"340000000000000000000000000000000000000"});
    CHECK(Dec{"100000000000000000000000000000000000000"}.div_checked(Dec{"0.25"}, 20) == Dec{"400000000000000000000000000000000000000"});
}
```

nkaralis (commented):
I made some comments in BigDecimal.hpp

src/rdf4cpp/BigDecimal.hpp
```diff
@@ -1,931 +1,898 @@
...
+                return unscaled_value >= 0;
+            }
 
-    [[nodiscard]] constexpr UnscaledValue_t get_unscaled_value() const noexcept {
-        return unscaled_value;
-    }
+        private:
+            // op_checked has to be add_checked or sub_checked with the same OverflowMode
+            template<OverflowMode m, bool (*op_checked)(UnscaledValue_t const &t, UnscaledValue_t const &o, UnscaledValue_t &result)>
+            constexpr bool add_or_sub(BigDecimal const &other, BigDecimal &result) const noexcept {
```
nkaralis, line 305: wide intermediate results make the numeric operations overflow, even if the final result would fit? can we use wider types for this cases only?

src/rdf4cpp/datatypes/xsd/Decimal.cpp
```diff
@@ -185,6 +179,54 @@ capabilities::Inlineable<xsd_decimal>::cpp_type capabilities::Inlineable<xsd_dec
     auto const exponent = util::unpack_integral<uint32_t, InlinedDecimal::exponent_size>(data.exponent);
     return cpp_type{unscaled_value, exponent};
 }
+
+template<>
+template<>
+capabilities::Promotable<xsd_decimal>::promoted_cpp_type<0> capabilities::Promotable<xsd_decimal>::promote<0>(cpp_type const &value) noexcept {
```
nkaralis, line 185: `from_string` throws for small values (`xsd:decimal 1e-60  cast to xsd:float`). this should not be `noexcept`

src/rdf4cpp/datatypes/xsd/Decimal.cpp
```diff
@@ -185,6 +179,54 @@ capabilities::Inlineable<xsd_decimal>::cpp_type capabilities::Inlineable<xsd_dec
...
+    // 19.1.2.1 Casting to xs:float
+    // If ST is xs:decimal then TV is xs:float(SV cast as xs:string) and the conversion is complete.
+    // https://www.w3.org/TR/xpath-functions/#casting-to-float
+    auto const str = writer::StringWriter::oneshot([&value](auto &w) {
```
nkaralis: how long can this string get? do we need to do something against large strings?

nkaralis (commented):
Thank you for the latest changes. Two comments from my side

src/rdf4cpp/datatypes/xsd/Decimal.cpp
liss-h: decimal is limited so the string is limited.

nkaralis (approved):
Thank you

PR #375, opened 2025-06-10
Fix: parsing of absolute IRI with dots

mcb5637:
this test failed before: https://github.com/w3c/rdf-tests/blob/main/sparql/sparql10/i18n/normalization-02.ttl


tests/nodes/tests_IRIFactory.cpp
```diff
@@ -190,6 +194,7 @@ TEST_CASE("prefix") {
     CHECK(fact.from_prefix("pre", "bar").value().identifier() == "http://ex.org/bar");
     CHECK(fact.from_prefix("foo", "bar").value().identifier() == "http://foo.org/bar");
     CHECK(fact.from_prefix("bar", "bar").error() == IRIFactoryError::UnknownPrefix);
+    CHECK(fact.from_prefix("pre", "a/./b/../c").value().identifier() == "http://ex.org/a/./b/../c");
```
nkaralis, line 197: please also add a test case with the prefix having dots

nkaralis (changes requested):
Thank you for the PR. I only have one comment.

nkaralis (approved):
thank you

PR #379, opened 2025-06-20
Fix: date/time comparison

mcb5637:


src/rdf4cpp/datatypes/registry/util/DateTimeUtils.hpp
```diff
@@ -243,29 +243,15 @@ static inline std::partial_ordering compare_time_points(const rdf4cpp::TimePoint
...
-            auto b_sys = apply_timezone(b, *btz);
-            auto p14 = cmp_opt(apply_timezone(a, rdf4cpp::Timezone::max_value()), b_sys);
-            auto m14 = cmp_opt(apply_timezone(a, rdf4cpp::Timezone::min_value()), b_sys);
-            if (p14 != m14)
-                return std::partial_ordering::unordered;
-            return p14;
-        } else {
-            return cmp(a, b);
-        }
+    if (!atz.has_value()) {
```
nkaralis, line 246: if i understand correctly, if any of the operands do not have a timezone, a default one be applied to them. is that correct?

src/rdf4cpp/datatypes/registry/util/DateTimeUtils.hpp
```diff
@@ -243,29 +243,15 @@ static inline std::partial_ordering compare_time_points(const rdf4cpp::TimePoint
...
+        atz = rdf4cpp::util::time_point_replacement_timezone;
     }
+    if (!btz.has_value()) {
+        btz = rdf4cpp::util::time_point_replacement_timezone;
+    }
+
+    auto a_sys = apply_timezone(a, *atz);
+    return cmp_opt(a_sys, apply_timezone(b, *btz));
```
nkaralis, line 254: does `cmp_opt` need the if clause?

nkaralis (commented):
Thank you. I have two questions.

src/rdf4cpp/datatypes/registry/util/DateTimeUtils.hpp
mcb5637: yes

src/rdf4cpp/datatypes/registry/util/DateTimeUtils.hpp
mcb5637: yes, any timepont might overflow when applying the timezone. and in that case, the optional has no value.

nkaralis (approved):
Thank you

PR #386, opened 2025-06-26
Switch back to official github runners for macOS

liss-h:
Since now both the official runners and warp are broken, we can switch back to the official ones.


liss-h:
obviously we can ignore macos failures here since the runner is broken

bigerl (approved):

PR #388, opened 2025-06-26
Use macports instead of brew for gcc installation

bigerl:
works but there is a strange error with CMake in the static configuration. 

caching was tested and seems to work.


liss-h (changes requested):
Doesn't work yet, see workflow output

liss-h (approved):

PR #389, opened 2025-07-02
Fix: binops with unknown datatypes

liss-h:
Fixes an issue where a binary operation with an unknown right hand side would trigger an assertion.

Additionally, fixed an inconsistency when comparing uncomparable datatypes.

### DIFF
--- a/src/rdf4cpp/datatypes/xsd/Base64Binary.hpp
+++ b/src/rdf4cpp/datatypes/xsd/Base64Binary.hpp
@@ -85,7 +85,8 @@ extern template struct LiteralDatatypeImpl<xsd_base64_binary,
 namespace rdf4cpp::datatypes::xsd {
 
 struct Base64Binary : registry::LiteralDatatypeImpl<registry::xsd_base64_binary,
-                                                    registry::capabilities::FixedId> {};
+                                                    registry::capabilities::FixedId,
+                                                    registry::capabilities::Comparable> {};
 
 } // namespace rdf4cpp::datatypes::xsd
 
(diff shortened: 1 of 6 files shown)


src/rdf4cpp/datatypes/xsd/HexBinary.hpp
```diff
@@ -112,7 +112,8 @@ extern template struct LiteralDatatypeImpl<xsd_hex_binary,
 namespace rdf4cpp::datatypes::xsd {
 
 struct HexBinary : registry::LiteralDatatypeImpl<registry::xsd_hex_binary,
-                                                 registry::capabilities::FixedId> {};
+                                                 registry::capabilities::FixedId,
+                                                 registry::capabilities::Comparable> {};
```
nkaralis, line 116: is this only for `ORDER BY`?

src/rdf4cpp/datatypes/xsd/Base64Binary.hpp
```diff
@@ -85,7 +85,8 @@ extern template struct LiteralDatatypeImpl<xsd_base64_binary,
 namespace rdf4cpp::datatypes::xsd {
 
 struct Base64Binary : registry::LiteralDatatypeImpl<registry::xsd_base64_binary,
-                                                    registry::capabilities::FixedId> {};
+                                                    registry::capabilities::FixedId,
+                                                    registry::capabilities::Comparable> {};
```
nkaralis, line 89: is this only for ORDER BY?

nkaralis (commented):
Thank you. I only have one question.

src/rdf4cpp/datatypes/xsd/HexBinary.hpp
liss-h: also FILTER, ORDER BY always worked on the binary types because of the fallback ordering based on lexical form

nkaralis (approved):
Thank you

PR #391, opened 2025-07-03
change time datatypes to builtin int128

mcb5637:
changes date/time 128 bit integer from boost to builtin `__int128`, to make calculations a lot faster.

while the builtin has a smaller range than boost, both are bigger than what the `int64_t` year allows. therefore the usable range of date/time types stays the same.

also fixes some possible (but very unlikely in regular use) overflow errors in date/time calculations (to trigger them, you need to use very big numbers, at least several orders of magnitude bigger than the end of the universe).


liss-h (changes requested):
Could you add a little more context to the PR description? Like a summary of the changes and what the implications are (limits and such)

liss-h:
Note to self, review when 128bit integer PR is merged

PR #401, opened 2025-08-06
Feature: add trigonometry & exponential funcs

### LINKED ISSUE #126: Feature: XPATH 4.8 Trigonometric and exponential functions
note: those functions are handy but not part of the SPARQL 

https://www.w3.org/TR/xpath-functions/#trigonometry

- [ ] math:pi
- [ ] math:exp
- [ ] math:exp10
- [ ] math:log
- [ ] math:log10
- [ ] math:pow
- [ ] math:sqrt
- [ ] math:sin
- [ ] math:cos
- [ ] math:tan
- [ ] math:asin
- [ ] math:acos
- [ ] math:atan
- [ ] math:atan2

mcb5637:
close #126 

where xpath only defines these functions specifically for doubles, all other types result in an error (null Literal returned). We allow all types as arguments that can be converted to double.

### DIFF
--- a/src/rdf4cpp/Literal.cpp
+++ b/src/rdf4cpp/Literal.cpp
@@ -2608,21 +2608,7 @@ if (!var_name##_opt.has_value()) {                                              
 auto const var_name = *var_name##_opt;
 
 Literal Literal::math_exp(storage::DynNodeStoragePtr node_storage) const {
-
-    auto const th_opt = [&source_obj = (*this)]() -> std::optional<double> {
-        if (source_obj.null()) {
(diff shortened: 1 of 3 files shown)


bigerl (changes requested):
From my side this looks good after extending it to accept anything that is convertible to double. I changed that is [also how others implement it](https://graphdb.ontotext.com/documentation/9.6/enterprise/using-math-functions-with-sparql.html). 

@nkaralis @mcb5637 please double check that I didn't mess anything up.

bigerl (approved):

PR #407, opened 2025-09-30
Prevent exceptions of unintended type from escaping from parser

liss-h:
This was not actually an issue, but it is a good idea to have.


bigerl (approved):
Totally makes sense and looks good.

PR #408, opened 2025-10-02
Feature: Make regex from literal

nkaralis:


liss-h (approved):
Thank you for the PR. I've made some more changes (not really related to this PR, I just found some more things that were odd)

PR #410, opened 2025-11-06
Feature: RDF/XML parser

### LINKED ISSUE #403: XML parser
use https://gitlab.gnome.org/GNOME/libxml2/-/wikis/home (see https://conan.io/center/recipes/libxml2?version=2.14.5) 

to extend the rdf4cpp parser with RDF/XML support. Use the SAX2 API and make sure no DOM is built for the XML document.

mcb5637:
TODO:

- [x] ID reification
- [x] ID exclusivity
- [x] lang tag
- [x] xml literal
- [x] cleanup
- [x] check dependencies
- [x] integrate into IStreamQuadIterator


src/rdf4cpp/parser/ParsingFlags.hpp
```diff
@@ -21,6 +21,9 @@ enum struct ParsingFlag : uint8_t {
     NTriples = 0b01 << 4,
     NQuads   = 0b10 << 4,
     TriG     = 0b11 << 4,
+    RdfXml  = 0b100 << 4,
+
+    SyntaxMask = 0b111 << 4,
```
liss-h: This is not a syntax (and not a parsing flag either). Please it define as a constant outside of the enum

tests/bench_SerDe.cpp
```diff
@@ -27,6 +27,7 @@ void deserialize(std::filesystem::path const &in_path, Dataset &ds, storage::Dyn
     parser::IStreamQuadIterator qit{in_file,
                                     reinterpret_cast<parser::ReadFunc>(&fread),
                                     reinterpret_cast<parser::ErrorFunc>(&ferror),
+                                        reinterpret_cast<parser::EOFFunc>(&feof),
```
liss-h: indentation is messed up

src/rdf4cpp/parser/IStreamQuadIterator.cpp
```diff
@@ -38,19 +39,33 @@ static int istream_error(void *voided_self) noexcept {
...
+
 IStreamQuadIterator::IStreamQuadIterator(void *stream,
                                          ReadFunc read,
                                          ErrorFunc error,
+                                         EOFFunc eof,
                                          flags_type flags,
                                          state_type *state)
-    : impl{std::make_unique<Impl>(stream, read, error, flags, state)},
+    : impl{flags.get_syntax() == ParsingFlag::RdfXml ?
+        make_xml_impl(stream, read, error, eof, state) :
```
liss-h: why does this have a helper function to construct it but serd does not?

private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.hpp
```diff
@@ -14,17 +14,25 @@
...
+    // see https://github.com/NVIDIA/stdexec/issues/1143
+    struct SerdReaderDtorLambda {
+        void operator()(SerdReader* r) const {
+            serd_reader_end_stream(r);
+            serd_reader_free(r);
+        }
+    };
+    std::unique_ptr<SerdReader, SerdReaderDtorLambda> reader;
 
+    std::unique_ptr<state_type> state_owned = nullptr;
```
liss-h: please revert this unique_ptr hack

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,896 @@
...
+                            xmlFreeParserCtxt(c);
+                        })>
+                context_;
+        void *reader_obj_;
+        ReadFunc read_func_;
+        ErrorFunc error_func_;
+        EOFFunc eof_func_;
+        std::deque<value_type> result_queue_;
+        size_t next_bn_index_ = 0;
+        std::unique_ptr<state_type> owned_state_;
```
liss-h: same here. Please don't use unique_ptr this way (you already have another pointer that you can just call delete on in the dtor)

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,896 @@
...
+        state_type *state_;
+        dice::sparse_map::sparse_set<IRI> reserved_ids_;
+
+        static constexpr std::string_view reify_subject = "http://www.w3.org/1999/02/22-rdf-syntax-ns#subject";
```
liss-h: please create an issue that we force these IRIs (and the ones below) to always be present in the node storage (not in this PR, later)

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,896 @@
...
+
+    void IStreamQuadIterator::ImplXML::update_current_state() {
+        if (state_stack_.empty()) {
+            current_state_ = nullptr;
+            return;
+        }
+        current_state_ = std::visit([](auto &s) -> BaseState * {
+            return &s;
+        },
+                                    state_stack_.back());
```
liss-h: indentation messed up

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,896 @@
...
+            void on_characters(ImplXML &i, std::string_view chars) override;
+            void on_start_element(ImplXML &i, std::string_view local_name, std::string_view uri, std::span<Attribute> attributes) override;
+            void on_end_element(ImplXML &i) override;
+
+            EmptyElement()
+                : BaseState({}) {
+            }
+        };
+
+        BaseState *current_state_ = nullptr;
```
liss-h: Use inplace_polymorphic instead of the hack with variant and virtual base

liss-h (changes requested):
first review about architecture and design (not logic).

In general: please split the xml parse into multiple files. One file per state would be good. And please add comments to explain why stuff is happening, what each state is good for and so on. This is very hard to follow even if you know what rdf-xml roughly looks like.

I very much like the idea with the states. Just needs to be organized a little.

private/rdf4cpp/parser/XMLParser.cpp
liss-h: actually for this it does not help. disregard my comment

liss-h:
If you split it into multiple files just keep the definitions inline in the structs that makes it easier to read I think

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,896 @@
...
+                context_;
+        void *reader_obj_;
+        ReadFunc read_func_;
+        ErrorFunc error_func_;
+        EOFFunc eof_func_;
+        std::deque<value_type> result_queue_;
+        size_t next_bn_index_ = 0;
+        std::unique_ptr<state_type> owned_state_;
+        state_type *state_;
+        dice::sparse_map::sparse_set<IRI> reserved_ids_;
```
liss-h: putting IRI directly into a set is not optimal. You could probably use NodeBackendID since they should all come from the same node storage

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,896 @@
...
+
+#include <ranges>
+
+#include <libxml/parser.h>
+
+#include <dice/sparse-map/sparse_set.hpp>
+
+namespace rdf4cpp::parser {
+    // not in a header, because that would complicate linking it against libxml
+    // probably even requiring to make libxml public
```
liss-h: 👍 thats fine

mcb5637:
Splitting the states into multiple files is going to be complicated, because they would need headers which would expose libxml.

src/rdf4cpp/parser/IStreamQuadIterator.cpp
mcb5637: Because ImplXML is hidden in another cpp file without a header.
The reason for that is to not expose libxml related types into any header.

liss-h:
> Splitting the states into multiple files is going to be complicated, because they would need headers which would expose libxml.

It is possible with private includes.

src/rdf4cpp/parser/IStreamQuadIterator.cpp
liss-h: private includes solve this

private/rdf4cpp/parser/XMLParser.hpp
```diff
@@ -0,0 +1,132 @@
...
+        friend struct InitialState;
+        friend struct RDFState;
+        friend struct DescriptionState;
+        friend struct PredicateState;
+        friend struct TypedLiteralPredicateState;
+        friend struct EmptyElement;
+        friend struct XMLLiteralState;
+        friend struct CollectionState;
+
+        BaseState *current_state_ = nullptr;
```
liss-h: Please don't create self-referential structs.

liss-h (changes requested):
Thank you for the changes.

As I said before, I do like the idea of the states. However, the concrete implementation is not optimal.

The main concern here is the circular nature of the relationships of the states and the parser. The states can just arbitrarily modify the parser through the given `ImplXML &`. This is not good for general understandability and correctness.

I suggest a more functional approach with state transitions and secondary actions (something like `fn(CurrentState, Input) -> Transition`).

The state transitions determine which state the parser will enter next (or that it should pop a state). The secondary actions are emitting triples or errors.

Something like this:
```c++
struct Transition {
    using Action = std::variant<StatePop, StatePush, EmitStatement, EmitError>;
    std::vector<Action> actions_;
};
```

So instead of directly mutating the parser, the states tell the parser what to do (functional/data oriented design). There is a closed set of possible StateTransition actions that each have a clear purpose. (Rust can model this much cleaner than C++, but it still works). Performance is mostly a secondary concern here; maintainability is more important for rdf/xml (due to the complexity of the specification and since there are basically no large rdf/xml datasets).

The whole goal is to get rid of the circular mutability that is currently present and limit what states can do to the parser. That should make it both safer and easier to follow.

For this it makes sense to move some of the functionality out of the `ImplXML` (e.g. the static functions).

Note: I have not fully thought this through (since it would require me to do the refactor myself), but if you have questions or concerns, feel free to ask.

liss-h (changes requested):
Please remove the uses of nested structs (you can keep the ones where `ImplXYZ` is nested in `IStreamQuadIterator` for now)

private/rdf4cpp/parser/XMLParserStateCollector.hpp
```diff
@@ -0,0 +1,165 @@
...
+#include <rdf4cpp/parser/IStreamQuadIterator.hpp>
+#include <rdf4cpp/parser/ParsingError.hpp>
+#include <rdf4cpp/parser/ParsingState.hpp>
+
+#include <libxml/parser.h>
+
+#include <dice/sparse-map/sparse_set.hpp>
+
+namespace rdf4cpp::parser {
+    struct IStreamQuadIterator::ImplXMLStateCollector : Impl {
```
liss-h: I don't get the point of this. Why is it not just in the XmlParser?

mcb5637: because forward declaring a nested struct is only possible inside the parent struct and full types are required for the state-stack.
but with the reorganization it is no longer needed.

conanfile.py
```diff
@@ -39,6 +39,7 @@ def requirements(self):
         self.requires("dice-hash/0.4.11", transitive_headers=True)
         self.requires("dice-sparse-map/0.2.9", transitive_headers=True)
         self.requires("dice-template-library/1.19.0", transitive_headers=True)
+        self.requires("libxml2/2.15.0", options={"iconv": False})
```
liss-h, line 42: is there a difference between specifiying it here and in the `configure()` step?

src/rdf4cpp/parser/IStreamQuadIterator.hpp
```diff
@@ -70,11 +79,34 @@ struct IStreamQuadIterator {
...
+        [[nodiscard]] virtual uint64_t current_column() const noexcept = 0;
+
+        Impl() = default;
+        Impl(Impl const &) = delete;
+        Impl(Impl&&) = delete;
+        Impl &operator=(Impl const &) = delete;
+        Impl &operator=(Impl &&) = delete;
+    };
+    struct ImplSerd;
+    struct ImplXMLStateCollector;
```
liss-h: Can be removed

private/rdf4cpp/parser/XMLParserUtility.cpp
```diff
@@ -0,0 +1,167 @@
...
+    std::string_view XMLOutputQueue::current_base_iri() const {
+        return state_->iri_factory.get_base();
+    }
+
+    void XMLOutputQueue::add_error(ParsingError::Type ty, std::string msg, XMLStateInfo const &i) {
+        result_queue_.emplace_back(nonstd::unexpect, ty, i.line, i.column, std::move(msg));
+    }
+
+    void XMLOutputQueue::add_statement(Node subject, IRI predicate, Node object, IRI reify) {
+        if (subject.null() || predicate.null() || object.null()) {
```
liss-h, line 44: When does this happen?

private/rdf4cpp/parser/XMLParser.hpp
```diff
@@ -0,0 +1,101 @@
...
+        ImplXML &operator=(ImplXML &&) = delete;
+        ~ImplXML() override = default;
+
+        [[nodiscard]] std::optional<value_type> next() override;
+
+        [[nodiscard]] uint64_t current_line() const noexcept override;
+        [[nodiscard]] uint64_t current_column() const noexcept override;
+    };
+
+    struct StateTransition {
```
liss-h: Move to separate file

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,150 @@
+#include <rdf4cpp/parser/XMLParser.hpp>
+
+#include <ranges>
+
+namespace rdf4cpp::parser {
+    xmlSAXHandler IStreamQuadIterator::ImplXML::make_sax_handler() {
+        xmlSAXHandler r{};
+        std::memset(&r, 0, sizeof(xmlSAXHandler));
+        r.initialized = XML_SAX2_MAGIC;
+        r.getParameterEntity = [](void *, xmlChar const *e) {
```
liss-h: please move all these functions into static functions of the parser (like with on_error).

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,150 @@
...
+            std::string_view const v = s.get().base;
+            if (!v.empty()) {
+                base = v;
+                break;
+            }
+        }
+
+        std::string_view lang_tag = "";
+        for (auto const &s : state_stack_ | std::ranges::views::reverse) {
+            std::string_view const v = s.get().lang_tag;
```
liss-h: `s->lang_tag` (no need to use `get()`. this applies to a bunch of places)

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,150 @@
...
+        for (auto const &s : state_stack_ | std::ranges::views::reverse) {
+            std::string_view const v = s.get().base;
+            if (!v.empty()) {
+                base = v;
+                break;
+            }
+        }
+
+        std::string_view lang_tag = "";
+        for (auto const &s : state_stack_ | std::ranges::views::reverse) {
```
liss-h: `std::views::reverse`

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        xmlChar const *data;
+        int size = 1024;
+        int off = 0;
+        xmlCtxtGetInputWindow(context_.get(), 0, &data, &size, &off);
+        std::string_view const source{reinterpret_cast<char const *>(data), static_cast<size_t>(size)};
+
+        return XMLStateInfo{
+                static_cast<uint64_t>(xmlSAX2GetLineNumber(context_.get())),
```
liss-h: these are the same as `current_line()` and `current_column()`

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,150 @@
...
+          output_(state) {
+        xmlCtxtSetOptions(context_.get(), XML_PARSE_NOENT | XML_PARSE_PEDANTIC | XML_PARSE_NOCDATA | XML_PARSE_NO_XXE | XML_PARSE_BIG_LINES);
+        state_stack_.reserve(10);
+        state_stack_.emplace_back(std::in_place_type_t<xml_states::InitialState>{});
+
+        current_state().base = output_.current_base_iri();
+    }
+
+    std::optional<IStreamQuadIterator::value_type> IStreamQuadIterator::ImplXML::next() {
+        std::array<char, 1024> buffer;  // NOLINT(*-pro-type-member-init)
```
liss-h: a larger buffer of `4096` or `8192` is probably faster.

private/rdf4cpp/parser/XMLStates/XMLParserDescriptionState.hpp
```diff
@@ -0,0 +1,36 @@
...
+        void move_to(BaseState *b) noexcept override;
+
+        Node subject;
+        size_t list_current = 1;
+
+        explicit DescriptionState(InheritedAttributeInfo const &i, Node sub)
+            : BaseState(i), subject(sub) {
+        }
+
+        static std::pair<StateTransition, Node> enter(XMLOutputQueue &out, std::string_view local_name, std::string_view uri, std::span<XMLAttribute> attributes, XMLStateInfo const &info);
```
liss-h, line 40: Document return value

private/rdf4cpp/parser/XMLStates/XMLParserDescriptionState.cpp
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+        if (sub.has_value() && (parse_collection || parse_literal || parse_resource)) {
+            out.add_error(ParsingError::Type::BadSyntax, "expected only one of rdf:parseType, rdf:nodeID and rdf:resource", info);
+        }
+        if (datatype.has_value()) {
+            return StateTransition(std::in_place_type_t<TypedLiteralPredicateState>{}, inherited_attribute_info, subject, predicate, reify, *datatype);
+        } else if (sub.has_value()) {
+            out.add_statement(subject, predicate, *sub, reify);
+            return StateTransition(std::in_place_type_t<EmptyElement>{});
```
liss-h: This confuses me. Why do we expect an empty element here? Just an example xml would be enough

private/rdf4cpp/parser/XMLStates/XMLParserPredicateState.hpp
```diff
@@ -0,0 +1,40 @@
+#ifndef RDF4CPP_XMLPARSERPREDICATESTATE_H
+#define RDF4CPP_XMLPARSERPREDICATESTATE_H
+
+#include <rdf4cpp/parser/XMLParserUtility.hpp>
+#include <rdf4cpp/parser/XMLStates/XMLParserBaseState.hpp>
+
+namespace rdf4cpp::parser::xml_states {
+    /**
+     * state for https://www.w3.org/TR/rdf11-xml/#resourcePropertyElt (nested nodeElement / DescriptionState
```
liss-h, line 9: For these descriptions it would be helpful if you provided a small xml example and show where we are (the spec is hard to read)

liss-h (changes requested):
Thank you for the changes. They really make the code much easier to read. Here is another review, I expect that I will write another (probably final one) after these comments are addressed.

This one is mostly a request for some more example driven documentation and some smaller issues.

conanfile.py
mcb5637: As far as i can tell, no difference

private/rdf4cpp/parser/XMLParserUtility.cpp
mcb5637: see https://github.com/tentris/rdf4cpp/pull/410/changes#diff-9428cc08adad5531e7b7222293200ad746a483a925f035f377784a0d9773a08bR106-R109

private/rdf4cpp/parser/XMLParserStateTransition.hpp
```diff
@@ -0,0 +1,39 @@
...
+#define RDF4CPP_XMLPARSERSTATETRANSITION_H
+
+#include <rdf4cpp/parser/XMLParserUtility.hpp>
+
+#include <rdf4cpp/parser/XMLStates/XMLParserBaseState.hpp>
+
+#include <rdf4cpp/parser/XMLStates/XMLParserRDFState.hpp>
+
+#include <rdf4cpp/parser/XMLStates/XMLParserDescriptionState.hpp>
+
```
liss-h: many blank lines here

private/rdf4cpp/parser/XMLStates/XMLParserCollectionState.cpp
```diff
@@ -0,0 +1,40 @@
...
+namespace rdf4cpp::parser::xml_states {
+    StateTransition CollectionState::on_characters(XMLOutputQueue &out, std::string_view const chars, XMLStateInfo const &info) {
+        if (!trim_left(chars).empty()) {
+            out.add_error(ParsingError::Type::BadSyntax, "expected element, found characters", info);
+        }
+        return {};
+    }
+
+    StateTransition CollectionState::on_start_element(XMLOutputQueue &out, std::string_view const local_name, std::string_view const uri, std::span<XMLAttribute> const attributes, XMLStateInfo const &info) {
+        auto [transition, obj] = DescriptionState::enter(out, local_name, uri, attributes, info);
```
liss-h, line 16: What happens if I put something in the collection that is not a Description?

private/rdf4cpp/parser/XMLParser.cpp
```diff
@@ -0,0 +1,154 @@
...
+        });
+    }
+
+    void IStreamQuadIterator::ImplXML::on_error(void *th, char const *msg, ...) {  // NOLINT(*-dcl50-cpp)
+        va_list args;
+        auto t = static_cast<ImplXML *>(th);
+        va_start(args, msg);  // NOLINT(*-pro-bounds-array-to-pointer-decay)
+        std::string out{};
+        out.resize(1024, '\0');
+        auto l = vsnprintf(out.data(), out.size(), msg, args);  // NOLINT(*-pro-bounds-array-to-pointer-decay)
```
liss-h, line 88: I think the correct way to determine the buffer size is to call `vsnprintf` with a null argument first. Although I am not sure how the va_args need to be handled off the top of my head.

private/rdf4cpp/parser/XMLStates/XMLParserXMLLiteralState.hpp
```diff
@@ -0,0 +1,36 @@
+#ifndef RDF4CPP_XMLPARSERXMLLITERALESTATE_H
+#define RDF4CPP_XMLPARSERXMLLITERALESTATE_H
+
+#include <rdf4cpp/parser/XMLParserUtility.hpp>
+#include <rdf4cpp/parser/XMLStates/XMLParserPredicateState.hpp>
+
+namespace rdf4cpp::parser::xml_states {
+    /**
+     * state for https://www.w3.org/TR/rdf11-xml/#parseTypeLiteralPropertyElt
```
liss-h, line 9: add a note that we do not implement this spec compliant

private/rdf4cpp/parser/XMLStates/XMLParserXMLLiteralState.cpp
```diff
@@ -0,0 +1,53 @@
...
+        if (depth > 0) {
+            --depth;
+            source_input(info);
+            return {};
+        }
+        IRI datatype = out.make_hardcoded_iri("http://www.w3.org/1999/02/22-rdf-syntax-ns#XMLLiteral");
+        std::string_view l = literal;
+        l = l.substr(0, last_offset);
+        l.remove_prefix(data_start);
+        if (!l.empty() && l[0] == '/') {
```
liss-h, line 30: please add a comment to describe what this logic is for

private/rdf4cpp/parser/XMLStates/XMLParserInitialState.cpp
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    StateTransition InitialState::on_start_element(XMLOutputQueue &out, std::string_view const local_name, std::string_view const uri, [[maybe_unused]] std::span<XMLAttribute> attributes, XMLStateInfo const &info) {
+        if (iri_equal_pieces(RDFState::start_element, uri, local_name)) {
+            return StateTransition{
+                    std::in_place_type_t<RDFState>{},
+                    get_inherited_attributes(out, attributes, info),
+            };
+        }
+        out.add_error(ParsingError::Type::BadSyntax, "expected RDF, found ???", info);
```
liss-h: what's with the three question marks?

private/rdf4cpp/parser/XMLStates/XMLParserDescriptionState.cpp
```diff
@@ -0,0 +1,145 @@
...
+            } else if (iri_equal_pieces(node_id_attrib, att.uri(), att.local_name())) {
+                sub = out.make_bn(att.value(), info);
+            } else if (iri_equal_pieces(id_attrib, att.uri(), att.local_name())) {
+                reify = out.make_id(att.value(), inherited_attribute_info.base, info);
+            } else if (iri_equal_pieces(PredicateState::parse_type_attrib, att.uri(), att.local_name())) {
+                if (att.value() == PredicateState::parse_type_resource) {
+                    parse_resource = true;
+                } else if (att.value() == PredicateState::parse_type_collection) {
+                    parse_collection = true;
+                } else {
```
liss-h: this should probably handle the case where it does not match with anything we know of

private/rdf4cpp/parser/XMLStates/XMLParserDescriptionState.cpp
```diff
@@ -0,0 +1,145 @@
...
+                    parse_literal = true;
+                }
+            }
+        }
+        for (auto const &att : attributes) {
```
liss-h, line 70: Do we have to loop twice? If yes document why

liss-h (changes requested):
Thank you for all the changes. While this is still complex (because rdf-xml just is complex), I think it is very well structured.

I am just missing a few comments here an there to explain things.

Additionally, you use `std::in_place_type_t<XYZ>{}` very often, please just use `std::in_place_type<XYZ>` (without _t and curly braces).

liss-h:
@nkaralis Do you still have the time to implement the tests for this or should the task be delegated?

nkaralis:
I will implement them.

nkaralis:
@mcb5637 Have a look at the commented out tests. In the negative tests, there are multiple that are failing

private/rdf4cpp/parser/XMLStates/XMLParserCollectionState.cpp
mcb5637: see added comment

liss-h (approved):
Thanks for all the work you put in @mcb5637. I think this is ready to be merged.

PR #411, opened 2025-11-11
Use type_list from dice-template-library

liss-h:
Use dice-template-library type list


src/rdf4cpp/datatypes/registry/DatatypeConversion.hpp
```diff
@@ -340,38 +340,38 @@ consteval ConversionTable auto make_conversion_table() {
...
     if constexpr (!SubtypedLiteralDatatype<Type>) {
-        return mz::type_list<level_0_table>{};
+        return util::type_list<Level0Table>{};
     } else {
         using s_table = decltype(make_subtype_layer<Type>());
 
         // generate the linear promotion table for each supertype
-        auto const other_level_tables = util::type_list_map<s_table>([]<ConversionEntry ToSuper>() noexcept {
-            using level_p_table = decltype(make_promotion_layer<typename ToSuper::target_type>());
+        using OtherLevelTables = util::transform_t<s_table, []<ConversionEntry ToSuper>(std::type_identity<ToSuper>) {
```
nkaralis, line 360: why do we need to add `std::type_identity<ToSuper>` as an argument?

src/rdf4cpp/datatypes/registry/DatatypeConversion.hpp
```diff
@@ -381,13 +381,13 @@ consteval ConversionTable auto make_conversion_table() {
                     }
                 };
 
-                return PromotedSuperConversion{};
-            });
+                return std::type_identity<PromotedSuperConversion>{};
```
nkaralis, line 384: why do we need to use `std::type_identity`?

nkaralis (commented):
Thank you for the PR. The operations using the new `type_list` match the previous ones. I have only two questions.

src/rdf4cpp/datatypes/registry/DatatypeConversion.hpp
liss-h: yes

src/rdf4cpp/datatypes/registry/DatatypeConversion.hpp
liss-h: yes

src/rdf4cpp/datatypes/registry/DatatypeConversion.hpp
liss-h: Because that is how the function is called. It is called with a `type_identity` parameter

src/rdf4cpp/datatypes/registry/DatatypeConversion.hpp
liss-h: because the return type is expected to have a `::type` member

nkaralis (approved):

PR #412, opened 2025-11-13
Fix: make random bnode id generator xml compliant

liss-h:
XML is more strict than turtle for bnode labels. Speficially, no numbers at the front.


nkaralis (approved):
Thank you!

PR #415, opened 2025-11-20
Feature: graph and dataset anonymizer

liss-h:
Add a dataset and graph anonymizer that anonymizes all content but keeps the graph structure intact.

Additionally, fixes things with the iterators of Graph and Dataset so that Graph and Dataset are actually ranges.

### DIFF
--- a/src/rdf4cpp/Dataset.cpp
+++ b/src/rdf4cpp/Dataset.cpp
@@ -1,4 +1,4 @@
-#include "Dataset.hpp"
+#include <rdf4cpp/Dataset.hpp>
 #include <rdf4cpp/Graph.hpp>
 #include <rdf4cpp/writer/TryWrite.hpp>
 #include <rdf4cpp/writer/SerializationState.hpp>
@@ -181,6 +181,10 @@ Dataset::iterator &Dataset::iterator::operator++() noexcept {
     return *this;
 }
 
+void Dataset::iterator::operator++(int) noexcept {
+    ++*this;
+}
+
 Dataset::iterator::reference Dataset::iterator::operator*() const noexcept {
     return cur_;
 }
@@ -189,12 +193,12 @@ Dataset::iterator::pointer Dataset::iterator::operator->() const noexcept {
     return &cur_;
 }
 
-bool Dataset::iterator::operator==(Dataset::sentinel) const noexcept {
-    return giter_ == gend_;
+bool operator==(Dataset::iterator const &self, Dataset::sentinel) noexcept {
+    return self.giter_ == self.gend_;
 }
 
-bool Dataset::iterator::operator!=(Dataset::sentinel) const noexcept {
-    return !(*this == Dataset::sentinel{});
+bool operator==(Dataset::sentinel, Dataset::iterator const &self) noexcept {
+    return self.giter_ == self.gend_;
 }
 
 void Dataset::solution_iterator::fill_solution() noexcept {
@@ -243,6 +247,10 @@ Dataset::solution_iterator &Dataset::solution_iterator::operator++() noexcept {
     return *this;
 }
 
+void Dataset::solution_iterator::operator++(int) noexcept {
+    ++*this;
+}
+
 Dataset::solution_iterator::reference Dataset::solution_iterator::operator*() const noexcept {
     return cur_;
 }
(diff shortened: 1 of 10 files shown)


src/rdf4cpp/util/Anonymizer.hpp
```diff
@@ -0,0 +1,67 @@
...
+     * Passing in the same node yield the same anonymized node as last time, this is to ensure that
+     * the graph structure stays intact.
+     *
+     * Any node passed in gets turned into an IRI matching the following regex: <http://example.com/[a-zA-Z]{16}>
+     *      except for the null node and the default graph IRI which both stay as they are.
+     *
+     * @param non_anon non-anonymized node to anonymize
+     * @return anonymized node
+     */
+    [[nodiscard]] IRI anonymize(Node const &non_anon);
```
nkaralis, line 52: it would be better to keep the type of the Node. For example, an issue might be caused because of a Literal; mapping everything to IRIs will not allow us to find the issue. 

Regarding Literals, ideally, we need to also preserve the datatype. Is it possible to use the grammar of each datatype to create random lexical forms? If this needs a lot of effort, we can map literals to string literals for now.

BNodes do not need to be anonymized; they are already anonymous identifiers.

Extra suggestion: We could add modes of operation for allowing the user to choose if all nodes should be converted to IRIs or not.

src/rdf4cpp/util/Anonymizer.cpp
```diff
@@ -0,0 +1,102 @@
...
+struct GlobalCryptoRng {
+    using result_type = uint64_t;
+
+    constexpr GlobalCryptoRng() noexcept = default;
+
+    [[nodiscard]] result_type operator()() const {
+        std::array<unsigned char, sizeof(result_type)> buf;
+
+        int const ret = RAND_bytes(buf.data(), buf.size());
+        if (ret == 0) [[unlikely]] {
```
nkaralis: what about the case it returns `-1`? is it possible in our case?
https://docs.openssl.org/1.1.1/man3/RAND_bytes/#return-values

src/rdf4cpp/Dataset.cpp
```diff
@@ -217,6 +232,7 @@ Dataset::solution_iterator::solution_iterator(Dataset const *parent,
 
         if (pat_.graph().is_variable()) {
             iter_ = giter_->second.match(tpat).begin();
+            advance_until_result();
```
nkaralis, line 235: why is this needed here?

nkaralis (commented):
Thank you for the PR. I have some questions and comments.

src/rdf4cpp/util/Anonymizer.cpp
liss-h: I did not see that

src/rdf4cpp/Dataset.cpp
liss-h: for the case that in the first graph there is no match for that pattern

src/rdf4cpp/util/Anonymizer.hpp
liss-h: I talked with @bigerl about this, he thinks that we can do this in the future if we need it.

For now the goal of this was mostly to find issues with the hypertrie structure, the type of node does not make a difference for this.

If we want to use this for debugging of queries we probably need to go all the way, and even then it will not be a 100% solution for literals (different code paths might be taken based on literal value).

liss-h: I can implement the simple version where literals are mapped to xsd:string and bnodes are kept as bnodes

liss-h: For the BlankNodes: while they are technically anon, sometimes people give them names with a meaning. So I would still anonymize them

src/rdf4cpp/util/Anonymizer.cpp
liss-h: fixed

src/rdf4cpp/util/Anonymizer.hpp
nkaralis: Okay. Then let's stick to Node -> IRI for now.

PR #416, opened 2025-11-26
Fix: segfault with template cast, keep_node_storage and non-fixed target type

liss-h:


bigerl (approved):
Fx for the fix. Change and extension of tests make sense.

PR #417, opened 2025-11-26
Fix: constexpr string

liss-h:
Fix wrong size report of ConstexprString and improve code

### DIFF
--- a/src/rdf4cpp/datatypes/registry/util/ConstexprString.hpp
+++ b/src/rdf4cpp/datatypes/registry/util/ConstexprString.hpp
@@ -18,57 +18,63 @@ namespace rdf4cpp::datatypes::registry::util {
  */
 template<size_t N>
 struct ConstexprString {
-    std::array<char, N> value;
+    std::array<char, N> value{};
 
-    constexpr ConstexprString(const char (&str)[N]) noexcept {
+    constexpr ConstexprString() noexcept = default;
+
+    constexpr ConstexprString(char const (&str)[N]) noexcept {
         std::copy_n(str, N, value.data());
     }
+
     constexpr ConstexprString(std::string_view v) {
-        if (v.size()+1 != N)
-            throw std::invalid_argument{"size missmatch"};
-        std::copy_n(v.data(), N-1, value.data());
-        value[N-1] = 0;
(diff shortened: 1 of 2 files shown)


src/rdf4cpp/datatypes/registry/util/ConstexprString.hpp
```diff
@@ -18,57 +18,63 @@ namespace rdf4cpp::datatypes::registry::util {
...
     constexpr ConstexprString(std::string_view v) {
-        if (v.size()+1 != N)
-            throw std::invalid_argument{"size missmatch"};
-        std::copy_n(v.data(), N-1, value.data());
-        value[N-1] = 0;
+        if (v.size() != N - 1) {
+            throw std::invalid_argument{"size mismatch"};
+        }
+
+        std::copy_n(v.data(), v.size(), value.data());
```
bigerl: I would stick here with 
```suggestion
        std::copy_n(v.data(), N - 1, value.data());
```

since `v.size() == N - 1` is guranteed.

src/rdf4cpp/datatypes/registry/util/ConstexprString.hpp
```diff
@@ -18,57 +18,63 @@ namespace rdf4cpp::datatypes::registry::util {
...
 template<size_t N>
 struct ConstexprString {
-    std::array<char, N> value;
+    std::array<char, N> value{};
 
-    constexpr ConstexprString(const char (&str)[N]) noexcept {
+    constexpr ConstexprString() noexcept = default;
+
+    constexpr ConstexprString(char const (&str)[N]) noexcept {
         std::copy_n(str, N, value.data());
```
bigerl, line 29: I would throw here if not null-terminated or at least document that the argument must be null-terminated.

bigerl (changes requested):
Thx for the fix. Change looks good and much cleaner. I only have two minor comments.

src/rdf4cpp/datatypes/registry/util/ConstexprString.hpp
liss-h: added doc comment

bigerl (approved):

PR #418, opened 2025-11-26
Optimize: templated literal cast

liss-h:
Avoid unecessary round trip to node storage


bigerl (approved):
Thx for the PR. Refactoring makes sense. Was this found in profiling?

liss-h:
No I was looking at the code for some other bug fix and noticed that this is not optimal

PR #422, opened 2026-01-29
Fix: detect-POBR-diff workflow

mcb5637:


liss-h:
For this: please fix POBR diff workflow so that it works with official github runners (it should still work with our private runners)

liss-h (approved):

PR #426, opened 2026-02-12
JSON-LD Parser

### LINKED ISSUE #424: Json-LD parser
we need a json-ld parser

mcb5637:
close #424


private/rdf4cpp/parser/JsonLdParserTypes.hpp
```diff
@@ -0,0 +1,306 @@
...
+        struct NotSet {
+            constexpr auto operator<=>(NotSet const &) const = default;
+        };
+
+        enum class BaseDirection : uint8_t {
+            None,
+            Ltr,
+            Rtl,
+        };
+        std::optional<BaseDirection> try_parse_base_direction(std::string_view d);
```
nkaralis, line 92: is this function used?

private/rdf4cpp/parser/JsonLdParserTypes.hpp
```diff
@@ -0,0 +1,306 @@
...
+            }
+            [[nodiscard]] constexpr std::string_view get_search_key() const {
+                if (search_key.empty()) {
+                    return data;
+                }
+                return search_key;
+            }
+        };
+        struct TypedLiteralMapping {
+            std::string value;
```
nkaralis, line 138: are there cases where we can use `string_view` here? can we make use of `CoWString`?

private/rdf4cpp/parser/JsonLdParserTypes.hpp
```diff
@@ -0,0 +1,306 @@
...
+
+            [[nodiscard]] constexpr std::optional<std::string> output() const {
+                return std::visit([]<typename T>(T const &e) -> std::optional<std::string> {
+                    if constexpr (std::same_as<T, std::string>) {
+                        return e;
+                    } else {
+                        return std::nullopt;
+                    }
+                },
+                                  *this);
```
nkaralis: formatting here is off.  i don't think we need an extra line for `*this`

tests/parser/tests_JSON_LD_Parser.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+void jsonld_test_negative(std::string json_str, std::string_view base_iri) {
+    parse_test_helpers::parser_test_negative(std::move(json_str), base_iri, ParsingFlag::JsonLd);
+}
+
+std::string remote_test_file_to_str(std::string_view file_name) {
+    return parse_test_helpers::parser_test_remote_test_file_to_str(file_name, "https://raw.githubusercontent.com/w3c/json-ld-streaming/refs/heads/main/tests/stream-toRdf", "./jsonld_test_cache");
+}
+
+TEST_CASE("test cases from rdf-tests") {
```
nkaralis: to avoid issues with the IDE, we should probably rename this, as the same name is used in the xml tests.

private/rdf4cpp/parser/JsonLdParserTypes.hpp
```diff
@@ -0,0 +1,306 @@
...
+    static constexpr std::string_view keyword_type = "@type";
+    static constexpr std::string_view keyword_value = "@value";
+    static constexpr std::string_view keyword_version = "@version";
+    static constexpr std::string_view keyword_vocab = "@vocab";
+
+    // not treated as keyword in other places
+    static constexpr std::string_view keyword_default = "@default";
+
+    static constexpr std::string_view rdf_json_datatype = "http://www.w3.org/1999/02/22-rdf-syntax-ns#JSON";
+    static constexpr std::string_view iri_nil = "http://www.w3.org/1999/02/22-rdf-syntax-ns#nil";
```
nkaralis: these are also defined for xml. might make sense to define them somewhere else (together with other frequently used IRIs)

private/rdf4cpp/parser/JsonLdParserTypes.hpp
```diff
@@ -0,0 +1,306 @@
...
+            constexpr bool operator==(IRIMapping const &r) const noexcept {
+                return (*this <=> r) == std::strong_ordering::equal;
+            }
+
+            [[nodiscard]] constexpr bool is_keyword(std::string_view k) const {
+                return type == IRIMappingType::Keyword && data == k;
+            }
+            [[nodiscard]] constexpr std::string_view get_search_key() const {
+                if (search_key.empty()) {
+                    return data;
```
nkaralis, line 132: should this be renamed to `get_serach_key_or_data`?

private/rdf4cpp/parser/JsonLdParserTypes.hpp
```diff
@@ -0,0 +1,306 @@
...
+        };
+        std::optional<BaseDirection> try_parse_base_direction(std::string_view d);
+
+        enum class ParseState : uint8_t {
+            NotStarted,
+            InProgress,
+            Done,
+        };
+
+        enum class IRIMappingType : uint8_t {
```
nkaralis: I find it confusing that a case of this is `BlankNode`. Is `IRIMappingType` defined in the standard? Please add a docstring

private/rdf4cpp/parser/JsonLdParserTypes.hpp
```diff
@@ -0,0 +1,306 @@
...
+        struct LiteralMapping {
+            std::string value;
+            std::optional<std::string> language;
+            BaseDirection direction;
+        };
+
+        // result of the value expansion function
+        using ExpandedValue = std::variant<IRIMapping, LiteralMapping, TypedLiteralMapping, simdjson::ondemand::value>;
+
+        struct LanguageMapping : std::variant<NotSet, Null, std::string> {
```
nkaralis, line 154: please document the difference between `NotSet` and `Null`

private/rdf4cpp/parser/JsonLdParserTypes.hpp
```diff
@@ -0,0 +1,306 @@
...
+
+            using TermDefinitionBase::TermDefinitionBase;
+        };
+
+        struct Context {
+            std::vector<TermDefinition> terms{};
+            std::string base_iri;
+            std::optional<std::string> vocab = std::nullopt;
+            Context const *previous_context = nullptr;
+            LanguageMapping language = NotSet{};
```
nkaralis, line 274: is this needed in context and in every term definition?

tests/parser/tests_JSON_LD_Parser.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+    jsonld_test_positive(remote_test_file_to_str("0036-in.jsonld"), remote_test_file_to_str("0036-out.nq"), "https://w3c.github.io/json-ld-streaming/tests/t0036");
+    jsonld_test_positive(remote_test_file_to_str("0113-in.jsonld"), remote_test_file_to_str("0113-out.nq"), "https://w3c.github.io/json-ld-streaming/tests/t0113");
+    jsonld_test_positive(remote_test_file_to_str("0114-in.jsonld"), remote_test_file_to_str("0114-out.nq"), "https://w3c.github.io/json-ld-streaming/tests/t0114");
+    jsonld_test_positive(remote_test_file_to_str("0115-in.jsonld"), remote_test_file_to_str("0115-out.nq"), "https://w3c.github.io/json-ld-streaming/tests/t0115");
+    jsonld_test_positive(remote_test_file_to_str("0116-in.jsonld"), remote_test_file_to_str("0116-out.nq"), "https://w3c.github.io/json-ld-streaming/tests/t0116");
+    jsonld_test_positive(remote_test_file_to_str("0117-in.jsonld"), remote_test_file_to_str("0117-out.nq"), "https://w3c.github.io/json-ld-streaming/tests/t0117");
+    // bn as predicate
+    //jsonld_test_positive(remote_test_file_to_str("0118-in.jsonld"), remote_test_file_to_str("0118-out.nq"), "https://w3c.github.io/json-ld-streaming/tests/t0118");
```
nkaralis, line 84: JSON-LD allows blank nodes in the predicate position (https://www.w3.org/TR/json-ld11/#relationship-to-rdf)

can we skolemize them? this can also be addressed in a separate PR.

private/rdf4cpp/parser/JsonLdParserTypes.cpp
```diff
@@ -0,0 +1,114 @@
...
+            return *this;
+        }
+        std::default_sentinel_t ValueArrayIter::end() {
+            return std::default_sentinel;
+        }
+        void ValueArrayIter::push_index(json_ld::KeyPath &p) {
+            return std::visit([&]<typename T>(T &) {
+                if constexpr (std::same_as<T, simdjson::ondemand::array_iterator>) {
+                    p.keys.emplace_back(current_index_);
+                }
```
nkaralis, line 110: do we need an unreachable assertion for the other cases?

private/rdf4cpp/parser/JsonLdParserTypes.cpp
```diff
@@ -0,0 +1,114 @@
...
+            cache_ = std::visit([]<typename T>(T &t) -> simdjson::ondemand::value {
+                if constexpr (std::same_as<T, simdjson::ondemand::value>) {
+                    return t;
+                } else if constexpr (std::same_as<T, simdjson::ondemand::array_iterator>) {
+                    return *t;
+                } else {
+                    return {};
+                }
+            },
+                                current_);
```
nkaralis: please fix the formatting of the visit functions

private/rdf4cpp/parser/JsonLdParserTypes.cpp
```diff
@@ -0,0 +1,114 @@
+#include "JsonLdParserTypes.hpp"
+#include <rdf4cpp/util/CharMatcher.hpp>
+#include <uni_algo/all.h>
+
+namespace rdf4cpp::parser {
+    std::optional<json_ld::BaseDirection> json_ld::try_parse_base_direction(std::string_view d) {
```
nkaralis, line 6: please insert a space between function definitions.

nkaralis (commented):
Thank you for the PR.
This is a first review focusing mostly on ParserTypes.

private/rdf4cpp/parser/JsonLdParserTypes.hpp
mcb5637: Ops, i added it to simplify code somewhere else and then forgot to actually use it xD

private/rdf4cpp/parser/JsonLdParserTypes.hpp
mcb5637: This is what auto format is set up to do

private/rdf4cpp/parser/JsonLdParserTypes.hpp
mcb5637: data is just a fallback for the case that no search_key is defined

private/rdf4cpp/parser/JsonLdParserTypes.hpp
mcb5637: Yes, see https://json-ld.org/playground/#startTab=tab-nquads&json-ld=%7B%22%40context%22%3A%7B%22%40language%22%3A%22de%22%2C%22foo%22%3A%7B%22%40id%22%3A%22http%3A%2F%2Fexample.com%2Ffoo%22%2C%22%40language%22%3A%22en%22%7D%7D%2C%22foo%22%3A%22something%20english%22%2C%22http%3A%2F%2Fexample.com%2Fbar%22%3A%22something%20german%22%7D

tests/parser/tests_JSON_LD_Parser.cpp
mcb5637: https://www.w3.org/TR/json-ld11-api/#deserialize-json-ld-to-rdf-algorithm says it is obsolete and may be removed in the future.

private/rdf4cpp/parser/JsonLdParserTypes.cpp
mcb5637: No, added a comment explaining it

private/rdf4cpp/parser/JsonLdParserTypes.hpp
mcb5637: I re-checked iri_expansion and every use of this value requires a copy.

private/rdf4cpp/parser/JsonLdParserPath.hpp
```diff
@@ -0,0 +1,87 @@
+#ifndef RDF4CPP_JSONLDPARSERPATH_HPP
+#define RDF4CPP_JSONLDPARSERPATH_HPP
+
+#include <string>
+#include <variant>
+#include <vector>
+
+#include <simdjson.h>
+
+namespace rdf4cpp::parser {
```
nkaralis, line 10: Please rename single-letter variables in this file.

private/rdf4cpp/parser/JsonLdParserPath.hpp
```diff
@@ -0,0 +1,87 @@
...
+            } else {
+                return std::tuple(simdjson::INCORRECT_TYPE, T{});
+            }
+        }
+        simdjson::error_code c;
+        std::variant<simdjson::ondemand::array, simdjson::ondemand::object> e = o;
+        for (size_t i = 0; i < key.keys.size() - 1; ++i) {
+            auto const &next_key = key.keys[i + 1];  // NOLINT(*-pro-bounds-avoid-unchecked-container-access)
+            auto const &current_key = key.keys[i];   // NOLINT(*-pro-bounds-avoid-unchecked-container-access)
+            if (std::holds_alternative<size_t>(next_key)) {
```
nkaralis, line 56: what does `next_key` being `size_t` guarantee that we are dealing with an array? please provide an example in comments for both cases.

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,712 @@
...
+        }
+        return make_iri(m);
+    }
+    nonstd::expected<json_ld::DirectionLiteralResult, IStreamQuadIterator::error_type> IStreamQuadIterator::ImplJsonLd::make_literal(json_ld::LiteralMapping const &t, json_ld::IRIMapping const &graph) {
+        try {
+            if (t.direction != json_ld::BaseDirection::None) {
+                std::string_view dir = t.direction == json_ld::BaseDirection::Ltr ? "ltr" : "rtl";
+                if (direction_ == ParsingFlag::JsonLdDirectionI18n) {
+                    auto lang = t.language.has_value() ? static_cast<std::string_view>(*t.language) : "";
+                    auto dt = std::format("https://www.w3.org/ns/i18n#{}_{}", una::cases::to_lowercase_utf8(lang), dir);
```
nkaralis, line 41: if there is no language tag, is it valid to have `https://www.w3.org/ns/i18n#_ltr` or `https://www.w3.org/ns/i18n#_rtl` (basically `_` after `#`). if it is not valid, is it handled in `make_iri`?

private/rdf4cpp/parser/JsonLdParserTypes.hpp
```diff
@@ -0,0 +1,312 @@
...
+                    return data;
+                }
+                return search_key;
+            }
+        };
+        struct TypedLiteralMapping {
+            std::string value;
+            std::string type;
+        };
+        struct LiteralMapping {
```
nkaralis: suggestion: would it be better to name this `StringLiteralMapping` or `StringLikeLiteralMapping`?

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,712 @@
+#include "JsonLdParser.hpp"
+#include <uni_algo/case.h>
+
+namespace rdf4cpp::parser {
```
nkaralis, line 4: This file also contains many single-letter variables. Please use variable names that are easier to follow.

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,712 @@
...
+                    .obj_out = p.obj_out,
+                    .is_top_level = p.is_top_level,
+                    .is_reverse = p.is_reverse,
+                    .is_json_literal = p.is_json_literal,
+                    .is_included = p.is_included,
+                }));
+            }
+            co_return;
+        }
+        auto expanded = expand_parser_.expand_level({
```
nkaralis, line 174: is this now the second step of the algorithm?

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,712 @@
...
+            *p.obj_out = nil;
+            p.obj_out = nullptr;
+        }
+        if (curr_sub->type != json_ld::IRIMappingType::None && curr_sub->type != json_ld::IRIMappingType::None) {
+            co_yield make_quad(p.active_graph, *curr_sub, *curr_pred, nil);
+        }
+    }
+    IStreamQuadIterator::ImplJsonLd::result_generator IStreamQuadIterator::ImplJsonLd::parse() {
+        simdjson::ondemand::parser parser{};
+        auto c = parser.allocate(json_data_.size() * 5);
```
nkaralis: how is this size chosen?

private/rdf4cpp/parser/JsonLdParser.hpp
```diff
@@ -0,0 +1,95 @@
...
+    }  // namespace json_ld
+    namespace params {
+        struct ParseParams {
+            simdjson::ondemand::value element;
+            json_ld::Context const &active_ctx;  // NOLINT(*-avoid-const-or-ref-data-members)
+            std::string_view base_iri;
+            json_ld::IRIMapping const &active_graph;     // NOLINT(*-avoid-const-or-ref-data-members)
+            json_ld::IRIMapping const &active_subject;   // NOLINT(*-avoid-const-or-ref-data-members)
+            json_ld::IRIMapping const &active_property;  // NOLINT(*-avoid-const-or-ref-data-members)
+            std::variant<std::monostate, json_ld::IRIMapping, Node> *obj_out = nullptr;
```
nkaralis: why is `std::monostate` needed here?

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,712 @@
...
+            auto l = make_literal(t, p.active_graph);
+            if (!l.has_value()) {
+                co_yield nonstd::make_unexpected(l.error());
+                co_return;
+            }
+            if (p.active_subject.type != json_ld::IRIMappingType::None && p.active_property.type != json_ld::IRIMappingType::None) {
+                co_yield make_quad(p.active_graph, p.active_subject, p.active_property, l->object);
+                if (l->extra_quads.has_value()) {
+                    for (Quad e : *l->extra_quads) {
+                        if (!e.graph().null() && !e.subject().null() && !e.predicate().null() && !e.object().null()) {
```
nkaralis: replace with range

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,712 @@
...
+                    .is_json_literal = p.is_json_literal,
+                    .is_included = p.is_included,
+                }));
+            }
+            co_return;
+        }
+
+        // 4
+        if (std::holds_alternative<json_ld::LiteralMapping>(expanded)) {
+            auto &t = std::get<json_ld::LiteralMapping>(expanded);
```
nkaralis: can this be moved after the invalid checks?

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,712 @@
...
+                    }
+                }
+            }
+            if (p.obj_out != nullptr) {
+                *p.obj_out = l->object;
+            }
+            co_return;
+        }
+        if (std::holds_alternative<json_ld::TypedLiteralMapping>(expanded)) {
+            auto &t = std::get<json_ld::TypedLiteralMapping>(expanded);
```
nkaralis: same here

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,712 @@
...
+            if (p.is_reverse) {
+                co_yield nonstd::make_unexpected(json_ld::make_error(ParsingError::Type::BadSyntax, "invalid reverse property value"));
+                co_return;
+            }
+            auto l = make_literal(t);
+            if (!l.has_value()) {
+                co_yield nonstd::make_unexpected(l.error());
+                co_return;
+            }
+            if (p.active_subject.type != json_ld::IRIMappingType::None && p.active_property.type != json_ld::IRIMappingType::None) {
```
nkaralis, line 260: can you provide an example that leads to this clause not being satisfied?

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,712 @@
...
+            co_return;
+        }
+
+        if (std::holds_alternative<json_ld::ExpandedMap>(expanded)) {
+            auto &map = std::get<json_ld::ExpandedMap>(expanded);
+            auto const &ctx = map.active_context == nullptr ? p.active_ctx : *map.active_context;
+
+            simdjson::ondemand::object obj;
+            if (!map.expanded_from_no_map) {
+                obj = p.element;
```
nkaralis, line 275: what's happening here?

nkaralis (commented):
Second round of questions and comments.

private/rdf4cpp/parser/JsonLdParser.cpp
mcb5637: yes, both are valid (there is a note in https://www.w3.org/TR/json-ld11-api/#object-to-rdf-conversion explaining that)

private/rdf4cpp/parser/JsonLdParser.cpp
mcb5637: very rough estimate. due to how this parser works, most strings get parsed more than once, so they might end up more than once in this buffer. 5 seems like a reasonably save choice with enough room for error.

private/rdf4cpp/parser/JsonLdParser.hpp
mcb5637: see comment added to the type

private/rdf4cpp/parser/JsonLdParser.cpp
mcb5637: the comment 1 was there from a older version, i combined 2 algorithms into one function here, therefore no numbering

private/rdf4cpp/parser/JsonLdParser.cpp
mcb5637: https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40value%22%3A%22abc%22%2C%22%40type%22%3A%22http%3A%2F%2Ffoo%22%7D

private/rdf4cpp/parser/JsonLdParser.cpp
mcb5637: added a comment to the member

private/rdf4cpp/parser/JsonLdParser.cpp
nkaralis: i suggest to define it as a constexpr in the .hpp file

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+            return IRIMapping{std::string{*value}, IRIMappingType::Keyword};
+        }
+        // 2
+        if (looks_like_keyword(*value)) {
+            return IRIMapping{};
+        }
+        // 3
+        if (parse_ctx != nullptr) {
+            auto i = parse_ctx->active_context.try_find_term(*value);
+            if (i != nullptr && i->parse_state != ParseState::Done && i != ignore_local) {
```
nkaralis, line 813: in which case is the parse state done in this case?

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+                auto post = value->substr(i + 1);
+                if (pre == "_") {
+                    return IRIMapping{std::format("bn_n{}", post), IRIMappingType::BlankNode};
+                }
+                if (post.starts_with("//")) {
+                    return IRIMapping{std::string(*value), IRIMappingType::IRI};
+                }
+                if (parse_ctx != nullptr) {
+                    auto term = parse_ctx->active_context.try_find_term(pre);
+                    if (term != nullptr && term->parse_state != ParseState::Done) {
```
nkaralis, line 849: same here

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+            auto r = iri_factory->from_maybe_relative_as_string(*value);
+            if (r.has_value()) {
+                return IRIMapping{std::string(*r), IRIMappingType::IRI};
+            } else {
+                return nonstd::make_unexpected(json_ld::make_error(ParsingError::Type::BadIri, std::format("invalid relative iri: {}", r.error())));
+            }
+        }
+        // 9
+        // no keyword, bn or valid iri, would have passed any further check
+        return IRIMapping{std::string(""), IRIMappingType::None};
```
nkaralis, line 897: the document states for step 9: `Return value as is.`
is value always the empty string?

private/rdf4cpp/parser/JsonLdExpandParser.cpp
```diff
@@ -0,0 +1,909 @@
...
+                    return false;
+                }
+                graph = true;
+            } else if (expanded_property->is_keyword(keyword_index)) {
+                if (index) {
+                    ob.reset();
+                    return false;
+                }
+                index = true;
+            } else if (expanded_property->is_keyword(keyword_index)) {
```
nkaralis: duplicate check?

private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -0,0 +1,708 @@
...
+            }
+            current_bn = std::move(next_bn);
+            curr_sub = &current_bn;
+            curr_pred = &rest;
+        }
+        if (p.obj_out != nullptr) {
+            *p.obj_out = nil;
+            p.obj_out = nullptr;
+        }
+        if (curr_sub->type != json_ld::IRIMappingType::None && curr_sub->type != json_ld::IRIMappingType::None) {
```
nkaralis: it uses the same condition twice, right?

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+                                    .override_protected = p.override_protected,
+                                });
+                                if (rec.has_value()) {
+                                    return rec;
+                                }
+                                p.term.iri_mapping = other_term->iri_mapping;
+                                p.term.iri_mapping.data.append(std::string_view(p.term.key).substr(colon_pos + 2));
+                            } else {
+                                p.term.iri_mapping.data = p.term.key;
+                                p.term.iri_mapping.type = std::string_view(p.term.key).substr(2) == "_:" ? IRIMappingType::BlankNode : IRIMappingType::IRI;
```
nkaralis: should it use `starts_with` instead of `substr`?

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+                    auto type = iri_expansion(p.active_context, v, false, true, nullptr, &p_ctx);
+                    if (!type.has_value()) {
+                        return type.error();
+                    }
+                    if (type->type != IRIMappingType::Keyword && type->type != IRIMappingType::IRI) {
+                        return make_error(ParsingError::Type::BadSyntax, "invalid type mapping (not IRI or keyword)");
+                    }
+                    static constexpr std::array invalid = {keyword_json, keyword_none, keyword_id, keyword_vocab};
+                    if (type->type == IRIMappingType::Keyword && !std::ranges::any_of(invalid, [&](std::string_view a) {
+                            return a == v;
```
nkaralis: should this be `a == type->data`?

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+                            }
+                            if (cont != keyword_set && cont != keyword_index) {
+                                return make_error(ParsingError::Type::BadSyntax, "invalid reverse property");
+                            }
+                            p.term.container_mapping.clear();
+                            p.term.container_mapping.emplace_back(cont);
+                        }
+                    }
+
+                    // TODO double check protected
```
nkaralis, line 520: what about this?

private/rdf4cpp/parser/JsonLdContextParser.hpp
```diff
@@ -0,0 +1,60 @@
...
+        };
+    }  // namespace params
+
+    namespace json_ld {
+        struct ContextParser {
+            using error_type = ParsingError;
+            IRIFactory *iri_factory;
+
+            nonstd::expected<Context, error_type> parse_context(simdjson::ondemand::value local_context,
+                                                                Context const &active_context,
```
nkaralis: you could also create a params struct for this one. two of the optional arguments are missing, right?

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+                    }
+                }
+            }
+            return false;
+        };
+
+        if (local_context.type() == simdjson::ondemand::json_type::object) {
+            // 2 & 3
+            simdjson::ondemand::object const o = local_context.get_object();
+            auto [c, prop] = try_get_field<bool>(o, keyword_propagate);
```
nkaralis, line 203: the use of three different variables for `propagate` make this block difficult to follow. can you simplify it?

```
auto [c, parsed_prop] = try_get_field<bool>(o, keyword_propagate);
bool const actual_propagate = c == simdjson::SUCCESS ? parsed_prop : propagate;
if (!actual_propagate && result->previous_context == nullptr) {
                result->previous_context = &active_context;
            }
```
i haven't checked if this is correct, but something similar should work.

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+            if (c == simdjson::SUCCESS) {
+                p = prop;
+            }
+            if (!p && result->previous_context == nullptr) {
+                result->previous_context = &active_context;
+            }
+            //error handling later
+
+            handle_ctx(o);  // 4
+        } else if (local_context.is_scalar() && local_context.is_null()) {
```
nkaralis: is it possible for this to be both a scalar and null?

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+            for (auto const &t : active_context.terms) {
```
nkaralis: which step of the algorithm is this? please provide the numbers if possible

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+                if (t.is_protected) {
+                    result = nonstd::unexpected{make_error(ParsingError::Type::BadSyntax, "invalid context nullification")};
+                    return result;
+                }
+            }
+            result = Context{
+                .base_iri{base_iri},
```
nkaralis: what about "original base URL". why is not tracked?

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -0,0 +1,906 @@
...
+                result->previous_context = &active_context;
+            }
+            simdjson::ondemand::array a{};
+            if (local_context.get(a) != simdjson::SUCCESS) {
+                result = nonstd::unexpected{make_error(ParsingError::Type::BadSyntax, "invalid local context")};
+                return result;
+            }
+            for (auto v : a) {
+                switch (v.type()) {
+                    case simdjson::ondemand::json_type::null:  // 5.1
```
nkaralis, line 224: duplicate code (same code in the `else if` clause above). are both needed?

nkaralis (commented):
Some more comments, mostly on ContextParser (`parse_context_term` is the only one left here).

private/rdf4cpp/parser/JsonLdContextParser.cpp
mcb5637: Done => the term was parsed successfully.
This happens if one term in a context refers to another term in the context and the refered to term was parsed first.
I added comments to the enum type to explain what they represent.

private/rdf4cpp/parser/JsonLdContextParser.cpp
mcb5637: clarified in a comment

private/rdf4cpp/parser/JsonLdContextParser.cpp
mcb5637: I think i read the spec correctly and it says dont check protected here. But every implementation i know of does. Please double check if i am correct.

private/rdf4cpp/parser/JsonLdContextParser.hpp
mcb5637: remote contexts are not supported by us and validate scoped contexts is never set to anything other that default, therefore i removed both.

private/rdf4cpp/parser/JsonLdContextParser.cpp
mcb5637: yes, in simdjson null is a scalar. (the `is_scalar` check is there because the root parse object cant be checked with `is_null`, but `is_scalar` will return false)

private/rdf4cpp/parser/JsonLdContextParser.cpp
mcb5637: Both are needed, one is inside an array, the other in place of that array.

liss-h:
As discussed with @nkaralis and @bigerl, we will merge this now and create a follow up PR to fix potential issues when they come up.

PR #428, opened 2026-03-01
Feature/content type guessing

bigerl:
Continuing with my experiments from last week, I've tried to add the content type guessing directly into rdf4cpp. 

PR implements content guessing using file name, extensions and content sniffing.


bigerl:
CI fails with unable to locate `libstdc++-14-dev`, seems unrelated to the PR.

PR #429, opened 2026-03-16
NTriple parser fails on relative IRI

bigerl:
- [ ] double check the specs for correct behavior
- [ ] fix


bigerl:
It's not a bug:
[Ntriple does not support relative IRIs](https://www.w3.org/TR/n-triples/#sec-iri). Only [Turtle does](https://www.w3.org/TR/turtle/#sec-iri). It also comes with a default [base IRI](https://www.w3.org/TR/turtle/#sec-iri-references).

PR #430, opened 2026-04-28
Feature/pcre2

### LINKED ISSUE #400: Find alternative for `re2`
Currently, we cannot update the `re2` dependency because newer versions of it drag in abseil which we probably don't really want.

mcb5637:
close #400 

for reference:
- pcre2 doc: https://pcre2project.github.io/pcre2/doc/pcre2api/
- xml spec: https://www.w3.org/TR/xmlschema-2/#regexs
- xmlpath spec: https://www.w3.org/TR/xpath-functions/#regex-syntax (extends xml spec)

Note: coverage test is broken due to outdated compiler, fix is in #426

### DIFF
--- a/private/rdf4cpp/regex/RegexImpl.cpp
+++ b/private/rdf4cpp/regex/RegexImpl.cpp
@@ -27,6 +27,12 @@ namespace rdf4cpp::regex {
         }();
         return *match_ctx;
     }
+    std::string Regex::Impl::translate_error_code(int error_code) {
+        std::string msg;
+        msg.resize(120); // https://pcre2project.github.io/pcre2/doc/pcre2api/#geterrormessage says 120 chars is enough for any error message
+        msg.resize(pcre2_get_error_message_8(error_code, reinterpret_cast<PCRE2_UCHAR8 *>(msg.data()), msg.size()));
+        return msg;
+    }
 
     TriBool Regex::Impl::apply(pcre2_code_8 &c, std::string_view str) noexcept {
         assert(una::is_valid_utf8(str));
@@ -69,18 +75,12 @@ namespace rdf4cpp::regex {
         }
         code_ptr r{pcre2_compile_8(reinterpret_cast<PCRE2_SPTR8>(regex.data()), regex.size(), f, &error_code, &err_off, compile_ctx.get())};
         if (r == nullptr) {
-            std::string msg;
-            msg.resize(120);
-            msg.resize(pcre2_get_error_message_8(error_code, reinterpret_cast<PCRE2_UCHAR8 *>(msg.data()), msg.size()));
-            throw RegexError{"Failed to compile regex: " + msg};
+            throw RegexError{"Failed to compile regex: " + translate_error_code(error_code)};
         }
         if (flags.contains(RegexFlag::Optimize)) {
             error_code = pcre2_jit_compile_8(r.get(), 0);
             if (error_code != 0) {
-                std::string msg;
-                msg.resize(120);
-                msg.resize(pcre2_get_error_message_8(error_code, reinterpret_cast<PCRE2_UCHAR8 *>(msg.data()), msg.size()));
-                throw RegexError{"Failed to compile jit regex: " + msg};
+                throw RegexError{"Failed to compile jit regex: " + translate_error_code(error_code)};
             }
         }
         return r;
--- a/private/rdf4cpp/regex/RegexImpl.cpp
+++ b/private/rdf4cpp/regex/RegexImpl.cpp
@@ -114,10 +114,10 @@ namespace rdf4cpp::regex {
         auto s = make_code(regex, flags, 0);
         return {std::move(m), std::move(s), flags};
     }
-    Regex::Impl::Impl(code_ptr m, code_ptr s, flag_type f)
-        : match(std::move(m)),
-          search(std::move(s)),
-          flags(f) {
+    Regex::Impl::Impl(code_ptr match, code_ptr search, flag_type flags)
+        : match(std::move(match)),
+          search(std::move(search)),
+          flags(flags) {
     }
 
 }  //namespace rdf4cpp::regex
--- a/private/rdf4cpp/regex/RegexImpl.cpp
+++ b/private/rdf4cpp/regex/RegexImpl.cpp
@@ -6,17 +6,21 @@ namespace rdf4cpp::regex {
         : Impl{make(regex, flags)} {
     }
 
-    bool Regex::Impl::regex_match(std::string_view const str) const noexcept {
+    TriBool Regex::Impl::regex_match(std::string_view const str) const noexcept {
         return apply(*match, str);
     }
 
-    bool Regex::Impl::regex_search(std::string_view const str) const noexcept {
+    TriBool Regex::Impl::regex_search(std::string_view const str) const noexcept {
         return apply(*search, str);
     }
 
-    bool Regex::Impl::apply(pcre2_code_8 &c, std::string_view str) noexcept {
+    TriBool Regex::Impl::apply(pcre2_code_8 &c, std::string_view str) noexcept {
         match_data_ptr const m{pcre2_match_data_create_from_pattern_8(&c, nullptr)};
-        return pcre2_match_8(&c, reinterpret_cast<PCRE2_SPTR8>(str.data()), str.size(), 0, 0, m.get(), nullptr) >= 0;
+        auto ec =  pcre2_match_8(&c, reinterpret_cast<PCRE2_SPTR8>(str.data()), str.size(), 0, 0, m.get(), nullptr);
+        if (ec == PCRE2_ERROR_NOMATCH) {
+            return TriBool::False;
+        }
+        return ec >= 0 ? TriBool::True : TriBool::Err;
     }
 
     Regex::Impl::code_ptr Regex::Impl::make_code(std::string_view regex, flag_type flags, int extra_flags) {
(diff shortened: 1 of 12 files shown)


private/rdf4cpp/regex/RegexReplacerImpl.cpp
```diff
@@ -1,95 +1,72 @@
...
-            size_t const end = pos;
-            size_t before_start = end - 1;
-            while (before_start != std::string_view::npos && res[before_start] == '\\') {
-                --before_start;
+    std::string RegexReplacer::Impl::translate_rewrite(std::string_view const s) {
+        std::string res{s};
+
+        auto pos = res.find_first_of("$\\");
+        while (pos < res.size()) {
+            if (res[pos] == '\\') {  // NOLINT(*-pro-bounds-avoid-unchecked-container-access)
```
liss-h: I disabled the clang-tidy check globally. It is useless anyway. You can remove the NOLINT comments regarding this

tests/util/tests_Regex.cpp
```diff
@@ -5,12 +5,216 @@
...
+        CHECK(Regex{"[^abc]", f}.regex_match("a") == rdf4cpp::TriBool::False);
+        CHECK(Regex{"[^abc]", f}.regex_match("d") == rdf4cpp::TriBool::True);
+        CHECK(Regex{"[a-c]", f}.regex_match("a") == rdf4cpp::TriBool::True);
+        CHECK(Regex{"[a-c]", f}.regex_match("b") == rdf4cpp::TriBool::True);
+        CHECK(Regex{"[a-c]", f}.regex_match("c") == rdf4cpp::TriBool::True);
+        CHECK(Regex{"[a-c]", f}.regex_match("d") == rdf4cpp::TriBool::False);
+        CHECK(Regex{"[^a-c]", f}.regex_match("a") == rdf4cpp::TriBool::False);
+        CHECK(Regex{"[^a-c]", f}.regex_match("d") == rdf4cpp::TriBool::True);
+        // CHECK(!Regex{"[^abc-[e]]", f}.regex_match("e"));
+        // CHECK(Regex{"[^abc-[e]]", f}.regex_match("d"));
```
liss-h, line 106: why are these commented out? (some more below)

src/rdf4cpp/regex/RegexFlags.hpp
```diff
@@ -12,6 +12,7 @@ enum struct RegexFlag : uint8_t {
     Literal = 1 << 2,
     Multiline = 1 << 3,
     RemoveWhitespace = 1 << 4,
+    Optimize = 1 << 5,
```
liss-h, line 16: Please add a comment what exactly this does

private/rdf4cpp/regex/RegexImpl.hpp
```diff
@@ -3,19 +3,39 @@
...
+        [[nodiscard]] static pcre2_match_context_8& get_match_context();
 
+    private:
+        using match_data_ptr = std::unique_ptr<pcre2_match_data_8, CallFree<pcre2_match_data_free_8>>;
+        [[nodiscard]] static TriBool apply(pcre2_code_8 &c, std::string_view str) noexcept;
+
+        static code_ptr make_code(std::string_view regex, flag_type flags, int extra_flags);
+        static std::string remove_whitespace(std::string_view str);
+        static Impl make(std::string_view regex, flag_type flags);
+        Impl(code_ptr m, code_ptr s, flag_type f);
```
liss-h: the params need better names

private/rdf4cpp/regex/RegexImpl.hpp
```diff
@@ -3,19 +3,39 @@
...
+        struct CallFree {
+            void operator()(auto* c) {
+                f(c);
+            }
+        };
 
-struct Regex::Impl {
-    re2::RE2 regex;
-    Regex::flag_type flags;
+        using code_ptr = std::unique_ptr<pcre2_code_8, CallFree<pcre2_code_free_8>>;
```
liss-h, line 21: why are these things public?

private/rdf4cpp/regex/RegexImpl.cpp
```diff
@@ -1,72 +1,123 @@
...
+            f |= PCRE2_UCP | PCRE2_NEVER_BACKSLASH_C;
+        }
+        if (flags.contains(RegexFlag::Multiline)) {
+            f |= PCRE2_MULTILINE;
+        }
+        code_ptr r{pcre2_compile_8(reinterpret_cast<PCRE2_SPTR8>(regex.data()), regex.size(), f, &error_code, &err_off, compile_ctx.get())};
+        if (r == nullptr) {
+            std::string msg;
+            msg.resize(120);
+            msg.resize(pcre2_get_error_message_8(error_code, reinterpret_cast<PCRE2_UCHAR8 *>(msg.data()), msg.size()));
```
liss-h: why resize twice in a row?

private/rdf4cpp/regex/RegexImpl.cpp
```diff
@@ -1,72 +1,123 @@
 #include "RegexImpl.hpp"
 
+#include <uni_algo/conv.h>
+
 namespace rdf4cpp::regex {
 
-namespace detail {
+    Regex::Impl::Impl(std::string_view const regex, flag_type const flags)
+        : Impl{make(regex, flags)} {
```
liss-h, line 8: Why is the code for make not just in this ctor?

private/rdf4cpp/regex/RegexReplacerImpl.cpp
```diff
@@ -1,95 +1,72 @@
...
+                if (res.size() <= pos) {
```
liss-h: I think this is probably wrong. We are in branches: `pos < res.size() && res[pos] == '\\'`, pos and res were not changed, so pos cannot be >= res.size()

private/rdf4cpp/regex/RegexReplacerImpl.cpp
```diff
@@ -1,95 +1,72 @@
...
+                    throw RegexError{"invalid escape sequence in replacement string"};
+                }
+                if (res.at(pos + 1) == '$') {
```
liss-h: why at? same below

private/rdf4cpp/regex/RegexReplacerImpl.cpp
```diff
@@ -1,95 +1,72 @@
...
-        if (pos == 0) {
-            // is at start, cannot be escaped => replace
-            replace(res, 0);
-        } else {
-            // find start of potential backslash sequence preceding the '$'
-            size_t const end = pos;
-            size_t before_start = end - 1;
-            while (before_start != std::string_view::npos && res[before_start] == '\\') {
-                --before_start;
+    std::string RegexReplacer::Impl::translate_rewrite(std::string_view const s) {
```
liss-h: Why did you turn translate_review into a static-member?

private/rdf4cpp/regex/RegexReplacerImpl.cpp
```diff
@@ -1,95 +1,72 @@
...
+        if (regex->flags.contains(RegexFlag::Literal)) {
+            opt |= PCRE2_SUBSTITUTE_LITERAL;
+        }
+        auto rep = [&] {
+            outsize = r.size();
+            return pcre2_substitute_8(regex->search.get(), reinterpret_cast<PCRE2_SPTR8>(str.data()), str.size(), 0, opt, nullptr, &Regex::Impl::get_match_context(), reinterpret_cast<PCRE2_SPTR8>(rewrite.data()), rewrite.size(), reinterpret_cast<PCRE2_UCHAR8 *>(r.data()), &outsize);
+        };
+        auto e = rep();
+        if (e == PCRE2_ERROR_NOMEMORY) {
+            r.resize(outsize + 1);
```
liss-h, line 60: why do we need to increase size by one? Should this be a while loop?

liss-h (changes requested):
Thank you for the PR. Here is the first round of comments

tests/util/tests_Regex.cpp
mcb5637: These syntax features are not supported by PCRE2 (RE2 also did not support them either)
The only way to fix this, would be writing an own regex library.

private/rdf4cpp/regex/RegexImpl.cpp
mcb5637: To work around the c api. it requires valid initialized memory, and returns how much space it actually used
(documentation says, 120 chars is enough space for any error message).
So the first resize initializes the memory, the second cuts of any unused memory (without reallocating anything).

private/rdf4cpp/regex/RegexImpl.cpp
mcb5637: because make possibly rewrites the input string and then initializes both `code_ptr` variables.
the other options are:
- initialize both to nullptr and then override them in the ctor body
- rewrite the string 2 times

i chose this option because it looks best (but that is just my optinion)

private/rdf4cpp/regex/RegexReplacerImpl.cpp
mcb5637: `PCRE2_SUBSTITUTE_OVERFLOW_LENGTH` tells `pcre2_substitute_8` to write the length that would be required to `outsize` (excluding the terminating `\0`, thats where the +1 comes from)

private/rdf4cpp/regex/RegexImpl.cpp
liss-h: ok, just add a comment that 120 comes from the documentation

private/rdf4cpp/regex/RegexReplacerImpl.cpp
liss-h: ok, please add a comment explaining that

liss-h (approved):
Thx

PR #432, opened 2026-07-15
Pin all reusable workflows/actions.

adrian-tentris:


.github/workflows/detect-pobr-diff.yml
```diff
@@ -25,5 +25,8 @@ jobs:
       abi-version-header: src/rdf4cpp/version.hpp
       abi-version-const: rdf4cpp::pobr_version
     secrets:
+      CONAN_CICD_USERNAME: ${{ secrets.CONAN_CICD_USERNAME }}
+      CONAN_CICD_PASSWORD: ${{ secrets.CONAN_CICD_PASSWORD }}
+      CONAN_CICD_ENDPOINT: ${{ secrets.CONAN_CICD_ENDPOINT }}
```
nkaralis, line 30: do we need the old secrets or can they be removed?

adrian-tentris: Once I switch to v2, I need both to make next PR green without adding them.

PR #448, opened 2026-07-22
Fix: regex replace of unknown match group

liss-h:
fixes a regression from #430


liss-h:
merging because CI passes and nobody is available to review

PR #460, opened 2026-07-27
Remove `noexcept` specifications on node storage functions

### LINKED ISSUE #405: Remove `noexcept` requirements on node storage functions
Remove requirements for `noexcept` on node storage functions.
This would allow us to have an "unreachable" node storage that always throws an exception when it is used (e.g. as the default node storage).

Currently we have to use `std::abort()` which is not ideal.

Removing the noexcepts may require some digging around in the library to find functions that assume that they are noexcept.

SandroMiudo:
Removal of noexcept in node storage functions.


liss-h:
Hello, can you ask claude to check that all codepaths are covered. I have the feeling that they are a few stray noexcepts in the higher level functions

bigerl (commented):
# Review

## What was checked

- Built the branch in the shared dev container (clang 21, RelWithDebInfo, Ninja): clean, 205/205 targets.
- Full test suite: 52/52 pass.
- Compiled and ran a small probe program against the built `librdf4cpp.a`:
  - `static_assert(NodeStorage<UnreachableNodeStorage>)` holds, `NodeStorageVTable::get<UnreachableNodeStorage>()` and `DynNodeStoragePtr{uns}` compile.
  - `IRI::find("http://example.org/x", uns)` now propagates `std::logic_error` and is catchable. This is the intended behavior of the PR and it works at this entry point.
  - Comparing two IRIs from a throwing storage via `Node::operator==`: `terminate called after throwing an instance of 'std::logic_error'`, SIGABRT (exit 134).
  - `Node::operator std::string()` on such a node: same, SIGABRT.

## Overview

The PR removes `noexcept` from the node storage lookup chain (vtable function pointers, `DynNodeStoragePtr::find_id`/`find_*_backend`, the four `NodeBackendHandle::*_backend()` accessors, and the public `find`/`try_get_in_node_storage` functions of `IRI`/`BlankNode`/`Literal`/`Variable`), so a storage backend may throw. It also adds `UnreachableNodeStorage`, a storage that throws `std::logic_error` from every operation.

## Main issue: the noexcept removal stops one layer too low

The removal covers the storage layer and the bottom-level `find`/`try_get` entry points, but almost none of their callers. Any exception that surfaces anywhere except a direct `find`/`try_get` call still crosses a `noexcept` boundary and calls `std::terminate` (verified at runtime, see above). Concretely:

**Cases where the PR's own change is defeated by a still-`noexcept` wrapper in the middle:**

- `Literal::find_datatype_iri` (`Literal.cpp:414`, `Literal.hpp:500`) is `noexcept` but calls the now-throwing `IRI::find`. It sits between the de-noexcept'd public `Literal::find_typed` / `find_typed_from_value` (`Literal.hpp:560`, `Literal.hpp:617`) and the storage, so for datatypes with non-fixed ids the exception those templates were supposed to propagate terminates the process instead.
- `Variable::find_named` / `Variable::find_anonymous` (`Variable.cpp:126`, `Variable.cpp:129`, `Variable.hpp:37`, `Variable.hpp:45`) are still `noexcept` and are the only public callers of the de-noexcept'd private `Variable::find`. For `Variable` the change is a no-op from the API surface.
- `Node::try_get_in_node_storage` (`Node.cpp:38`, `Node.hpp:128`) is still `noexcept` and dispatches to all four de-noexcept'd per-type overloads. The aggregate wrappers `Statement::try_get_in_node_storage` (`Statement.cpp:25`), `Quad` (`Quad.cpp:51`), `TriplePattern` (`TriplePattern.cpp:39`), `QuadPattern` (`QuadPattern.cpp:40`) are also still `noexcept`.
- `Literal::serialize` (`Literal.cpp:699`) is still `noexcept` although the PR removed `noexcept` from its inner visit lambda (`Literal.cpp:761`). That lambda change has no effect while the enclosing function is `noexcept`.

**Still-`noexcept` read paths that now call throwing operations:**

- Comparison core: `Node::eq_impl` (`Node.cpp:119`), `Node::order` (`Node.cpp:174`), `Node::compare_impl` (`Node.cpp:154`), and through them `operator==`, `operator<=>`, and the whole `eq`/`lt`/`order_*` family. Also `BlankNode::order` (`BlankNode.cpp:228`) and `Variable::order` (`Variable.cpp:249`).
- Serialization: `Node::serialize` (`Node.cpp:58`), `Node::operator std::string` (`Node.cpp:82`), `IRI::serialize` (`IRI.cpp:117`), `BlankNode::serialize`, `Variable::serialize` (`Variable.cpp:167`), and the quad writers `write_node`/`write_quad` (`private/rdf4cpp/writer/WriteQuad.hpp:11`, `:45`).
- Accessors: `IRI::identifier` (`IRI.cpp:187`), `BlankNode::identifier` and `fetch_or_serialize_identifier` (`BlankNode.cpp:135`, `:146`), `Variable::fetch_or_serialize_name` (`Variable.cpp:153`), `Literal::datatype` (`Literal.cpp:469`), `Literal::value` (`Literal.cpp:845`), `Literal::language_tag` (`Literal.cpp:617`), `Literal::serialize_lexical_form_impl` (`Literal.cpp:484`) and everything built on it.
- Graph/Dataset API: `Graph::contains` (`Graph.cpp:39`), `Dataset::contains` (`Dataset.cpp:31`), `Dataset::size(IRI const&)` (`Dataset.cpp:52`). Each starts with a cross-storage `try_get_in_node_storage` call. So `graph.contains(stmt)` with a statement from another storage aborts the process if the storage throws.

Note the asymmetry inside `Variable`: `is_anonymous()` and `name()` are correctly non-`noexcept`, while `serialize()`, `fetch_or_serialize_name()` and `order()` were missed. This suggests the sweep was done partially rather than by policy.

Suggestion: decide per layer whether it may throw. Either remove `noexcept` up the call chain consistently (comparison, serialization, accessors, Graph/Dataset), or keep the upper layers `noexcept` and translate storage exceptions at a defined boundary (catch and return null/false). The current state is the worst of both: the signature says "may throw" at the bottom and "will not throw" one level up, and the process dies in between.

## UnreachableNodeStorage is not integrated

- The class is referenced nowhere: no include, no test, no CMake entry, not in the umbrella header (`src/rdf4cpp.hpp` includes only `SyncReferenceNodeStorage.hpp`). `default_node_storage` is still bound to `reference_node_storage::default_instance` (`NodeStorage.cpp:11`), and `DynNodeStoragePtr{nullptr}` still stores a null vtable (`NodeStorage.hpp:200`), so a null storage pointer still dereferences null instead of throwing. The PR description says "Integration of unreachable storage node, to integrate default behaviour", but no integration happened. If the intent is that a default-constructed `DynNodeStoragePtr` reports a clear error, point its vtable at `NodeStorageVTable::get<UnreachableNodeStorage>()` instead of `nullptr`.
- Because no translation unit includes the header, it is never compiled by the build or CI, yet it ships to users (the install rule globs `*.hpp`, `cmake/install_library.cmake`). Add `#include <rdf4cpp/storage/NodeStorage.hpp>` and `static_assert(NodeStorage<UnreachableNodeStorage>);` like both siblings (`SyncReferenceNodeStorage.hpp:77`, `UnsyncReferenceNodeStorage.hpp:83`). I verified conformance holds today, but nothing pins it.
- The four included headers (`BNodeBackend.hpp`, `FallbackLiteralBackend.hpp`, `IRIBackend.hpp`, `VariableBackend.hpp`) are not used. The file needs the four view headers plus `identifier/NodeBackendID.hpp`.
- Minor: no doc comment saying what the type is for (both siblings have one); the `#endif` comment says `RDF4CPP_UNREACHABLENODESTORAGE_H` but the guard is `RDF4CPP_STORAGE_UNREACHABLENODESTORAGE_HPP`; missing trailing newline (only header in the tree without one); `throw std::logic_error(...)` uses parentheses while the codebase brace-initializes thrown exceptions (21 existing sites use `{...}`).

## Documentation

- Typo introduced in `BlankNode.hpp:60`: "if the Fblank node identifier is not valid".
- The doc comments of `find`/`try_get_in_node_storage` in all four headers still describe only the null-return contract and gained no `@throws`. The `try_` prefix now no longer implies a non-throwing call; that should be documented.

## Tests

No tests were added or changed. Nothing exercises `UnreachableNodeStorage` or asserts that the de-noexcept'd API actually propagates exceptions. A test like

```cpp
NodeStorage auto ns = reference_node_storage::UnreachableNodeStorage{};
CHECK_THROWS_AS(IRI::find("http://example.org/x", ns), std::logic_error);
```

in `tests/nodes/tests_NodeStorage_helper_types.cpp` would both compile the new header for the first time and would have caught the still-`noexcept` wrappers above (a version of it going through `Node::try_get_in_node_storage` currently aborts instead of throwing).

## Performance and ABI

Negligible. The vtable stays a `static constexpr` object with the same 12 pointers in the same layout, `noexcept` is not part of Itanium name mangling, the `NodeStorage` concept never required `noexcept`, and `noexcept` member functions convert implicitly to the non-noexcept pointer types, so the reference storages keep compiling unchanged. Object-size delta is a few dozen bytes of exception tables per TU. Only external code that spells these function pointer types with `noexcept` would need adjustment.

## Verdict

The direction is fine and the storage-layer part is done correctly, but the change is incomplete: the exception path only survives one specific class of entry points, everything else turns a storage error into `std::terminate` (demonstrated at runtime), and the advertised default-behavior integration of `UnreachableNodeStorage` is missing entirely. Needs another pass over the call sites plus tests before merging.

(unchecked review by claude. take with caution)

liss-h:
I think you can remove the unreachable node storage

bigerl (commented):
# Re-review (state: 48342a4)

## What was checked

- Built the branch in the dev container (clang 21, RelWithDebInfo): clean. Full test suite: 52/52 pass.
- Re-ran a probe program against the built `librdf4cpp.a`, this time with a custom storage whose lookups throw `std::runtime_error`:
  - `IRI::find`: propagates, catchable. OK.
  - `Node::operator==` on foreign IRIs: propagates now (terminated before). Fixed.
  - `Statement::try_get_in_node_storage`: propagates. Fixed.
  - `Variable::find_named`: propagates. Fixed.
  - `Literal::find_typed<xsd::String>`: propagates. Fixed.
  - `Node::operator std::string`: still `std::terminate` (SIGABRT). Not fixed, see below.

## Resolved from the previous review

All items from the last round were addressed: the `try_get_in_node_storage` chain (Node, Statement, Quad, patterns), `Variable::find_named`/`find_anonymous`, `Literal::find_datatype_iri`, the comparison core, the serialize functions and accessors, `Graph::contains`/`Dataset::contains`/`Dataset::size`, `write_node`/`write_quad`, the "Fblank" typo, and `UnreachableNodeStorage` was removed cleanly (no dangling references). Good sweep overall.

## Remaining gaps

The sweep is still not complete. One class is proven at runtime, the others are verified in code.

**1. `operator std::string` still terminates (proven, SIGABRT).** All four conversions lost `noexcept` on the function, but each wraps the now-throwing `serialize` in a `noexcept` lambda passed to `StringWriter::oneshot`:

- `IRI.cpp:131`, `BlankNode.cpp:173`, `Literal.cpp:808`, `Variable.cpp:195`: `oneshot([this](auto &w) noexcept { return this->serialize(w); })`
- `BlankNode::operator std::string` is additionally still `noexcept` itself (`BlankNode.hpp:88`, `BlankNode.cpp:172`), the only one of the four.

`oneshot` does not require a `noexcept` callable (`writer/BufWriter.hpp:223`), so the fix is removing `noexcept` from the four lambdas plus the BlankNode operator.

**2. Whole subsystems were not swept.**

- **IRIFactory**: `create_and_validate` (`IRIFactory.cpp:244`), `from_relative` and `from_maybe_relative` (`IRIFactory.cpp:208`, `:213`) are `noexcept` but call `IRI::make_unchecked`, which calls the now-throwing `find_or_make_id`. Declarations: `IRIFactory.hpp:58`, `:66`, `:130`. This is the funnel every parser uses to intern IRIs.
- **serd parser callbacks** (`private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp`): `get_iri` (`:65`, plus its inner `noexcept` lambda at `:66`) and `get_prefixed_iri` (`:89`) are `noexcept` and call the factory without any try/catch, unlike the sibling callbacks `get_bnode` and `get_literal` which do catch. `on_stmt` (`:294`) is `noexcept` and stringifies the graph node (`std::string{*graph_node}` at `:337`, `:377`), which reads the storage. With a throwing backend, parsing any document aborts the process instead of reporting a `ParsingError`.
- **`IRI::IRI(DatatypeIDView, DynNodeStoragePtr)`** (`IRI.cpp:24`, `IRI.hpp:19`) is `noexcept` but its dynamic-datatype branch calls `IRI::make_unchecked`. This constructor backs `IRI::datatype<T>()`, `Literal::make_typed_unchecked` and the cast paths, so any user-registered datatype hits it.
- **`Graph::match` / `Dataset::match`** (`Graph.cpp:52`, `Graph.hpp:145`, `Dataset.hpp:155`) are `noexcept` but construct a `solution_iterator` whose constructor (correctly de-noexcept'd, `Graph.cpp:163`) compares nodes against the pattern and therefore reads the storage.

**3. Thin wrappers split from their base within the same file.**

- `BlankNode::order_ne`/`eq`/`ne` (`BlankNode.cpp:236` ff.) kept `noexcept` while `order`/`order_eq` directly above them lost it. Same for `Variable::order_ne`/`eq`/`ne` (`Variable.cpp:256` ff.). Consequence: `Node::eq` and `Node::operator==` still terminate for blank nodes and variables, while working for IRIs and literals.
- `Literal::language_tag_eq` (both overloads, `Literal.cpp:637`, `:646`) and `as_language_tag_eq` (`:653`) kept `noexcept` while `language_tag()` (`:617`) and `make_boolean` (`:220`) lost it.
- `Literal::dynamic_datatype_eq_impl` (`Literal.cpp:155`) and the templates `datatype_eq<T>()` / `as_datatype_eq<T>()` (`Literal.hpp:642` ff.) kept `noexcept` while the non-template `datatype_eq` overloads (`Literal.hpp:661`, `:670`) lost it.

**4. Individual spots.**

- `Literal::numeric_unop_impl`: the `noexcept` lambda at `Literal.cpp:1091` calls `this->value()` (`:1096`, `:1098`), which reads the literal backend.
- `Literal::as_regex_matches`: the `noexcept` lambda at `Literal.cpp:1982` catches only `std::runtime_error`; any other exception type from the storage (reached via `make_regex` reading the lexical form) terminates. Exceptions that do derive from `runtime_error` are silently swallowed into "no match", which is also wrong.
- `IncreasingIdGenerator` constructor (`IncreasingIdGenerator.cpp:15`, `.hpp:27`): declared `noexcept`, documented `@throws ParsingError`, and the body comments "throws if invalid". The contradiction predates this PR, but the file was touched in this round and the sweep should catch it.
- Dead declaration: `Literal::serialize_impl` (`Literal.hpp:282`) is still declared `noexcept` and has no definition or caller. Remove it.
- `tests/serializer/tests_Serialize.cpp:18`: the test helper lambda is `noexcept` and calls `node.serialize`; a storage exception would abort the test binary instead of failing one test.

**5. Minor over-removal.** `lang_matches(std::string_view, std::string_view)` (`Literal.cpp:2547`) and `Dataset::solution_iterator::fill_solution` (`Dataset.cpp:204`) lost `noexcept` although they never touch the storage. Harmless, but it blurs the policy line; keeping `noexcept` where nothing can throw preserves information for callers and the optimizer.

## Documentation, versioning, tests (unchanged from last round)

- Still no `@throws` documentation anywhere for the newly throwing `find`/`try_get_in_node_storage`/accessor functions, and `src/rdf4cpp/storage/README.md` does not mention that backends may throw.
- `README.md:139` promises a minor version bump when the stable high-level API breaks. Changing the exception contract of most of the public API arguably qualifies; worth a decision, not necessarily a blocker.
- Still no test with a throwing storage. One doctest case would have caught every class of issue above, including the parser and `operator std::string` paths:

```cpp
struct ThrowingNodeStorage { /* find_* / find_or_make_* all throw */ };
NodeStorage auto ns = ThrowingNodeStorage{};
CHECK_THROWS_AS(IRI::find("http://ex", ns), std::runtime_error);
CHECK_THROWS_AS((void)(foreign_iri == other_foreign_iri), std::runtime_error);
CHECK_THROWS_AS((void)static_cast<std::string>(foreign_iri), std::runtime_error);
```

## Nit

- `Graph.cpp:165`: removing `noexcept` from the `solution_iterator` constructor left the member-initializer continuation lines misaligned.

## Verdict

Clear progress: everything flagged last round is fixed, and the probe confirms the comparison, migration, and find paths now propagate correctly. But the sweep worked from the previous review's list rather than from the call graph, so the layers nobody listed explicitly (IRIFactory, serd callbacks, the DatatypeIDView constructor, match(), the eq/ne wrappers, the oneshot lambdas) still terminate. The parser path is the most important one: a throwing storage kills the process on the first parsed IRI. Recommend one more systematic pass (grep for `noexcept` in every file that reaches `DynNodeStoragePtr`, or just add the throwing-storage test and chase the aborts), plus the missing `@throws` docs and the test.

(unchecked review by claude. take with caution)

liss-h:
I would not refactor the IRIFactory (apart for removing noexcept). We need to get rid of the expected there at some point I think (mix of expected and exceptions does not make sense, but this is another issue).

src/rdf4cpp/IRI.cpp
```diff
@@ -127,7 +127,7 @@ bool IRI::serialize(writer::BufWriterParts const writer) const noexcept {
     return true;
 }
 
-IRI::operator std::string() const noexcept {
+IRI::operator std::string() const {
     return writer::StringWriter::oneshot([this](auto &w) noexcept {
```
bigerl: **[critical]** `operator std::string` lost `noexcept`, but this lambda is still `noexcept`. A throwing `serialize` therefore still calls `std::terminate` (reproduced: SIGABRT with a throwing storage). `StringWriter::oneshot` does not require a `noexcept` callable (`writer/BufWriter.hpp:223`), so the `noexcept` can just be dropped. Same pattern in `BlankNode.cpp:173`, `Literal.cpp:808`, `query/Variable.cpp:195`.

src/rdf4cpp/BlankNode.cpp
```diff
@@ -157,7 +157,7 @@ FetchOrSerializeResult BlankNode::fetch_or_serialize_identifier(std::string_view
     return FetchOrSerializeResult::Fetched;
 }
 
-bool BlankNode::serialize(writer::BufWriterParts const writer) const noexcept {
+bool BlankNode::serialize(writer::BufWriterParts const writer) const {
     if (null()) {
         return rdf4cpp::writer::write_str("null", writer);
     }
```
bigerl, line 163: **[critical]** (refers to lines 172-175 below) `BlankNode::operator std::string()` is the only one of the four conversion operators that is still `noexcept` itself (declaration `BlankNode.hpp:88`), and its `oneshot` lambda is `noexcept` too. Both need the `noexcept` removed, otherwise a throwing `serialize` terminates.

src/rdf4cpp/Literal.cpp
```diff
@@ -796,37 +796,37 @@ bool Literal::serialize(writer::BufWriterParts const writer, NodeSerializationOp
...
 }
 
-bool Literal::serialize_simplified_lexical_form(writer::BufWriterParts const writer) const noexcept {
+bool Literal::serialize_simplified_lexical_form(writer::BufWriterParts const writer) const {
     return serialize_lexical_form_impl<true>(ConsumeSerialize{writer});
 }
 
-Literal::operator std::string() const noexcept {
+Literal::operator std::string() const {
     return writer::StringWriter::oneshot([this](auto &w) noexcept {
```
bigerl: **[critical]** Same as in `IRI.cpp:131`: the lambda is still `noexcept` while `serialize` can now throw. Remove the `noexcept` from the lambda.

src/rdf4cpp/query/Variable.cpp
```diff
@@ -164,7 +164,7 @@ FetchOrSerializeResult Variable::fetch_or_serialize_name(std::string_view &out,
     return FetchOrSerializeResult::Fetched;
 }
 
-bool Variable::serialize(writer::BufWriterParts const writer) const noexcept {
+bool Variable::serialize(writer::BufWriterParts const writer) const {
     if (null()) {
         return rdf4cpp::writer::write_str("null", writer);
     }
```
bigerl, line 170: **[critical]** (refers to line 195 below) The `oneshot` lambda in `Variable::operator std::string()` is still `noexcept` while `serialize` can now throw. Remove the `noexcept` from the lambda.

src/rdf4cpp/BlankNode.cpp
```diff
@@ -229,7 +229,7 @@ std::strong_ordering BlankNode::order(BlankNode const &other) const noexcept {
...
 }
 
-bool BlankNode::order_eq(BlankNode const &other) const noexcept {
+bool BlankNode::order_eq(BlankNode const &other) const {
     return order(other) == std::strong_ordering::equivalent;
 }
```
bigerl, line 235: **[critical]** (refers to lines 236-249 below) `order_ne`, `eq` and `ne` kept `noexcept` although `order`/`order_eq` right above lost it and read the backend. Consequence: `Node::eq` and `Node::operator==` still terminate for blank nodes from a throwing storage, while they propagate for IRIs and literals.

src/rdf4cpp/query/Variable.cpp
```diff
@@ -250,7 +250,7 @@ std::strong_ordering Variable::order(Variable const &other) const noexcept {
...
 }
 
-bool Variable::order_eq(Variable const &other) const noexcept {
+bool Variable::order_eq(Variable const &other) const {
     return order(other) == std::strong_ordering::equal;
 }
 bool Variable::order_ne(Variable const &other) const noexcept {
```
bigerl: **[critical]** `order_ne`, `eq` and `ne` (lines 256-266) kept `noexcept` although `order`/`order_eq` lost it. Same split as in `BlankNode.cpp`; `Node::operator==` still terminates for variables.

src/rdf4cpp/Literal.cpp
```diff
@@ -614,7 +614,7 @@ Literal Literal::as_simplified_lexical_form(storage::DynNodeStoragePtr node_stor
     return Literal::make_simple_unchecked(lex, needs_escape, select_node_storage(node_storage));
 }
 
-std::string_view Literal::language_tag() const noexcept {
+std::string_view Literal::language_tag() const {
     if (this->datatype_eq<datatypes::rdf::LangString>()) {
         if (this->is_inlined()) {
             auto [tag, _] = rdf4cpp::datatypes::registry::DatatypeRegistry::LangTagInlines::from_inlined(this->handle_.node_id().literal_id());
```
bigerl, line 620: **[critical]** (refers to lines 637-656 below) Both `language_tag_eq` overloads and both `as_language_tag_eq` overloads kept `noexcept`, but they call `language_tag()` (line 617, de-noexcept'd, reads the literal backend) and `make_boolean` (line 220, de-noexcept'd, writes to the storage).

src/rdf4cpp/Literal.hpp
```diff
@@ -658,7 +658,7 @@ struct Literal : Node {
      * @param datatype the datatype to compare against
```
bigerl, line 649: **[critical]** The template `datatype_eq<T>()` (line 642) and `as_datatype_eq<T>()` kept `noexcept` while these non-template overloads lost it. The template path reaches the now-throwing `datatype()`/`IRI::identifier()` through `dynamic_datatype_eq_impl` (`Literal.cpp:155`, declaration line 272), which is also still `noexcept`.

src/rdf4cpp/IRI.cpp
```diff
@@ -69,7 +69,7 @@ IRI IRI::to_node_storage(storage::DynNodeStoragePtr node_storage) const {
     return IRI{storage::identifier::NodeBackendHandle{node_id, node_storage}};
```
bigerl, line 69: **[critical]** (refers to `IRI.cpp:24`, declaration `IRI.hpp:19`) The `IRI(DatatypeIDView, DynNodeStoragePtr)` constructor is still `noexcept`, but its dynamic-datatype branch calls `IRI::make_unchecked` -> `find_or_make_id`. It backs `IRI::datatype<T>()`, `Literal::make_typed_unchecked` and the cast paths, so any user-registered datatype hits it.

src/rdf4cpp/Graph.hpp
```diff
@@ -140,7 +140,7 @@ struct Graph {
     void add(Statement const &statement);
 
     [[nodiscard]] size_t size() const noexcept;
-    [[nodiscard]] bool contains(Statement const &statement) const noexcept;
+    [[nodiscard]] bool contains(Statement const &statement) const;
 
     [[nodiscard]] solution_sequence match(query::TriplePattern const &triple_pattern) const noexcept;
```
bigerl: **[critical]** `Graph::match` is still `noexcept` (`Graph.cpp:52`), but it constructs a `solution_iterator` whose constructor was correctly de-noexcept'd and compares nodes against the pattern, which reads the storage.

src/rdf4cpp/Dataset.hpp
```diff
@@ -135,10 +135,10 @@ struct Dataset {
...
 
-    [[nodiscard]] bool contains(Quad const &quad) const noexcept;
+    [[nodiscard]] bool contains(Quad const &quad) const;
 
     [[nodiscard]] size_t size() const noexcept;
-    [[nodiscard]] size_t size(IRI const &graph_name) const noexcept;
+    [[nodiscard]] size_t size(IRI const &graph_name) const;
 
     Graph *find_graph(Node const &graph);
     Graph *find_graph();
```
bigerl, line 144: **[critical]** (refers to line 155 below) `Dataset::match` is still `noexcept`, but its `solution_iterator` calls `find_graph` and `Graph::match` and compares nodes. Same issue as `Graph::match`.

src/rdf4cpp/Literal.cpp
```diff
@@ -1119,7 +1119,7 @@ Literal Literal::numeric_unop_impl(OpSelect op_select, storage::DynNodeStoragePt
```
bigerl, line 1119: **[major]** (refers to lines 1091-1098 above) The lambda in `numeric_unop_impl` is `noexcept` but calls `this->value()`, which reads the literal backend and can now throw.

src/rdf4cpp/Literal.cpp
```diff
@@ -1974,7 +1974,7 @@ Literal Literal::as_regex_matches(regex::Regex const &pattern, storage::DynNodeS
     return Literal::make_boolean(res, select_node_storage(node_storage));
 }
 
-Literal Literal::as_regex_matches(Literal const &pattern, Literal const &flags, storage::DynNodeStoragePtr node_storage) const noexcept {
+Literal Literal::as_regex_matches(Literal const &pattern, Literal const &flags, storage::DynNodeStoragePtr node_storage) const {
     if (this->null() || !this->is_string_like()) {
         return Literal{};
     }
```
bigerl, line 1980: **[major]** (refers to the lambda at line 1982) The `noexcept` lambda catches only `std::runtime_error`. `make_regex` reads the lexical form from the storage, so other exception types now terminate here, and storage errors that do derive from `runtime_error` are silently turned into "no match". Suggest dropping the `noexcept` and rethinking the catch.

src/rdf4cpp/bnode_mngt/reference_backends/generator/IncreasingIdGenerator.cpp
```diff
@@ -17,7 +17,7 @@ IncreasingIdGenerator::IncreasingIdGenerator(std::string prefix, size_t const in
     BlankNode::validate(prefix_); // throws if invalid
```
bigerl, line 17: **[major]** (refers to the constructor at line 15) The constructor is declared `noexcept` (`IncreasingIdGenerator.hpp:27`) but is documented `@throws ParsingError` and its body calls `BlankNode::validate` with the comment "throws if invalid". This contradiction predates the PR, but this sweep is the right moment to fix it.

src/rdf4cpp/Literal.hpp
```diff
@@ -294,7 +294,7 @@ struct Literal : Node {
      * @return whatever consume returned
```
bigerl, line 285: **[minor]** (refers to line 282 above) `serialize_impl` is a stale declaration: still `noexcept`, no definition, no caller. Remove it.

src/rdf4cpp/Literal.cpp
```diff
@@ -481,7 +481,7 @@ IRI Literal::datatype() const noexcept {
 }
 
 template<bool simplified, typename C>
-auto Literal::serialize_lexical_form_impl(C &&consume) const noexcept {
+auto Literal::serialize_lexical_form_impl(C &&consume) const {
     if (this->is_inlined()) {
         if (this->datatype_eq<datatypes::rdf::LangString>()) {
             auto const lex = this->lang_tagged_get_de_inlined()
```
bigerl, line 487: **[minor]** (refers to the visit lambda at line 511) The `LexicalFormLiteralBackendView` lambda is `noexcept` but calls `this->datatype_id()`, which for non-fixed datatypes reads the literal backend again and can throw. Narrow window (the storage already answered once), but the `noexcept` buys nothing here.

src/rdf4cpp/Literal.cpp
```diff
@@ -2544,7 +2544,7 @@ std::string normalize_unicode(std::string_view utf8) {
     return una::norm::to_nfc_utf8(utf8);
 }
 
-bool lang_matches(std::string_view const lang_tag, std::string_view const lang_range) noexcept {
+bool lang_matches(std::string_view const lang_tag, std::string_view const lang_range) {
```
bigerl: **[minor]** Over-removal: `lang_matches(std::string_view, std::string_view)` never touches a node storage and everything in its body is noexcept. Keeping `noexcept` here preserves information for callers and the optimizer.

src/rdf4cpp/Dataset.cpp
```diff
@@ -201,7 +201,7 @@ bool operator==(Dataset::sentinel, Dataset::iterator const &self) noexcept {
     return self.giter_ == self.gend_;
 }
 
-void Dataset::solution_iterator::fill_solution() noexcept {
+void Dataset::solution_iterator::fill_solution() {
```
bigerl: **[minor]** Over-removal: `fill_solution` only calls noexcept operations, and the sibling `Dataset::iterator::operator++` doing the same work kept `noexcept`. Either both or neither.

bigerl (commented):
Inline version of the re-review findings, one thread per item so they can be resolved one by one. Details and runtime proof are in the re-review comment above.

Adjusted after @liss-h's comment: the IRIFactory item below is only about removing the stray `noexcept`s. The `expected`/exception mix there stays as it is (separate issue).

The following items concern files that are not part of this diff, so they cannot carry inline comments:

- `src/rdf4cpp/IRIFactory.cpp:244`, `:208`, `:213` (declarations `IRIFactory.hpp:58`, `:66`, `:130`): `create_and_validate`, `from_relative` and `from_maybe_relative` are still `noexcept` but call `IRI::make_unchecked` -> `find_or_make_id`. Every parser interns IRIs through this funnel. Only remove the `noexcept`s, nothing else.
- `private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp`: `get_iri` (`:65`, plus its inner `noexcept` lambda at `:66`) and `get_prefixed_iri` (`:89`) are `noexcept` and call the IRI factory without try/catch, unlike `get_bnode` and `get_literal` which do catch. `on_stmt` (`:294`) is `noexcept` and stringifies the graph node (`std::string{*graph_node}` at `:337`, `:377`), which reads the storage. With a throwing backend, parsing aborts the process on the first IRI instead of reporting a `ParsingError`.
- `tests/serializer/tests_Serialize.cpp:18`: the helper lambda is `noexcept` and calls `node.serialize`; a storage exception aborts the test binary instead of failing one test.
- No `@throws` documentation was added for the newly throwing `find`/`try_get_in_node_storage`/accessor functions, and `src/rdf4cpp/storage/README.md` does not mention that backends may throw.
- `README.md:139` promises a minor version bump for breaking changes to the stable high-level API; the exception contract change arguably qualifies.
- Still no test with a throwing storage backend; one such doctest case would have caught every item in this list.

(unchecked review by claude. take with caution)

src/rdf4cpp/Graph.cpp
```diff
@@ -154,20 +154,20 @@ bool Graph::solution_iterator::check_solution() noexcept {
...
+void Graph::solution_iterator::forward_to_solution() {
     while (iter_ != std::default_sentinel && !check_solution()) {
         ++iter_;
     }
 }
 
 Graph::solution_iterator::solution_iterator(typename Graph::iterator beg,
-                                            query::TriplePattern const &pat) noexcept : iter_{beg},
+                                            query::TriplePattern const &pat) : iter_{beg},
                                                                                         pat_{pat},
```
bigerl: **[nit]** Removing `noexcept` from the constructor left the member-initializer continuation lines misaligned.

private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.hpp
```diff
@@ -45,15 +45,15 @@ struct IStreamQuadIterator::ImplSerd final : Impl {
...
+    nonstd::expected<IRI, SerdStatus> get_iri(SerdNode const *node);
+    nonstd::expected<IRI, SerdStatus> get_prefixed_iri(SerdNode const *node);
     nonstd::expected<Literal, SerdStatus> get_literal(SerdNode const *literal, SerdNode const *datatype, SerdNode const *lang) noexcept;
     SerdStatus inspect_node(Node const &node) noexcept;
 
     static SerdStatus on_error(void *voided_self, SerdError const *error) noexcept;
     static SerdStatus on_base(void *voided_self, SerdNode const *uri) noexcept;
     static SerdStatus on_prefix(void *voided_self, SerdNode const *name, SerdNode const *uri) noexcept;
-    static SerdStatus on_stmt(void *voided_self, SerdStatementFlags, SerdNode const *graph, SerdNode const *subj, SerdNode const *pred, SerdNode const *obj, SerdNode const *obj_datatype, SerdNode const *obj_lang) noexcept;
+    static SerdStatus on_stmt(void *voided_self, SerdStatementFlags, SerdNode const *graph, SerdNode const *subj, SerdNode const *pred, SerdNode const *obj, SerdNode const *obj_datatype, SerdNode const *obj_lang);
```
liss-h: this needs to stay noexcept. It is called from C

liss-h (approved):
Thank you, the more complex fixes will be done in a seperate PR

PR #462, opened 2026-07-29
Testing built in functions - return null nodes if invoked from null nodes

SandroMiudo:


liss-h (changes requested):
missing checks for is_* functions (e.g. is_numeric)

nkaralis:
> missing checks for is_* functions (e.g. is_numeric)

we don't have `as_is_numeric`. we only have `is_numeric` and the result is `bool`

liss-h:
just check that it returns false

liss-h (approved):
Thank you

PR #463, opened 2026-07-30
Inlcude macos-26 arm64 image to ci - gcc 15, darwin 25

### LINKED ISSUE #461: Date parsing broken in macOS Debug builds: days-per-month table is corrupted at link time (Homebrew GCC 15)
(reproduced with and described by claude)

## Summary

A Debug build of rdf4cpp on macOS with Homebrew GCC 15.3.0 rejects almost every valid date. Example: `"2026-07-27T14:00:00.123+00:00"^^xsd:dateTimeStamp` fails with `parsing error: 2026-07-27 is invalid`. The same inputs are accepted by RelWithDebInfo and Release builds on macOS and by every build on Linux. Everyone developing on macOS hits this in Debug builds, including through downstream projects (found via failing tentris-frontend unit tests).

## Root cause

`last_day_in_month` in `src/rdf4cpp/Timezone.hpp` keeps the days-per-month table as a function-local constant:

```cpp
static constexpr std::chrono::day last_day_in_month(Year year, Month month) noexcept {
    RDF4CPP_ASSERT(month.ok());
    constexpr unsigned char common[]{31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
    ...
}
```

GCC emits the array into `__TEXT,__const` under a stripped Mach-O temporary label (`L.str.18`). In a binary with a large `__TEXT,__const` the Apple linker (ld-1267, also with `-Wl,-ld_classic`) resolves the relocation to a wrong address. At runtime the function reads `{0,0,0,51,0,0,0,0,0,1,0,0}` instead of the table, `YearMonthDay::ok()` is false for nearly every date, and every date datatype throws (`Date.cpp`, `DateTime.cpp`, `DateTimeStamp.cpp`). `2024-02-29` still parses because the leap year branch uses the literal `29u`.

Evidence that this is a link-time fault, not a source bug:

- In one linked binary the `adrp/add` pair loads the array from `0x1005f4614` while the bytes sit at `0x1005f0bd8`.
- Recompiling only the consumer translation unit at `-O2` against the identical Debug `librdf4cpp.a` changes which months are rejected.
- clang is not affected. It emits a preserved `l___const.*` label for exactly this kind of constant.

Affected: v0.1.13 and v0.1.15 (datetime code unchanged between them), macOS 26.5.2, Homebrew GCC 15.3.0 (`aarch64-apple-darwin25`), `CMAKE_BUILD_TYPE=Debug`. Not affected: `-O1` and higher, clang, Linux.

## Failing test

Fails in a macOS gcc-15 Debug build, passes everywhere else:

```cpp
#include <doctest/doctest.h>
#include <rdf4cpp.hpp>

TEST_CASE("xsd date datatypes accept valid dates") {
    using namespace rdf4cpp;
    IRI const dts{"http://www.w3.org/2001/XMLSchema#dateTimeStamp"};
    CHECK_NOTHROW(Literal::make_typed("2026-07-27T14:00:00.123+00:00", dts));
    CHECK_NOTHROW(Literal::make_typed("2026-07-27T14:00:00.000+00:00", dts));
    CHECK_NOTHROW(Literal::make_typed("2026-07-27T14:00:00Z", dts));
    CHECK_NOTHROW(Literal::make_typed("2026-12-31T00:00:00Z", dts));
    CHECK_NOTHROW(Literal::make_typed("2026-07-27", IRI{"http://www.w3.org/2001/XMLSchema#date"}));
}
```

## Suggested fix

The underlying defect sits in GCC or the Apple linker. rdf4cpp can sidestep it by moving the table to namespace scope, which gives it a named, preserved symbol:

```cpp
inline constexpr unsigned char days_in_month_common[]{31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
```

SandroMiudo:
- fixed cmake retrieval of swdf @bigerl

### DIFF
--- a/.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
+++ b/.github/workflows/reusable_run_test_and_examples_by_linking_type.yml
@@ -45,14 +45,20 @@ jobs:
           if [[ ${CC_VER:0:4} != gcc- ]] ; then exit 1; fi
           GCC_VERSION=${CC_VER:4:6}
           echo GCC_VERSION="${GCC_VERSION}" >> $GITHUB_ENV
-          
+
+          if [[ "${{ inputs.os }}" == "macos-15" ]]; then
+            echo MACPORTS_VERSION=2.10.7 >> $GITHUB_ENV
+          elif [[ "${{ inputs.os }}" == "macos-26" ]]; then
+            echo MACPORTS_VERSION=2.12.5 >> $GITHUB_ENV
+          fi
+
       - name: Cache MacPorts
-        if: ${{ !contains(inputs.os, 'macos') }}
+        if: ${{ contains(inputs.os, 'macos') }}
         uses: actions/cache@v6.1.0
         id: cache-macports
         with:
           path: /opt/local/
-          key: macports-v2.10.7-${{ inputs.os }}-${{ inputs.compiler }}
+          key: macports-${{ env.MACPORTS_VERSION }}-${{ inputs.os }}-${{ inputs.compiler }}
 
       - name: Fix MacPorts permissions
         if: ${{ contains(inputs.os, 'macos') && steps.cache-macports.outputs.cache-hit == 'true' }}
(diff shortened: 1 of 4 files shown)


bigerl:
@SandroMiudo, now the tests are failing that we expected to fail. Please patch it now.

liss-h (approved):

PR #464, opened 2026-07-31
Fix/json ld parser

bigerl:
(claude implemented)

This addresses (most) of the [claude fable max review](https://github.com/user-attachments/files/30579874/code-review-pr426.md) reported issues on the json-ld parser, i.e. 32/35 issues as well as 4 more issues found along the way. 

The issues were addressed by creating a regression test first, where possible. 

The issue column refers to the finding number in `code-review-pr426.md`, `new` marks an issue found while working on the review.

<table>
<thead>
<tr><th>ID</th><th>Name</th><th>Description</th><th>Commit</th><th>Issue</th></tr>
</thead>
<tbody>
<tr><td colspan="5"><strong>Correctness of the JSON-LD parser</strong></td></tr>
<tr><td>CP01</td><td>Malformed JSON escaped as <code>simdjson::simdjson_error</code> instead of a parse-error quad</td><td>Empty input threw out of the <code>IStreamQuadIterator</code> constructor, truncated input threw from <code>operator++</code>. <code>parse()</code> now converts <code>simdjson_error</code> and <code>InvalidNode</code> into a <code>ParsingError</code>, the behavior <code>IStreamQuadIterator.hpp</code> documents and the serd backend already had.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/23acb48">23acb48</a></td><td>1</td></tr>
<tr><td>CP02</td><td>The parser state leaked when construction threw</td><td><code>state_</code> was a raw pointer released only in the destructor, while a later member initializer resumed the generator and could throw. Ownership is a <code>unique_ptr</code> now.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/23acb48">23acb48</a></td><td>10</td></tr>
<tr><td>CP03</td><td><code>set_base</code> on the user provided state was never restored</td><td>A document with <code>"@base"</code> changed relative IRI resolution of every later parse that shared the <code>ParsingState</code>. The context parser resolves through its own <code>IRIFactory</code> now, and only re-validates the base when it changes.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/23acb48">23acb48</a></td><td>2</td></tr>
<tr><td>CP04</td><td>The compact IRI consistency check skipped one-character suffixes</td><td>The colon search excluded one character too many, so every term like <code>ex:x</code> bypassed the step 14 re-expansion check.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/23acb48">23acb48</a></td><td>5</td></tr>
<tr><td>CP05</td><td>Undefined behavior on large integral numbers</td><td><code>1e19</code> reached <code>static_cast&lt;int64_t&gt;</code> and saturated to <code>9223372036854775807</code>. Integral values outside int64 now take a canonical decimal path.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/23acb48">23acb48</a></td><td>7</td></tr>
<tr><td>CP06</td><td>Numbers in CRLF documents threw <code>InvalidNode</code></td><td>The token trim did not strip <code>\r</code>.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/23acb48">23acb48</a></td><td>6</td></tr>
<tr><td>CP07</td><td><code>KeepBlankNodeIds</code> and <code>NoParseBlankNode</code> were ignored</td><td>Both are honored now. <code>BlankNode::make</code> failures are caught and reported instead of escaping.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/23acb48">23acb48</a></td><td>9</td></tr>
<tr><td>CP08</td><td>A string <code>@context</code> reported "invalid local context"</td><td>The message naming the remote-context limitation was only reachable for a string inside an array, not for the plain string form.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/bb69500">bb69500</a></td><td>new</td></tr>
<tr><td>CP09</td><td>A term whose <code>@id</code> or <code>@reverse</code> value has the form of a keyword was rejected</td><td>The specification leaves the term undefined instead, see create term definition 14.2.3 and 13.4. Enables the w3c tests <code>pr36</code>, <code>pr37</code>, <code>pr38</code>, <code>pr39</code> and <code>e120</code>.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/08ea3a3">08ea3a3</a></td><td>new</td></tr>
<tr><td>CP10</td><td>Invalid language tags produced unserializable literals</td><td><code>{"@language": "a b"}</code> yielded <code>"v"@a b</code>, which cannot be written as N-Triples. The tag is checked against the RDF grammar now and reported as a <code>ParsingError</code>.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/08ea3a3">08ea3a3</a></td><td>new</td></tr>
<tr><td colspan="5"><strong>Correctness of <code>try_compare_graphs_fast</code></strong> (the comparison behind every parser test, so a defect here hides parser defects)</td></tr>
<tr><td>CG01</td><td>Non-injective blank node mappings were accepted</td><td><code>{_:x &lt;p&gt; _:x}</code> compared equal to <code>{_:y &lt;p&gt; _:z}</code>. The mapping is tracked in both directions now.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/4b016c9">4b016c9</a></td><td>3</td></tr>
<tr><td>CG02</td><td>Different graph sizes aborted the process</td><td><code>RDF4CPP_ASSERT</code> calls <code>std::abort()</code> under NDEBUG as well, in a public header that documents no such precondition. Size and arity mismatches return an error now.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/4b016c9">4b016c9</a></td><td>4</td></tr>
<tr><td>CG03</td><td>The comparator was not a strict weak ordering</td><td>Blank nodes were ordered by their arbitrary handles against non-blank nodes, which is both non-canonical and undefined behavior in <code>std::ranges::sort</code>. Non-blank nodes now sort before blank nodes.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/4b016c9">4b016c9</a></td><td>new</td></tr>
<tr><td colspan="5"><strong>Performance</strong></td></tr>
<tr><td>PF01</td><td>The similarity pre-check was quadratic</td><td>The counts come from a <code>boost::unordered_flat_map</code> now, keyed on a pair of positions and the nodes at them and hashed with <code>dice::hash</code>. One <code>emplace</code> per key fills it, its result decides whether the quad index still has to be appended. Verified equivalent to the previous implementation over 2576 randomized quads. The second commit fixes the count, <code>std::ranges::unique</code> returns the removed duplicates, not the unique elements.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/4b016c9">4b016c9</a>, <a href="https://github.com/rdf4cpp/rdf4cpp/commit/8ebf4b6">8ebf4b6</a>, <a href="https://github.com/rdf4cpp/rdf4cpp/commit/d873073">d873073</a></td><td>24</td></tr>
<tr><td>PF02</td><td>Term lookup was a linear scan</td><td>A lazily built <code>boost::unordered_flat_map</code>, used above 16 terms so that small contexts keep the cheaper scan. Parse time of a document with 2800 context terms drops from 45.7 ms to 6.6 ms, at 200 terms from 9.3 ms to 4.4 ms, best of five runs. The same document needs 7.5 ms and 4.9 ms with a <code>std::unordered_map</code>, so the open addressing map takes another 11 to 12 percent off the whole parse.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/e24214d">e24214d</a>, <a href="https://github.com/rdf4cpp/rdf4cpp/commit/d873073">d873073</a></td><td>22</td></tr>
<tr><td>PF03</td><td>The input stream was read in 1024 byte blocks</td><td>64 KiB now. The simdjson capacity factor stays at 5 and says why: simdjson advances one string buffer per document and never reuses it, and this parser reads the same fields again in every pass.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/e24214d">e24214d</a></td><td>23</td></tr>
<tr><td colspan="5"><strong>Test infrastructure</strong></td></tr>
<tr><td>TI01</td><td>Remote test files were cached without checking the HTTP status</td><td>A 404 body was written to the on-disk cache and replayed forever. <code>CURLOPT_FAILONERROR</code> plus a response code check now, and the uninitialized <code>CURLcode</code> is gone.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/16f0553">16f0553</a></td><td>11</td></tr>
<tr><td>TI02</td><td>The CI cache key for test inputs was the constant <code>parser_test</code></td><td>The cache was written once and never updated. It hashes the test sources now and has restore-keys.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/deec41b">deec41b</a></td><td>12</td></tr>
<tr><td>TI03</td><td>Deduplication used value equality while the comparison uses term equality</td><td>The deduplicated set could diverge from the comparison that runs on it afterwards. Both sides use term equality now, and the truth graph is deduplicated as well, which <code>try_compare_graphs_fast</code> requires.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/16f0553">16f0553</a></td><td>15</td></tr>
<tr><td colspan="5"><strong>Build and documentation</strong></td></tr>
<tr><td>BD01</td><td>C++23 was not declared</td><td>The library moved to C++23 for <code>std::generator</code> but nothing said so. <code>target_compile_features(rdf4cpp PUBLIC cxx_std_23)</code> exports it, README, <code>conanfile.py</code> and the getting started guide agree now.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/deec41b">deec41b</a></td><td>13</td></tr>
<tr><td>BD02</td><td><code>ParsingFlags</code> did not say which parser honors which flag</td><td>Documented per flag, including that the direction flags are JSON-LD only and that a flag the selected parser does not know is ignored without a diagnostic.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/deec41b">deec41b</a></td><td>8</td></tr>
<tr><td>BD03</td><td>The opaque-path base change was neither pinned nor explained</td><td><code>urn:</code> cases in <code>tests_IRIFactory</code> and a comment naming RFC 3986 5.2.3. The new behavior is the RFC one, the old unconditional merge was the bug.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/deec41b">deec41b</a></td><td>14</td></tr>
<tr><td>BD04</td><td><code>from_maybe_relative_as_string</code> hid a thread_local lifetime contract</td><td>The contract is stated on the method now, including that the result points into the caller's memory for an absolute input. The validation is shared with <code>create_and_validate</code> through a new <code>IRIFactory::validate</code>.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/e24214d">e24214d</a></td><td>18</td></tr>
<tr><td>BD05</td><td>Comments narrated changes instead of describing the code</td><td>"moved here from create_term" and similar are gone, together with the question marks and arrows found in a sweep of the whole parser. <code>e014</code> is enabled and pinned to the result the json-ld-1.1 prefix rules produce.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/6a0dc2f">6a0dc2f</a></td><td>32, 33, 34, 35</td></tr>
<tr><td>BD06</td><td><code>try_compare_graphs_fast</code> is a test helper in the public API</td><td>It stays public and installed, removing it would break the API of 0.1.16 and 0.1.17. Its doc comment states now what it guarantees, that it expects deduplicated graphs, and that a size mismatch comes back as a pair of null nodes. Its defects are CG01 to CG03.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/4b016c9">4b016c9</a></td><td>16</td></tr>
<tr><td>BD07</td><td>The limitations were undocumented</td><td>The users guide lists the JSON-LD limitations, the README lists all 57 deactivated tests of the w3c suite with the reason for each, and every deactivated test carries its reason in the test file.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/7b77100">7b77100</a>, <a href="https://github.com/rdf4cpp/rdf4cpp/commit/a9ed040">a9ed040</a>, <a href="https://github.com/rdf4cpp/rdf4cpp/commit/ae9fcc8">ae9fcc8</a></td><td>new</td></tr>
<tr><td colspan="5"><strong>Simplification</strong></td></tr>
<tr><td>SI01</td><td>Dead members and a wrong operator</td><td><code>ExpandedMapEntry::as_named_graph</code> was never assigned, <code>::active_property</code> never read, and <code>operator^</code> for <code>ContainerMapping</code> performed an OR and had no caller.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/421d849">421d849</a></td><td>26, 27, 28</td></tr>
<tr><td>SI02</td><td>Duplicated emission, expansion and recursion</td><td>The two literal emission sites share an <code>emit_literal</code> generator, the <code>value</code> overload of <code>value_expansion</code> delegates strings to the <code>string_view</code> overload, and array recursion copies <code>ParseParams</code> instead of re-packing all 11 fields.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/421d849">421d849</a></td><td>29, 30, 31</td></tr>
<tr><td>SI03</td><td>Vocabulary and blank node labels spelled inline</td><td>The rdf vocabulary of the compound direction form is named once. The blank node label prefixes are named constants with the collision rule written down.</td><td><a href="https://github.com/rdf4cpp/rdf4cpp/commit/23acb48">23acb48</a></td><td>19, 20</td></tr>
</tbody>
</table>

## Behavior changes to confirm before merging

1. JSON-LD parse errors arrive as `ParsingError` quads. Callers that catch `simdjson_error` no
   longer see it.
2. The JSON-LD parser no longer changes the base of a shared `ParsingState`.
3. `KeepBlankNodeIds` and `NoParseBlankNode` now take effect for JSON-LD.
4. Contexts like `{"ex": "http://ex/", "ex:x": "http://other/x"}` are rejected now, as the spec
   requires.
5. A term mapping to something keyword-like is ignored instead of raising an error, so documents
   that previously failed now produce quads.
6. Invalid language tags are reported instead of silently producing an invalid literal.
7. `cxx_std_23` is exported, so consumers are moved to C++23 by the package.

### DIFF
--- a/private/rdf4cpp/parser/JsonLdContextParser.cpp
+++ b/private/rdf4cpp/parser/JsonLdContextParser.cpp
@@ -2,10 +2,10 @@
 
 namespace rdf4cpp::parser::json_ld {
     IRIFactoryError ContextParser::set_resolution_base(std::string_view const base) {
-        if (iri_factory.get_base() == base) {
+        if (iri_factory->get_base() == base) {
             return IRIFactoryError::Ok;
         }
-        return iri_factory.set_base(base);
+        return iri_factory->set_base(base);
     }
     nonstd::expected<Context, ContextParser::error_type> ContextParser::parse_context(simdjson::ondemand::value local_context, params::ParseContextParams p) {
         // https://www.w3.org/TR/json-ld11-api/#context-processing-algorithm
@@ -45,7 +45,7 @@ namespace rdf4cpp::parser::json_ld {
                                 result = nonstd::unexpected{make_error(ParsingError::Type::BadSyntax, "invalid base IRI")};
                                 return true;
                             }
-                            auto r = iri_factory.from_maybe_relative_as_string(*v);
+                            auto r = iri_factory->from_maybe_relative_as_string(*v);
                             if (r.has_value()) {
                                 result->base_iri = *r;
                             } else {
@@ -257,6 +257,7 @@ namespace rdf4cpp::parser::json_ld {
 
         // scoped contexts are validated after the whole context is parsed, so that they can
         // refer to terms defined later in the same context
+        // moved here from https://www.w3.org/TR/json-ld11-api/#create-term-definition 21.3
         if (result.has_value()) {
             for (auto const &t : result->terms) {
                 if (t.needs_context_check && t.context.has_value()) {
@@ -913,7 +914,7 @@ namespace rdf4cpp::parser::json_ld {
             if (set_resolution_base(active_context.base_iri) != IRIFactoryError::Ok) {
                 return nonstd::make_unexpected(make_error(ParsingError::Type::BadIri, "invalid base iri"));
             }
-            auto r = iri_factory.from_maybe_relative_as_string(*value);
+            auto r = iri_factory->from_maybe_relative_as_string(*value);
             if (r.has_value()) {
                 return IRIMapping{std::string(*r), IRIMappingType::IRI};
             } else {
--- a/private/rdf4cpp/parser/JsonLdContextParser.cpp
+++ b/private/rdf4cpp/parser/JsonLdContextParser.cpp
@@ -1,6 +1,12 @@
 #include "JsonLdContextParser.hpp"
 
 namespace rdf4cpp::parser::json_ld {
+    IRIFactoryError ContextParser::set_resolution_base(std::string_view const base) {
+        if (iri_factory.get_base() == base) {
+            return IRIFactoryError::Ok;
+        }
+        return iri_factory.set_base(base);
+    }
     nonstd::expected<Context, ContextParser::error_type> ContextParser::parse_context(simdjson::ondemand::value local_context, params::ParseContextParams p) {
         // https://www.w3.org/TR/json-ld11-api/#context-processing-algorithm
         // 1
@@ -35,11 +41,11 @@ namespace rdf4cpp::parser::json_ld {
                         result->base_iri = "";
                     } else {
                         if (IRIView{*v}.is_relative()) {
-                            if (iri_factory->set_base(result->base_iri) != IRIFactoryError::Ok) {
+                            if (set_resolution_base(result->base_iri) != IRIFactoryError::Ok) {
                                 result = nonstd::unexpected{make_error(ParsingError::Type::BadSyntax, "invalid base IRI")};
                                 return true;
                             }
-                            auto r = iri_factory->from_maybe_relative_as_string(*v);
+                            auto r = iri_factory.from_maybe_relative_as_string(*v);
                             if (r.has_value()) {
                                 result->base_iri = *r;
                             } else {
@@ -369,8 +375,10 @@ namespace rdf4cpp::parser::json_ld {
                 p.term.iri_mapping = *ex;
 
                 bool const has_slash = p.term.key.find('/') != std::string::npos;
+                // a colon anywhere but as first or last character makes the key a compact IRI,
+                // so the part to search in drops the first and the last character
                 std::string_view colon_check_part = p.term.key;
-                colon_check_part = colon_check_part.substr(0, colon_check_part.length() - 2);
+                colon_check_part = colon_check_part.substr(0, colon_check_part.length() - 1);
                 if (!colon_check_part.empty()) {
                     colon_check_part = colon_check_part.substr(1);
                 }
@@ -839,7 +847,10 @@ namespace rdf4cpp::parser::json_ld {
                 auto pre = value->substr(0, i);
                 auto post = value->substr(i + 1);
                 if (pre == "_") {
-                    return IRIMapping{std::format("bn_n{}", post), IRIMappingType::BlankNode};
+                    if (keep_document_bnode_labels) {
+                        return IRIMapping{std::string{post}, IRIMappingType::BlankNode};
+                    }
+                    return IRIMapping{std::format("{}{}", document_bnode_prefix, post), IRIMappingType::BlankNode};
                 }
                 if (post.starts_with("//")) {
                     return IRIMapping{std::string(*value), IRIMappingType::IRI};
@@ -881,10 +892,10 @@ namespace rdf4cpp::parser::json_ld {
             if (active_context.base_iri.empty()) {
                 return IRIMapping{std::string(""), IRIMappingType::None};
             }
-            if (iri_factory->set_base(active_context.base_iri) != IRIFactoryError::Ok) {
+            if (set_resolution_base(active_context.base_iri) != IRIFactoryError::Ok) {
                 return nonstd::make_unexpected(make_error(ParsingError::Type::BadIri, "invalid base iri"));
             }
-            auto r = iri_factory->from_maybe_relative_as_string(*value);
+            auto r = iri_factory.from_maybe_relative_as_string(*value);
             if (r.has_value()) {
                 return IRIMapping{std::string(*r), IRIMappingType::IRI};
             } else {
--- a/private/rdf4cpp/parser/JsonLdContextParser.cpp
+++ b/private/rdf4cpp/parser/JsonLdContextParser.cpp
@@ -216,6 +216,9 @@ namespace rdf4cpp::parser::json_ld {
             handle_ctx(o);  // 4
         } else if (local_context.is_scalar() && local_context.is_null()) { // 5.1
             return handle_null();
+        } else if (local_context.type() == simdjson::ondemand::json_type::string) {  // 5.2
+            // a string names a remote context, the same case as inside the array below
+            return nonstd::unexpected{make_error(ParsingError::Type::BadSyntax, "remote context not supported")};
         } else {
             if (!p.propagate && result->previous_context == nullptr) {
                 result->previous_context = &p.active_context;
--- a/private/rdf4cpp/parser/JsonLdContextParser.cpp
+++ b/private/rdf4cpp/parser/JsonLdContextParser.cpp
@@ -109,7 +109,7 @@ namespace rdf4cpp::parser::json_ld {
                             .previous_terms = previous_terms,
                         };
                         auto r = iri_expansion(result.value(), v, true, true, nullptr, &p_ctx);
-                        if (!r.has_value() || r->type != IRIMappingType::IRI) {  // technically blank node is only deprecated, not removed
+                        if (!r.has_value() || r->type != IRIMappingType::IRI) {  // a blank node as @vocab is deprecated, not removed
                             result = nonstd::unexpected{make_error(ParsingError::Type::BadSyntax, "invalid vocab mapping")};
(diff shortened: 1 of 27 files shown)


private/rdf4cpp/parser/JsonLdParser.hpp
```diff
@@ -54,12 +55,13 @@ namespace rdf4cpp::parser {
 
     struct IStreamQuadIterator::ImplJsonLd final : Impl {
     private:
+        // holds the state only if it was not provided by the caller
+        std::unique_ptr<state_type> owned_state_;
         state_type *state_;
-        bool state_is_owned_;
```
mcb5637, line 58: every other parser uses the same pattern, makes sense to change them as well

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -1,6 +1,12 @@
 #include "JsonLdContextParser.hpp"
 
 namespace rdf4cpp::parser::json_ld {
+    IRIFactoryError ContextParser::set_resolution_base(std::string_view const base) {
```
mcb5637: every other parser modifies its parsing state as well. having just one not do it is not great API design.
either change all others too, or revert this change and document it (although the non-const pointer given to the parser would be enough of a hint to me)

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -246,7 +255,8 @@ namespace rdf4cpp::parser::json_ld {
             }
         }
 
-        // check scoped context for errors, moved here from create_term
```
mcb5637, line 244: please re-add the moved from comment, it specifies that scoped context checks were moved from https://www.w3.org/TR/json-ld11-api/#create-term-definition 21.3 to here

mcb5637 (changes requested):
Overall, looks good, but i have a few comments:

private/rdf4cpp/parser/JsonLdContextParser.cpp
bigerl: Good point, please revert for now. 

I think we need to revisit this at some time but not now. Please create an issue. 

For context:

The three backends are not doing the same kind of thing.

Turtle `@base` is a document-level declaration, so propagating it to shared state is coherent. But `xml:base` is element-scoped, and the XML backend still writes it into the shared factory per element via `set_base_unchecked`, so it already leaks a value that has no document-level meaning. And in JSON-LD, `@base` lives in a context, contexts nest and revert, an `set_resolution_base` is called in between to resolve a relative `@base` against the current one, which is also scratch use

PR #469, opened 2026-08-05
Fix: Resolve wrong null node handling

SandroMiudo:
Fixes handled:

- Inside the ```as_contains``` bool node was returned, fixed to return the null literal
- Inside the ```regex_replace``` replacament node was not validated, before accessing it.

Tests extension:

- Extend permutation tests for 2+-arg built in calls

### DIFF
--- a/src/rdf4cpp/Literal.cpp
+++ b/src/rdf4cpp/Literal.cpp
@@ -2061,8 +2061,9 @@ Literal Literal::as_contains(Literal const &needle, storage::DynNodeStoragePtr n
         return Literal{};
     }
 
-    if (!needle.is_string_like())
-        return Literal::make_boolean(true, select_node_storage(node_storage));
+    if (!needle.is_string_like()) {
+        return Literal{};
+    }
 
     auto const needle_lex = needle.lexical_form();
     auto const res = this->contains(needle_lex);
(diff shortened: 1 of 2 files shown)


src/rdf4cpp/Literal.cpp
```diff
@@ -2022,7 +2022,7 @@ Literal Literal::regex_replace(regex::RegexReplacer const &replacer, storage::Dy
 }
 
 Literal Literal::regex_replace(Literal const &pattern, Literal const &replacement, Literal const &flags, storage::DynNodeStoragePtr node_storage) const {
-    if (this->null() || !this->is_string_like()) {
+    if (this->null() || !this->is_string_like() || replacement.null()) {
```
nkaralis: here, we should check that replacement and flags are simple literals (datatypes::xsd::String)

liss-h: Yes, every arg (except `this`) must be `datatype_eq<xsd::String>()`

liss-h: actually also applies to regex match

SandroMiudo: pattern & flags are already handled properly in the ```make_regex``` it was just that the replacement was not validated.

SandroMiudo: same applies to the regex match

liss-h (approved):

PR #471, opened 2026-08-11
Unify IRIFactory error reporting

### LINKED ISSUE #465: Use only exceptions in IRIFactory, not mix of exceptions and expected

liss-h:
only exceptions not mix of exceptions and expected.

Close #465

### DIFF
--- a/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -260,7 +260,7 @@ SerdStatus IStreamQuadIterator::ImplSerd::on_base(void *voided_self, const SerdN
             self->last_error = ParsingError{.error_type = ParsingError::Type::BadSyntax,
                                             .line = serd_reader_get_current_line(self->reader.get()),
                                             .col = serd_reader_get_current_col(self->reader.get()),
-                                            .message = "Unknown internal error. note: position may not be accurate and instead point to the end of the triple."};
+                                            .message = "Unknown internal error. note: position may not be accurate and instead point to the end of the line."};
         }
     }
 
@@ -287,7 +287,7 @@ SerdStatus IStreamQuadIterator::ImplSerd::on_prefix(void *voided_self, SerdNode 
             self->last_error = ParsingError{.error_type = ParsingError::Type::BadSyntax,
                                             .line = serd_reader_get_current_line(self->reader.get()),
                                             .col = serd_reader_get_current_col(self->reader.get()),
-                                            .message = "Unknown internal error. note: position may not be accurate and instead point to the end of the triple."};
+                                            .message = "Unknown internal error. note: position may not be accurate and instead point to the end of the line."};
         }
     }
 
--- a/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -282,7 +282,7 @@ SerdStatus IStreamQuadIterator::ImplSerd::on_prefix(void *voided_self, SerdNode 
             self->last_error = ParsingError{.error_type = ParsingError::Type::BadSyntax,
                                             .line = serd_reader_get_current_line(self->reader.get()),
                                             .col = serd_reader_get_current_col(self->reader.get()),
-                                            .message = std::format("Invalid prefix: {}. note: position may not be accurate and instead point to the end of the line.", ii.what())};
+                                            .message = std::format("{}. note: position may not be accurate and instead point to the end of the line.", ii.what())};
         } catch (...) {
             self->last_error = ParsingError{.error_type = ParsingError::Type::BadSyntax,
                                             .line = serd_reader_get_current_line(self->reader.get()),
--- a/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -108,25 +108,18 @@ nonstd::expected<IRI, SerdStatus> IStreamQuadIterator::ImplSerd::get_prefixed_ir
     auto const prefix = uri_node_view.substr(0, sep_pos);
     auto const suffix = uri_node_view.substr(sep_pos + 1);
 
+    if (!state->iri_factory.contains_prefix(prefix)) {
+        // NOTE: line, col not entirely accurate as this function is called after a triple was parsed
+        this->last_error = ParsingError{.error_type = ParsingError::Type::BadCurie,
+                                        .line = serd_reader_get_current_line(this->reader.get()),
+                                        .col = serd_reader_get_current_col(this->reader.get()),
+                                        .message = "unknown prefix. note: position may not be accurate and instead point to the end of the triple."};
+
+        return nonstd::make_unexpected(SERD_ERR_BAD_CURIE);
+    }
+
     try {
         return state->iri_factory.from_prefix(prefix, suffix, state->node_storage);
-    } catch (InvalidIRI const &ii) {
-        if (ii.type() == IRIParseError::UnknownPrefix) {
-            // NOTE: line, col not entirely accurate as this function is called after a triple was parsed
-            this->last_error = ParsingError{.error_type = ParsingError::Type::BadCurie,
-                                            .line = serd_reader_get_current_line(this->reader.get()),
-                                            .col = serd_reader_get_current_col(this->reader.get()),
-                                            .message = "unknown prefix. note: position may not be accurate and instead point to the end of the triple."};
-
-            return nonstd::make_unexpected(SERD_ERR_BAD_CURIE);
-        } else {
-            this->last_error = ParsingError{.error_type = ParsingError::Type::BadIri,
-                                            .line = serd_reader_get_current_line(this->reader.get()),
-                                            .col = serd_reader_get_current_col(this->reader.get()),
-                                            .message = std::format("unable to expand curie into valid iri. {}. note: position may not be accurate and instead point to the end of the triple.", ii.what())};
-
-            return nonstd::make_unexpected(SERD_ERR_BAD_SYNTAX);
-        }
     } catch (InvalidNode const &in) {
         this->last_error = ParsingError{.error_type = ParsingError::Type::BadIri,
                                         .line = serd_reader_get_current_line(this->reader.get()),
--- a/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
+++ b/private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
@@ -108,18 +108,25 @@ nonstd::expected<IRI, SerdStatus> IStreamQuadIterator::ImplSerd::get_prefixed_ir
     auto const prefix = uri_node_view.substr(0, sep_pos);
     auto const suffix = uri_node_view.substr(sep_pos + 1);
 
-    if (!state->iri_factory.contains_prefix(prefix)) {
-        // NOTE: line, col not entirely accurate as this function is called after a triple was parsed
-        this->last_error = ParsingError{.error_type = ParsingError::Type::BadCurie,
-                                        .line = serd_reader_get_current_line(this->reader.get()),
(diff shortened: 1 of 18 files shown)


private/rdf4cpp/parser/JsonLdParser.cpp
```diff
@@ -81,6 +89,8 @@ namespace rdf4cpp::parser {
             return json_ld::DirectionLiteralResult{Literal::make_simple(lit.value, state_->node_storage)};
         } catch (InvalidNode const &e) {
             return nonstd::make_unexpected(json_ld::make_error(ParsingError::Type::BadLiteral, e.what()));
+        } catch (...) {
+            return nonstd::make_unexpected(json_ld::make_error(ParsingError::Type::Internal, "unknown internal error"));
```
bigerl, line 93: missed sibling: `make_literal(TypedLiteralMapping)` below (line 103) still catches only `InvalidNode`. A non-InvalidNode exception from `Literal::make_typed` (bad_alloc, a datatype `from_string` throwing something else) escapes `IStreamQuadIterator` as a raw exception instead of a per-quad Internal error.

private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -865,8 +860,12 @@ namespace rdf4cpp::parser::json_ld {
                     v.append(post);
                     return IRIMapping{std::move(v), IRIMappingType::IRI, std::string{*value}};
                 }
-                if (IRIView(*value).quick_validate() == IRIFactoryError::Ok) {
+
+                try {
+                    IRIView(*value).quick_validate();
                     return IRIMapping{std::string(*value), IRIMappingType::IRI};
+                } catch (...) {
+                    // ignore
```
bigerl, line 868: swallows too much: the `return` sits inside the `try`, so bad_alloc or an internal error classifies a valid absolute IRI as "not an IRI" and expansion silently falls through to @vocab/@base (wrong triples, no error). Also one throw/unwind per non-IRI value on the expansion hot path. A non-throwing check (e.g. `optional<IRIParseError> try_validate()`) fixes both.

src/rdf4cpp/IRIView.hpp
```diff
@@ -24,6 +26,25 @@ enum struct IRIFactoryError {
...
+[[nodiscard]] std::string_view to_string_view(IRIParseError ipe) noexcept;
+
+struct InvalidIRI : InvalidNode {
+    using error_type = IRIParseError;
+
+private:
+    error_type type_;
+
+public:
+    explicit InvalidIRI(error_type ty)
```
bigerl, line 55: diagnostic regression: the offending IRI string is no longer in the message (old `IRI::validate`: `IRI {} is invalid: {}`). Consider a ctor overload taking the IRI. Also eager `std::format` on every throw; several catch sites only read `type()` and discard the message.

src/rdf4cpp/IRIView.hpp
```diff
@@ -1,19 +1,21 @@
...
 #include <string_view>
 
 namespace rdf4cpp {
+
 /**
  * std::format gets you a printable string.
  */
-enum struct IRIFactoryError {
-    Ok,
+enum struct IRIParseError : uint8_t {
```
bigerl, line 17: includes: `uint8_t` needs `<cstdint>`, `std::ranges::copy` (line 229) needs `<algorithm>`. Compiles only via transitive includes of the tested stdlibs.

src/rdf4cpp/IRIView.hpp
```diff
@@ -193,52 +214,19 @@ struct IRIView {
...
-            default:
-                s = "Unknown";
-                break;
-        }
-        return std::formatter<std::string_view>::format(s, ctx);
+struct std::formatter<rdf4cpp::IRIParseError> : std::formatter<string_view> {
+    template<typename Ctx>
+    auto format(rdf4cpp::IRIParseError const p, Ctx &ctx) const {
+        auto sv = to_string_view(p);
+        return std::ranges::copy(sv, ctx.out()).out;
```
bigerl: format spec ignored: `parse()` is inherited from `formatter<string_view>`, so `{:>20}` compiles but does nothing (the old formatter honored it). Delegate: `return formatter<string_view>::format(to_string_view(p), ctx);`

src/rdf4cpp/IRIView.cpp
```diff
@@ -5,6 +5,35 @@
 #include <rdf4cpp/util/CharMatcher.hpp>
 
 namespace rdf4cpp {
+
+std::string_view to_string_view(IRIParseError const ipe) noexcept {
```
bigerl: pattern: existing enum-to-string helpers are constexpr in the header (RDFNodeType.hpp, LiteralType.hpp).

src/rdf4cpp/IRIFactory.cpp
```diff
@@ -194,38 +194,32 @@ static std::string_view to_absolute(IRIView::AllParts const &base, std::string_v
 
 
 IRIFactory::IRIFactory(std::string_view base) {
```
bigerl, line 196: breaking: was `std::invalid_argument`, now `InvalidIRI` (a runtime_error). Downstream `catch (std::invalid_argument)` stops matching silently. Worth a changelog note.

src/rdf4cpp/IRIFactory.cpp
```diff
@@ -194,38 +194,32 @@ static std::string_view to_absolute(IRIView::AllParts const &base, std::string_v
...
     }
     return r;
 }
 
-nonstd::expected<IRI, IRIFactoryError> IRIFactory::from_prefix(std::string_view prefix, std::string_view local, storage::DynNodeStoragePtr node_storage) const {
+IRI IRIFactory::from_prefix(std::string_view prefix, std::string_view local, storage::DynNodeStoragePtr node_storage) const {
     auto i = prefixes.find(prefix);
     if (i == prefixes.end()) {
-        return nonstd::make_unexpected(IRIFactoryError::UnknownPrefix);
+        throw InvalidIRI{IRIParseError::UnknownPrefix};
```
bigerl: hot path: one throw (plus eagerly formatted message) per unknown-prefix occurrence, i.e. per triple in a bulk parse, and the serd caller only checks `type()`. A non-throwing lookup for the parser path avoids this.

src/rdf4cpp/IRIFactory.hpp
```diff
@@ -114,8 +117,9 @@ struct IRIFactory {
      * Changes the base IRI. Validates the new base IRI before setting.
      * @param b
      * @return
```
bigerl: stale `@return` on a void function. The prefix-map ctor (line 37) is also missing `@throws InvalidIRI`.

private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -245,11 +248,20 @@ SerdStatus IStreamQuadIterator::ImplSerd::on_base(void *voided_self, const SerdN
...
+        } catch (InvalidIRI const &ii) {
+            self->last_error = ParsingError{.error_type = ParsingError::Type::BadSyntax,
+                                            .line = serd_reader_get_current_line(self->reader.get()),
+                                            .col = serd_reader_get_current_col(self->reader.get()),
+                                            .message = std::format("Error setting base: {}. note: position may not be accurate and instead point to the end of the line.", ii.what())};
+        } catch (...) {
+            self->last_error = ParsingError{.error_type = ParsingError::Type::BadSyntax,
+                                            .line = serd_reader_get_current_line(self->reader.get()),
+                                            .col = serd_reader_get_current_col(self->reader.get()),
+                                            .message = "Unknown internal error. note: position may not be accurate and instead point to the end of the triple."};
```
bigerl: copy-paste: "end of the triple" here and in on_prefix (line 290); the sibling messages in these handlers say "end of the line".

private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -264,11 +276,18 @@ SerdStatus IStreamQuadIterator::ImplSerd::on_prefix(void *voided_self, SerdNode
...
                                         .message = "Encountered prefix while parsing. hint: prefixes are not allowed in the current document. note: position may not be accurate and instead point to the end of the line."};
     } else {
-        if (self->state->iri_factory.assign_prefix(node_into_string_view(name), node_into_string_view(uri)) != IRIFactoryError::Ok) {
+        try {
+            self->state->iri_factory.assign_prefix(node_into_string_view(name), node_into_string_view(uri));
+        } catch (InvalidIRI const &ii) {
+            self->last_error = ParsingError{.error_type = ParsingError::Type::BadSyntax,
+                                            .line = serd_reader_get_current_line(self->reader.get()),
+                                            .col = serd_reader_get_current_col(self->reader.get()),
+                                            .message = std::format("Invalid prefix: {}. note: position may not be accurate and instead point to the end of the line.", ii.what())};
```
bigerl: message lost the offending prefix name and stutters: "Invalid prefix: Invalid IRI: invalid prefix." (old text included the name).

private/rdf4cpp/parser/IStreamQuadIteratorSerdImpl.cpp
```diff
@@ -125,13 +123,18 @@ nonstd::expected<IRI, SerdStatus> IStreamQuadIterator::ImplSerd::get_prefixed_ir
...
+                                            .message = std::format("unable to expand curie into valid iri. {}. note: position may not be accurate and instead point to the end of the triple.", ii.what())};
 
             return nonstd::make_unexpected(SERD_ERR_BAD_SYNTAX);
         }
-    }
+    } catch (...) {
+        this->last_error = ParsingError{.error_type = ParsingError::Type::BadCurie,
+                                        .line = serd_reader_get_current_line(this->reader.get()),
+                                        .col = serd_reader_get_current_col(this->reader.get()),
+                                        .message = "Unknown internal error. note: position may not be accurate and instead point to the end of the triple."};
```
bigerl, line 141: classification drift: internal errors become BadCurie here, Type::Internal in the json-ld and xml backends; `get_iri` catches `InvalidNode`, this one `InvalidIRI`. 14 hand-rolled "unknown internal error" handlers across the parsers now; one shared exception-to-ParsingError helper would stop the drift.

src/rdf4cpp/datatypes/xsd/AnyURI.cpp
```diff
@@ -11,9 +11,11 @@ namespace rdf4cpp::datatypes::registry {
...
 template<>
 capabilities::Default<xsd_any_uri>::cpp_type capabilities::Default<xsd_any_uri>::from_string(std::string_view s) {
-    auto e = IRIView{s}.quick_validate(true);
-    if (e != IRIFactoryError::Ok) {
-        throw InvalidNode(std::format("http://www.w3.org/2001/XMLSchema#anyURI parsing error: invalid IRI: {}", e));
+
+    try {
+        IRIView{s}.quick_validate(true);
+    } catch (InvalidIRI const &iri) {
+        throw InvalidNode(std::format("http://www.w3.org/2001/XMLSchema#anyURI parsing error: invalid IRI: {}", iri.type()));
```
bigerl: reuse `iri.what()` instead of re-deriving the text from `type()`; or drop the try/catch, `InvalidIRI` already is an `InvalidNode`.

private/rdf4cpp/parser/JsonLdContextParser.cpp
bigerl: do not resolve here.

src/rdf4cpp/IRIFactory.cpp
bigerl: no need to mention explicitly. InvalidNode is already caught downstream.

src/rdf4cpp/IRIView.cpp
bigerl: yes, move to header and make constexpr

src/rdf4cpp/IRIView.hpp
bigerl: yes but accept the offending IRI as string view and make sure to consume it directly (no dangling string view).

bigerl (changes requested):
Thx for the PR. Find some comments below. Will address.

PR #475, opened 2026-08-18
Fix: use fetch content for serd instead of individual file download

liss-h:


CMakeLists.txt
```diff
@@ -28,56 +28,44 @@ find_package(simdjson REQUIRED)
...
+FetchContent_Declare(serd
+        GIT_REPOSITORY https://github.com/dice-group/serd.git
+        GIT_TAG 12321d54430b4e9caf30903eff0203a372c94fd0 # v0.32.10+dice
+)
+FetchContent_MakeAvailable(serd)
+
+# keep serd's license at a stable path, it is picked up from there while packaging (see conanfile.py)
+file(COPY "${serd_SOURCE_DIR}/COPYING" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/serd")
+
 set(serd_source_files
```
nkaralis, line 44: can we get rid of this `set` op?

liss-h: I think it is considered bad practice to use globs for lists for source files

nkaralis (approved):
Thank you

PR #476, opened 2026-08-19
Fix: use fetchcontent for test download instead of individual file download with curl

liss-h:
Were getting rate-limited/blocked


nkaralis (approved):
Thank you

PR #477, opened 2026-08-20
Perf Fix: Literal ordering

nkaralis:


src/rdf4cpp/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -56,6 +56,11 @@ struct DatatypeRegistry {
 
     using compare_fptr_t = std::partial_ordering (*)(std::any const &, std::any const &) noexcept;
 
+    /**
+     * Compares two inlined values of the same datatype directly from their payloads.
+     */
+    using compare_inlined_fptr_t = std::partial_ordering (*)(storage::identifier::LiteralID, storage::identifier::LiteralID) noexcept;
```
liss-h, line 62: For this PR it would be good to measure if this particular actually makes a difference. It definitely could, but I don't want to include code that is not proven to improve things

src/rdf4cpp/datatypes/registry/DatatypeRegistry.hpp
```diff
@@ -604,8 +610,10 @@ inline void DatatypeRegistry::add() noexcept {
     auto const compare_fptr = []() -> compare_fptr_t {
         if constexpr (datatypes::ComparableLiteralDatatype<LiteralDatatype_t>) {
             return [](std::any const &lhs, std::any const &rhs) noexcept -> std::partial_ordering {
-                auto const &lhs_val = std::any_cast<typename LiteralDatatype_t::cpp_type>(lhs);
-                auto const &rhs_val = std::any_cast<typename LiteralDatatype_t::cpp_type>(rhs);
+                // cast to a reference: std::any_cast<T> returns by value, so binding its result to
```
liss-h, line 613: Doesn't this also apply to a bunch of other callsites?

src/rdf4cpp/Literal.cpp
```diff
@@ -1117,7 +1119,48 @@ Literal Literal::numeric_unop_impl(OpSelect op_select, storage::DynNodeStoragePt
...
+            return entry.inlining_ops->compare_inlined_fptr(lhs.backend_handle().node_id().literal_id(),
+                                                            rhs.backend_handle().node_id().literal_id());
+        }
+    }
+
+    if (datatype == datatypes::xsd::String::datatype_id) {
+        // xsd:String has no compare of its own, so this is the generic <=> on cpp_type = std::string_view
+        auto const &lhs_backend = lhs.handle_.literal_backend();
+        auto const &rhs_backend = rhs.handle_.literal_backend();
+        return lhs_backend.get_lexical().lexical_form <=> rhs_backend.get_lexical().lexical_form;
```
liss-h, line 1140: I am now wondering if we need to account for `needs_escape`. I don't think so but we should make sure of it

liss-h (commented):
Thank you, I have some comments

src/rdf4cpp/datatypes/registry/DatatypeRegistry.hpp
nkaralis: separate PR?

liss-h: ok

src/rdf4cpp/Literal.cpp
nkaralis: added tests

liss-h (approved):

PR #479, opened 2026-08-24
Fix: inverted datetime type reporting

### LINKED ISSUE #478: is_timepoint and is_duration are inverted for datatypes without such ops
`Literal::is_timepoint` and `Literal::is_duration` compare the registry lookup with `== nullptr`:

- `src/rdf4cpp/Literal.cpp:821-835` (at `babef89`): `is_timepoint` returns the result of `get_timepoint_ops(...) == nullptr`, `is_duration` the same with duration ops.
- `is_numeric` directly above uses `!= nullptr`.

The registry returns non-null when the datatype has such ops, so the two predicates answer true exactly when the datatype has no timepoint or duration ops.

Measured on the current head: `make_typed("x", IRI("http://example.org/mytype")).is_timepoint()` returns true for a dynamic datatype that has no timepoint ops.

The comparison dates to 0eb79e4 (2025-02-17). Impact is limited today: the inversion is only observable through dynamic datatypes, and nothing in the library calls the two predicates. Fix is flipping both comparisons plus a regression test with one fixed and one dynamic datatype.

liss-h:
Close #478

### DIFF
--- a/tests/nodes/tests_Literal.cpp
+++ b/tests/nodes/tests_Literal.cpp
@@ -1649,8 +1649,8 @@ TEST_CASE("is_numeric/is_timepoint/is_duration regression") {
 
     auto const t = Literal::make_typed_from_value<datatypes::xsd::DateTime>({});
     CHECK(t.is_timepoint());
-    CHECK_FALSE(l.is_duration());
-    CHECK_FALSE(l.is_numeric());
+    CHECK_FALSE(t.is_duration());
+    CHECK_FALSE(t.is_numeric());
 
     auto const d = Literal::make_typed_from_value<datatypes::xsd::Duration>({});
     CHECK_FALSE(d.is_timepoint());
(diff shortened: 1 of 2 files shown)


tests/nodes/tests_Literal.cpp
```diff
@@ -1640,3 +1640,10 @@ TEST_CASE("trigonometry/exponential funcs") {
         CHECK(make(1.0).math_atan2(str).null());
     }
 }
+
+TEST_CASE("is_numeric/is_timepoint/is_duration regression") {
```
nkaralis, line 1644: Should we also add `true` cases?

nkaralis (commented):
Thank you for the PR. Only one comment

nkaralis (approved):
Thank you for adding the tests. Let's wait for CI and then it can be merged

PR #486, opened 2026-08-31
Fix self-compare ordering

### LINKED ISSUE #484: Ordering comparison of a term with itself returns a boolean for non-literals instead of a type error
`Node::compare_impl` (src/rdf4cpp/Node.cpp) short-circuits identical handles before it
checks the term kind. It returns `equivalent` for any two identical handles. The
literal-only gate comes after it.

SPARQL 1.1 section 17.3 defines `<`, `<=`, `>`, `>=` only for literals of comparable
datatypes. There is no row for IRIs or blank nodes, so comparing one with itself is a
type error. Today:

- `?x <= ?x` and `?x >= ?x` evaluate to true for a blank node or IRI. FILTER keeps
  rows the spec drops.
- `?x < ?x` and `?x > ?x` evaluate to false instead of an error. A bare FILTER drops
  the row either way. The difference shows when the result is negated or composed
  (`!`, `||`, `IF`, `COALESCE`, `BIND`).
- Identical literals whose datatype has no defined order hit the same branch. The
  strict spec reading is a type error there too.

Only operand pairs that evaluate to the identical term are affected. Different
non-literal terms already produce a type error through the kind check. The `as_lt`,
`as_le`, `as_gt`, `as_ge` wrappers inherit the same values. `Node::order` and the
`order_*` comparators are correct as they are: ORDER BY needs a total order over all
terms, so the short-circuit must stay there. `eq` and `ne` are correct: RDFterm-equal
returns TRUE for the same term.

Fix sketch: in `compare_impl`, move the identical-handle short-circuit behind the
literal check, or restrict it to literal handles. Do not touch the short-circuit in
`order`. The same code is in rdf4cpp and rdf4cpp-private, fix both.

bigerl:
see https://github.com/tentris/rdf4cpp/issues/484

### DIFF
--- a/src/rdf4cpp/Node.cpp
+++ b/src/rdf4cpp/Node.cpp
@@ -161,16 +161,17 @@ std::partial_ordering Node::compare_impl(Node const &other) const {
         return std::partial_ordering::unordered;
     }
 
-    if (handle_ == other.handle_) {
-        return std::partial_ordering::equivalent;
-    }
-
     if (handle_.type() != other.handle_.type() || handle_.type() != storage::identifier::RDFNodeType::Literal) {
-        // mismatched node types are not comparable
-        // and nodes other than literals are not comparable with <,<=,>,>=
+        // "Operators invoked without appropriate operands result in a type error."
+        // - https://www.w3.org/TR/sparql11-query/#OperatorMapping
+        // The operator table lists <, <=, > and >= only for literal operands, so an IRI,
+        // a blank node or a variable is a type error. This also holds when both operands
+        // are the same node. Mismatched node types have no row either.
         return std::partial_ordering::unordered;
     }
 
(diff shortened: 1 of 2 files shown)


src/rdf4cpp/Node.cpp
```diff
@@ -161,16 +161,17 @@ std::partial_ordering Node::compare_impl(Node const &other) const {
         return std::partial_ordering::unordered;
     }
 
-    if (handle_ == other.handle_) {
```
liss-h, line 164: ~~we can keep the short circuit when `type() != Literal`
if (type() != Literal && handle_ == other.handle_) { return eq; }~~

liss-h (approved):
Nevermind, I misread something and didn't have in mind that compare_impl is only used for non eq operators

PR #487, opened 2026-09-01
Feature: Deferred numeric ops

nkaralis:
- `DeferredValue`: a value plus its datatype IRI, not (yet) in a node storage. A null datatype IRI is the null-value and propagates like the null-literal.                                                                                                                                                            
- `numeric_add_deferred / sub / mul / div`                                                                                                                                                                                     - `make_deferred_from_value<T>`, `make_deferred_from_litera`l, `materialize_deferred`                                                                                                                                           
- `Literal::make_typed_from_value(std::any, IRI const &, ns)`: runtime counterpart of the existing template overload

### DIFF
--- a/src/rdf4cpp/IRI.hpp
+++ b/src/rdf4cpp/IRI.hpp
@@ -11,14 +11,6 @@ namespace rdf4cpp {
  * IRI Resource node.
  */
 struct IRI : Node {
-private:
-    /**
-     * Constructs the corresponding IRI from a given datatype id and places it into node_storage if
-     * it does not exist already.
-     */
-    IRI(datatypes::registry::DatatypeIDView id, storage::DynNodeStoragePtr node_storage);
-
-public:
     /**
      * Constructs the corresponding datatype id for this iri. Return value can be safely used to
      * index the registry and yields the correct result.
@@ -44,6 +36,14 @@ public:
     explicit IRI(std::string_view iri, storage::DynNodeStoragePtr node_storage = storage::default_node_storage);
 
     /**
+     * Constructs the corresponding IRI from a given datatype id and places it into node_storage if
+     * it does not exist already.
+     * @param id datatype id
+     * @param node_storage optional custom node_storage used to store the IRI
+     */
+    explicit IRI(datatypes::registry::DatatypeIDView id, storage::DynNodeStoragePtr node_storage = storage::default_node_storage);
+
+    /**
      * Constructs the null-iri
      */
     [[nodiscard]] static IRI make_null() noexcept;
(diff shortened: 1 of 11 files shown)


src/rdf4cpp/IRI.hpp
```diff
@@ -63,6 +63,15 @@ struct IRI : Node {
      */
     [[nodiscard]] static IRI make_unchecked(std::string_view iri, storage::DynNodeStoragePtr node_storage = storage::default_node_storage);
 
+    /**
+     * Constructs the corresponding IRI from a given datatype id and places it into node_storage if
+     * it does not exist already.
+     * @param id datatype id
+     * @param node_storage optional custom node_storage used to store the IRI
+     */
+    [[nodiscard]] static IRI from_datatype_id(datatypes::registry::DatatypeIDView id,
```
liss-h: Why this function instead of making the existing ctor public?

src/rdf4cpp/Literal.hpp
```diff
@@ -1679,6 +1692,69 @@ struct Literal : Node {
...
+
+/**
+ * @brief Constructs a DeferredValue from a compatible type.
+ * The datatype is specified at compile time, which guarantees that the value and the datatype match.
+ * @tparam T the datatype
+ * @param value instance for which the DeferredValue is created
+ * @param node_storage node storage the datatype IRI is placed in
+ */
+template<datatypes::LiteralDatatype T>
+[[nodiscard]] auto make_deferred_from_value(typename T::cpp_type const &value,
```
liss-h, line 1720: DeferredValue return. Not auto

src/rdf4cpp/Literal.cpp
```diff
@@ -990,87 +1016,145 @@ Literal Literal::cast(IRI const &target, storage::DynNodeStoragePtr node_storage
...
+Literal materialize_deferred(DeferredValue value, storage::DynNodeStoragePtr node_storage) {
+    return Literal::make_typed_from_value(std::move(value.first), value.second, node_storage);
+}
+
+DeferredValue numeric_add_deferred(DeferredValue const &lhs, DeferredValue const &rhs) {
+    return numeric_binop_deferred_impl(
+            [](auto const &num_ops) noexcept {
+                return num_ops.add_fptr;
+            },
+            lhs, rhs, lhs.second.backend_handle().storage());
```
liss-h: This silently assumes that we want to use lhs's node storage. Add node storage param to all following functions

liss-h (changes requested):
Thank you for the PR. I only have minor comments. The logic looks good

src/rdf4cpp/Literal.hpp
nkaralis: does not work, because IRI is an incomplete type (used in the pair)

liss-h: unfortunate

liss-h (approved):
Thanks

PR #489, opened 2026-09-01
Warning: named return value optimization

### LINKED ISSUE #287: Enable `-Wnrvo` on GCC in CI
Warning against code that does not allow for named return value optimization.
Not sure which GCC version supports it.

mcb5637:
close #287

### DIFF
--- a/private/rdf4cpp/parser/JsonLdContextParser.cpp
+++ b/private/rdf4cpp/parser/JsonLdContextParser.cpp
@@ -210,10 +210,12 @@ namespace rdf4cpp::parser::json_ld {
 
             handle_ctx(o);  // 4
         } else if (local_context.is_scalar() && local_context.is_null()) { // 5.1
-            return handle_null();
+            result = handle_null();
+            return result;
         } else if (local_context.type() == simdjson::ondemand::json_type::string) {  // 5.2
             // a string names a remote context, the same case as inside the array below
-            return nonstd::unexpected{make_error(ParsingError::Type::BadSyntax, "remote context not supported")};
+            result = nonstd::unexpected{make_error(ParsingError::Type::BadSyntax, "remote context not supported")};
+            return result;
         } else {
             if (!p.propagate && result->previous_context == nullptr) {
                 result->previous_context = &p.active_context;
@@ -228,11 +230,10 @@ namespace rdf4cpp::parser::json_ld {
                     case simdjson::ondemand::json_type::null:  // 5.1
                     {
                         v.is_null();
-                        auto r = handle_null();
-                        if (!r.has_value()) {
-                            return r;
+                        result = handle_null();
+                        if (!result.has_value()) {
+                            return result;
                         }
-                        result = r;
                         break;
                     }
                     case simdjson::ondemand::json_type::object:  // 5.4
@@ -262,7 +263,8 @@ namespace rdf4cpp::parser::json_ld {
                         .override_protected = true,
                     });
                     if (!lc.has_value()) {
-                        return nonstd::unexpected(make_error(ParsingError::Type::BadSyntax, std::format("invalid scoped context ({})", lc.error().message)));
+                        result = nonstd::unexpected(make_error(ParsingError::Type::BadSyntax, std::format("invalid scoped context ({})", lc.error().message)));
+                        return result;
                     }
                 }
             }
@@ -271,25 +273,30 @@ namespace rdf4cpp::parser::json_ld {
         return result;
     }
     std::optional<ContextParser::error_type> ContextParser::parse_context_term(params::ParseContextTermParams p) {
+        std::optional<error_type> res;
         // https://www.w3.org/TR/json-ld11-api/#create-term-definition
         // 1
         if (p.term.parse_state == ParseState::Done) {
-            return std::nullopt;
+            res = std::nullopt;
+            return res;
         }
         if (p.term.parse_state == ParseState::InProgress) {
-            return make_error(ParsingError::Type::BadSyntax, "cyclic IRI mapping");
+            res = make_error(ParsingError::Type::BadSyntax, "cyclic IRI mapping");
+            return res;
         }
 
         // 2
         if (p.term.key.empty()) {
-            return make_error(ParsingError::Type::BadSyntax, "invalid term definition (empty term)");
+            res = make_error(ParsingError::Type::BadSyntax, "invalid term definition (empty term)");
+            return res;
         }
         p.term.parse_state = ParseState::InProgress;
 
         // 3
         auto [value_ec, value] = try_get_field<simdjson::ondemand::value>(p.local_context, p.term.key);
         if (value_ec != simdjson::SUCCESS) {
-            return make_error(ParsingError::Type::BadSyntax, "unknown key?");  // should not happen
+            res = make_error(ParsingError::Type::BadSyntax, "unknown key?");  // should not happen
+            return res;
         }
 
         // 6 (out of order, because type gets handled differently)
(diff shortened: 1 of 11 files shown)


private/rdf4cpp/parser/JsonLdContextParser.cpp
```diff
@@ -589,10 +620,12 @@ namespace rdf4cpp::parser::json_ld {
                             };
                             auto m = iri_expansion(p.active_context, p.term.key, false, true, nullptr, &p_ctx);
                             if (!m.has_value()) {
-                                return m.error();
+                                res = m.error();
+                                return res;
```
DanMadeThis, line 624: Isn't it already an nrvo?

private/rdf4cpp/parser/XMLParserUtility.cpp
```diff
@@ -108,16 +108,18 @@ namespace rdf4cpp::parser {
     }
 
     IRI XMLOutputQueue::make_id(std::string_view const local_name, std::string_view const base, XMLStateInfo const &i) {
+        auto iri = IRI::make_null();;
         if (!is_ncname(local_name)) {
             add_error(ParsingError::Type::BadIri, std::format("{}: is not a valid NCName (required for rdf:ID)", local_name), i);
-            return IRI::make_null();
+            return iri;
```
DanMadeThis, line 114: Do we propagate the error that way? Just a general question

src/rdf4cpp/bnode_mngt/reference_backends/scope/ReferenceNodeScope.hpp
```diff
@@ -62,11 +62,12 @@ struct ReferenceNodeScope {
...
 
     [[nodiscard]] Node get_or_generate_node(std::string_view label, storage::DynNodeStoragePtr node_storage = storage::default_node_storage) {
-        if (auto const node = try_get_node(label); !node.null()) {
+        auto node = try_get_node(label);
+        if (!node.null()) {
             return node;
         }
 
-        auto const node = generate_node(node_storage);
+        node = generate_node(node_storage);
```
DanMadeThis, line 70: Is this part of nrvo?

src/rdf4cpp/BigDecimal.hpp
```diff
@@ -614,9 +614,9 @@ struct BigDecimal {
      * @return
      */
     [[nodiscard]] constexpr BigDecimal div(const BigDecimal &other, Exponent_t max_scale_increase, RoundingMode mode = RoundingMode::Floor) const noexcept {
+        BigDecimal res{0, 0};
         if (other.unscaled_value == 0)
-            return BigDecimal{0, 0};  // undefined behavior (cpp_int throws)
-        BigDecimal res{0};
```
DanMadeThis, line 619: Should we add this back afterwards, so if the condition of if is false before it would reset to just one zero now it does not, is it correct?

src/rdf4cpp/datatypes/registry/DatatypeRegistry.cpp
```diff
@@ -263,10 +263,12 @@ std::optional<DatatypeRegistry::DatatypeConverter> DatatypeRegistry::get_common_
     };
 
     // call find_conv_impl with entries in correct order (lesser s rank, greater s rank)
+    std::optional<DatatypeConverter> res;
```
DanMadeThis, line 266: Should it be `= std::nullopt`?

src/rdf4cpp/Literal.cpp
```diff
@@ -1759,49 +1759,53 @@ std::optional<Literal> Literal::chrono_div_impl(Literal const &other, storage::D
...
     }
 
     if (other_entry->numeric_ops.has_value()) {
         // this is duration & other is scalar
-        return run_binop_cast_rhs(other, *other_entry, this_entry->duration_ops->duration_scalar_type, node_storage,
+        res = run_binop_cast_rhs(other, *other_entry, this_entry->duration_ops->duration_scalar_type, node_storage,
             [this_entry](std::any const &lhs, std::any const &rhs) noexcept {
                 return this_entry->duration_ops->duration_scalar_div(lhs, rhs);
             });
+        return res;
```
DanMadeThis, line 1805: No `res.has_value()` check is it correct?

src/rdf4cpp/Namespace.cpp
```diff
@@ -18,12 +18,14 @@ IRI Namespace::operator+(std::string_view suffix) const {
...
-        cache_.emplace(suffix, iri.backend_handle().id());
-        return iri;
+        return [&]() {
+            std::string namespace_iri{namespace_iri_};
+            namespace_iri.append(suffix);
+
+            IRI iri{namespace_iri, node_storage_};
+            cache_.emplace(suffix, iri.backend_handle().id());
+            return iri;
+        }();
```
DanMadeThis: Why the lambda? I remember you said something about this in the weekly

DanMadeThis (changes requested):
@mcb5637 Thank you for the pr. I have some suggestion and questions.

private/rdf4cpp/parser/JsonLdContextParser.cpp
mcb5637: it is a rvo (return value optimization)
and having any rvo prevents nrvo for the entire function (because the compiler cant predict which of the values will be returned)

private/rdf4cpp/parser/XMLParserUtility.cpp
mcb5637: specific to the xml parser:
- add_error adds an error message to the output queue
- any null marks the containing triple as invalid (caused an error)
- parsing may continue anyway for the next triples

src/rdf4cpp/bnode_mngt/reference_backends/scope/ReferenceNodeScope.hpp
mcb5637: yes, because nrvo only works, if you have 1 variable that gets returned from all return statements.

src/rdf4cpp/BigDecimal.hpp
mcb5637: `BigDecimal{0, 0}` and `BigDecimal{0}` lead to the same value (unscaledValue == exponent == 0), just a different ctor.

src/rdf4cpp/datatypes/registry/DatatypeRegistry.cpp
mcb5637: it gets initialized in both branches of the if statement below, so any value would be never read before that.
(also, leaving it uninitialized before any possible reading gets us a compiler error)

src/rdf4cpp/Literal.cpp
mcb5637: yes, the function returns an optional anyway

src/rdf4cpp/Namespace.cpp
mcb5637: just a different way of getting around the 1 variable problem. i simplified it.

private/rdf4cpp/parser/XMLParserUtility.cpp
DanMadeThis: Thanks for the explantaion

src/rdf4cpp/datatypes/registry/DatatypeRegistry.cpp
DanMadeThis: Makes sense

src/rdf4cpp/Literal.cpp
DanMadeThis: Okay, got it

src/rdf4cpp/Namespace.cpp
DanMadeThis: I see you made a change here

DanMadeThis (approved):
@mcb5637 Thanks. LGTM.

PR #490, opened 2026-09-03
Feature: improved inlining

### LINKED ISSUE #257: Decide what to do with the 13 free-tagging-bits inside `NodeBackendID`
Since #250 `NodeBackendID` has 13 free-tagging bits (previously it was 3),
thats a bit excessive but we also don't have a good idea where we would need them atm.

We need to decide what to do with them.

liss-h:
Improve the inlining of all datatypes, but specifically doubles.

Changes:
- reduce the number of tagging bits to 5 (close #257)
- change the way double inlining works. We are now betting on the fact that most doubles are probably written by humans in base-10. (e.g. writing 1.2E3 in turtle is written as base-10). Assuming this fact we can convert the double to decimal which (if the assumption was true) gives a much smaller representation than the regular binary format would have given and thus allows us to inline much more values.

Stripping bits from mantissa and exponent of the base-2 double does not nearly get the inlining coverage (on this test data) that the decimal approach can achieve (~4% vs 100%)

### Future Work
- Summing naively still does not give good inlining results ~4%. Next up will be a PR for a compensated sum that solves this issue.


src/rdf4cpp/datatypes/registry/util/DateTimeUtils.hpp
```diff
@@ -289,7 +289,7 @@ struct __attribute__((__packed__)) InliningHelperPacked {
     static constexpr std::size_t tv_width = width - 11;
 
     uint16_t tz_offset : 11;
-    uint32_t time_value : tv_width;
+    uint64_t time_value : tv_width;
```
nkaralis, line 292: is this related to the increased number of tagging bits?

src/rdf4cpp/storage/identifier/LiteralID.hpp
```diff
@@ -14,7 +14,7 @@ namespace rdf4cpp::storage::identifier {
  * <p>LiteralIDs are available in the range [0,(2^42-1)].</P>
  */
 struct __attribute__((__packed__)) LiteralID {
-    static constexpr size_t width = 42;
+    static constexpr size_t width = 50;
```
nkaralis, line 17: the comment needs to be updated. the readme describing the layout needs to be updated, right?

src/rdf4cpp/storage/identifier/NodeID.hpp
```diff
@@ -17,7 +17,7 @@ namespace rdf4cpp::storage::identifier {
  */
 struct __attribute__((__packed__)) NodeID {
     using underlying_type = uint64_t;
-    static constexpr size_t width = 48;
+    static constexpr size_t width = 56;
```
nkaralis, line 21: update comment (states 48)

tests/serializer/tests_Serialization.cpp
```diff
@@ -189,8 +189,8 @@ TEST_CASE("basic ntriple") {
     // reordered because graph is unordered_map
     CHECK_EQ(res,
             "<http://ex/sub> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> \"5\"^^<http://www.w3.org/2001/XMLSchema#int> .\n"
-            "<http://ex/sub> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://ex/obj> .\n"
             "<http://ex/sub> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> \"7\"^^<http://www.w3.org/2001/XMLSchema#int> .\n"
+            "<http://ex/sub> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://ex/obj> .\n"
```
nkaralis, line 193: the changes here are only reordering-related, right?

tests/datatype/tests_Double.cpp
```diff
@@ -103,3 +104,58 @@ TEST_CASE("double inlining") {
...
+}
+
+TEST_CASE("double inlining large") {
+    size_t total = 0;
+    size_t num_inlined = 0;
+    size_t num_sums_inlined = 0;
+
+    Literal sum = 0.0_xsd_double;
+
+    for (auto const &quad : parser::RDFFileParser{"doubles.ttl"}) {
```
nkaralis, line 140: instead of `doubles.ttl`, would it make sense to use randonly generated doubles?

tests/nodes/tests_NodeStorage_ids.cpp
```diff
@@ -39,7 +39,8 @@ TEST_SUITE("node storage identifier output") {
     }
 
     TEST_CASE("NodeID") {
-        check_string_repr(NodeID{LiteralID{1}, LiteralType::from_parts(LiteralTypeTag::Default, 1)}, "{ .underlying = 4398046511105 }");
+        check_string_repr(NodeID{LiteralID{1}, LiteralType::from_parts(LiteralTypeTag::Default, 1)},
```
nkaralis, line 42: what does this check?

src/rdf4cpp/datatypes/xsd/Double.cpp
```diff
@@ -46,24 +48,81 @@ nonstd::expected<capabilities::Numeric<xsd_double>::ceil_result_cpp_type, Dynami
...
+static_assert(sizeof(DecimalDoubleLayout) == sizeof(uint64_t));
+
+template<>
+capabilities::Inlineable<xsd_double>::cpp_type capabilities::Inlineable<xsd_double>::from_inlined(storage::identifier::LiteralID inlined) noexcept {
+    auto const decimal = util::unpack<DecimalDoubleLayout>(inlined);
 
-    return storage::identifier::LiteralID{shortened};
+    auto const significand = static_cast<cpp_type>(decimal.significand);
+
+    // dragonbox promised that the nearest double to significand * 10^exponent is the original value, so
```
nkaralis, line 87: i find the comment difficult to follow

nkaralis (commented):
Thank you for the PR. I have some questions and comments.

src/rdf4cpp/datatypes/registry/util/DateTimeUtils.hpp
liss-h: yes, tv_width > 32 now

tests/serializer/tests_Serialization.cpp
liss-h: yes

tests/nodes/tests_NodeStorage_ids.cpp
liss-h: what the formatter prints

tests/datatype/tests_Double.cpp
liss-h: the random doubles are probably not good decimals so this is a more accurate way to represent it

src/rdf4cpp/datatypes/xsd/Double.cpp
liss-h: basically, the base 10 number dragonbox gave us is the nearest possible representation of the double we gave it.

To retrieve that exact double again we are only allowed a single inaccurate floating point op where both operands are exact.

S = significant is exact
E = 10^N is exact for N in [0..22]
A = 10^M for M<0 is never exact

So we can do either `S / E` or `S * E` but we are not allowed to do `S * A` because it would round twice and therefore not land on the original value.

I hope that makes it clearer

nkaralis (approved):

PR #496, opened 2026-09-10
performance: `any_cast` with `const &`

### LINKED ISSUE #488: Always use any_cast with const &
`std::any_cast<T>` copies `std::any_cast<T const &>` does not

DanMadeThis:
Closes #488


mcb5637 (changes requested):
thanks for the PR.
changes look good, there are just 3 any_casts missing.
- LiteralBackendView.hpp 30 & 35
- SpecializedLiteralBackend.hpp 20

(assuming we dont want to change tests here, as they are not performance critical)

mcb5637 (approved):

PR #499, opened 2026-09-10
Perf improvement for compensated sum

liss-h:
Use multiplication by multiplicity if datatype is exact instead of loop.

Introduces a new registry function:
`Literal::from_multiplicity(uint64_t, IRI datatype)` for that purpose.

### DIFF
--- a/src/rdf4cpp/CompensatedSum.cpp
+++ b/src/rdf4cpp/CompensatedSum.cpp
@@ -30,6 +30,20 @@ void CompensatedSum::add(Literal const &lit, uint64_t multiplicity) {
 }
 
 void CompensatedSum::add(DeferredLiteral const &value, uint64_t multiplicity) {
+    if (multiplicity == 1) {
+        add_once(value);
+        return;
+    }
+
+    if (is_exact(value.datatype)) {
+        // use multiplication by multiplicity instead of loop
+        // no precision loss possible because the dt is exact
+        auto const mult = make_deferred_from_multiplicity(multiplicity, value.datatype);
+        add_once(numeric_mul_deferred(value, mult, node_storage_));
+        return;
+    }
+
+    // datatype is not exact, multiplying with multiplicity would result in precision loss
     // assumption: multiplicity is usually small
     while (multiplicity > 0) {
         add_once(value);
--- a/src/rdf4cpp/DeferredLiteral.cpp
+++ b/src/rdf4cpp/DeferredLiteral.cpp
@@ -18,6 +18,30 @@ DeferredLiteral make_deferred_from_literal(Literal const &lit) {
     return DeferredLiteral{lit.value(), lit.datatype()};
 }
 
+static DeferredLiteral make_deferred_from_multiplicity_impl(uint64_t multiplicity, datatypes::registry::DatatypeIDView const &dt_view, storage::DynNodeStoragePtr node_storage) {
+    auto const *entry = datatypes::registry::DatatypeRegistry::get_entry(dt_view);
+    if (entry == nullptr || !entry->numeric_ops.has_value()) {
+        return DeferredLiteral{};
+    }
+
+    if (entry->numeric_ops->is_stub()) {
+        auto const impl_dt = datatypes::registry::DatatypeRegistry::get_numeric_op_impl_conversion(*entry).target_type_id;
+        return make_deferred_from_multiplicity_impl(multiplicity, impl_dt, node_storage);
+    }
+
+    auto const &impl_ops = entry->numeric_ops->get_impl();
+    auto res = impl_ops.from_multiplicity_fptr(multiplicity);
+    if (!res.has_value()) {
+        return DeferredLiteral{};
+    }
+
+    return deferred_detail::make_deferred_from_value(std::move(*res), dt_view, node_storage);
+}
+
+DeferredLiteral make_deferred_from_multiplicity(uint64_t multiplicity, IRI const &datatype) {
+    return make_deferred_from_multiplicity_impl(multiplicity, datatype, datatype.backend_handle().storage());
+}
+
 Literal materialize_deferred(DeferredLiteral value, storage::DynNodeStoragePtr node_storage) {
     return Literal::make_typed_from_value(std::move(value.value), value.datatype, node_storage);
 }
--- a/src/rdf4cpp/Literal.cpp
+++ b/src/rdf4cpp/Literal.cpp
@@ -252,6 +252,10 @@ Literal Literal::make_boolean(TriBool const b, storage::DynNodeStoragePtr node_s
     return Literal::make_typed_from_value<datatypes::xsd::Boolean>(b == TriBool::True, node_storage);
 }
 
+Literal Literal::make_from_multiplicity(uint64_t multiplicity, IRI const &datatype, storage::DynNodeStoragePtr node_storage) {
+    return materialize_deferred(make_deferred_from_multiplicity(multiplicity, datatype), node_storage);
+}
+
 Literal Literal::make_string_uuid(storage::DynNodeStoragePtr node_storage) {
     boost::uuids::random_generator_mt19937 gen{};
     boost::uuids::uuid u = gen();
--- a/src/rdf4cpp/Literal.hpp
+++ b/src/rdf4cpp/Literal.hpp
@@ -476,7 +476,35 @@ public:
      *
      * @return the literal form of the given boolean
      */
-    static Literal make_boolean(TriBool b, storage::DynNodeStoragePtr node_storage = storage::default_node_storage);
+    [[nodiscard]] static Literal make_boolean(TriBool b, storage::DynNodeStoragePtr node_storage = storage::default_node_storage);
+
+    /**
+     * Make a literal by converting the given multiplicity to the given datatype
+     *
+     * @param multiplicity multiplicity
+     * @param datatype datatype iri
+     * @param node_storage node storage where to place literal
+     * @return literal corresponding to the multiplicity (or null-literal if the multiplicity was not representable)
+     */
+    [[nodiscard]] static Literal make_from_multiplicity(uint64_t multiplicity, IRI const &datatype, storage::DynNodeStoragePtr node_storage = storage::default_node_storage);
+
+    /**
+     * See other overload
+     */
+    template<datatypes::NumericLiteralDatatype T>
+    [[nodiscard]] static Literal make_from_multiplicity(uint64_t multiplicity, storage::DynNodeStoragePtr node_storage = storage::default_node_storage) {
+        if constexpr (datatypes::NumericStub<T>) {
+            return Literal::make_from_multiplicity<typename T::numeric_impl_type>(multiplicity, node_storage);
+        } else {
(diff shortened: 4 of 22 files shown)


src/rdf4cpp/DeferredLiteral.cpp
```diff
@@ -18,6 +18,30 @@ DeferredLiteral make_deferred_from_literal(Literal const &lit) {
...
+    }
+
+    auto const &impl_ops = entry->numeric_ops->get_impl();
+    auto res = impl_ops.from_multiplicity_fptr(multiplicity);
+    if (!res.has_value()) {
+        return DeferredLiteral{};
+    }
+
+    return deferred_detail::make_deferred_from_value(std::move(*res), dt_view, node_storage);
+}
```
DanMadeThis, line 39: Can be made NRVO, do we want that here?

src/rdf4cpp/datatypes/registry/util/ConstexprString.hpp
```diff
@@ -38,6 +38,10 @@ struct ConstexprString {
         // implicitly null terminated by default init of array
     }
 
+    [[nodiscard]] constexpr char const *c_str() const noexcept {
+        return value.data(); // always null-terminated
+    }
+
     [[nodiscard]] constexpr char const *data() const noexcept {
         return value.data();
     }
```
DanMadeThis, line 47: What is the difference between these to functions is it part of the standard or API spec?

src/rdf4cpp/Literal.hpp
```diff
@@ -476,7 +476,35 @@ struct Literal : Node {
...
+        } else {
+            // numeric impl
+            auto res = T::from_multiplicity(multiplicity);
+            if (!res.has_value()) {
+                return Literal{};
+            }
+
+            return Literal::make_typed_from_value<T>(*res, node_storage);
+        }
+    }
```
DanMadeThis, line 507: Can this also be made NRVO, and do we want that here also?

DanMadeThis (changes requested):
@liss-h Thanks for the pr. I have a couple of small suggestions and a general question.

src/rdf4cpp/DeferredLiteral.cpp
liss-h: this is RVO, which is guaranteed

src/rdf4cpp/datatypes/registry/util/ConstexprString.hpp
liss-h: same function, just that c_str() has the contract to always return null terminated. data() does not have to

src/rdf4cpp/Literal.hpp
liss-h: this is RVO, which is guaranteed

src/rdf4cpp/DeferredLiteral.cpp
liss-h: The difference between RVO (return-value-optimization) and NRVO (named-return-value-optimization) is whether the return value is given a name or not.

```c++
T func() {
    return T{}; // RVO opportunity (guaranteed to optimize by standard)
}
```

```c++
T func() {
    T value{};
    return value; // NRVO opportunity (not guaranteed to optimize by standard, but clang and gcc will optimize it)
}
```

DanMadeThis (approved):

src/rdf4cpp/Literal.hpp
```diff
@@ -476,7 +476,35 @@ struct Literal : Node {
...
+
+    /**
+     * Make a literal by converting the given multiplicity to the given datatype
+     *
+     * @param multiplicity multiplicity
+     * @param datatype datatype iri
+     * @param node_storage node storage where to place literal
+     * @return literal corresponding to the multiplicity (or null-literal if the multiplicity was not representable)
+     */
+    [[nodiscard]] static Literal make_from_multiplicity(uint64_t multiplicity, IRI const &datatype, storage::DynNodeStoragePtr node_storage = storage::default_node_storage);
```
nkaralis, line 489: the name is clear for its intended use-case. should it be more general though?

src/rdf4cpp/Literal.cpp
```diff
@@ -252,6 +252,10 @@ Literal Literal::make_boolean(TriBool const b, storage::DynNodeStoragePtr node_s
     return Literal::make_typed_from_value<datatypes::xsd::Boolean>(b == TriBool::True, node_storage);
 }
 
+Literal Literal::make_from_multiplicity(uint64_t multiplicity, IRI const &datatype, storage::DynNodeStoragePtr node_storage) {
+    return materialize_deferred(make_deferred_from_multiplicity(multiplicity, datatype), node_storage);
```
nkaralis, line 256: do we need to have a version that does not materialize (for perf reasons)?

src/rdf4cpp/datatypes/xsd/Float.cpp
```diff
@@ -46,6 +46,19 @@ nonstd::expected<capabilities::Numeric<xsd_float>::ceil_result_cpp_type, Dynamic
     return std::ceil(operand);
 }
 
+template<>
+nonstd::expected<capabilities::Default<xsd_float>::cpp_type, DynamicError> capabilities::Numeric<xsd_float>::from_multiplicity(uint64_t multiplicity) noexcept {
+    constexpr auto available_mantissa_bits = std::numeric_limits<cpp_type>::digits;
```
nkaralis, line 51: please add some comments here

src/rdf4cpp/datatypes/xsd/Double.cpp
```diff
@@ -48,6 +48,20 @@ nonstd::expected<capabilities::Numeric<xsd_double>::ceil_result_cpp_type, Dynami
     return std::ceil(operand);
 }
 
+template<>
+nonstd::expected<capabilities::Default<xsd_double>::cpp_type, DynamicError> capabilities::Numeric<xsd_double>::from_multiplicity(uint64_t multiplicity) noexcept {
+    constexpr auto available_mantissa_bits = std::numeric_limits<cpp_type>::digits;
```
nkaralis, line 53: same here. also, can this be extracted and used in both float and double?

src/rdf4cpp/datatypes/registry/util/ConstexprString.hpp
```diff
@@ -38,6 +38,10 @@ struct ConstexprString {
         // implicitly null terminated by default init of array
     }
 
+    [[nodiscard]] constexpr char const *c_str() const noexcept {
```
nkaralis, line 41: what's the difference with `data()` (right below this one)?

src/rdf4cpp/DeferredLiteral.cpp
```diff
@@ -18,6 +18,30 @@ DeferredLiteral make_deferred_from_literal(Literal const &lit) {
...
+
+DeferredLiteral make_deferred_from_multiplicity(uint64_t multiplicity, IRI const &datatype) {
+    return make_deferred_from_multiplicity_impl(multiplicity, datatype, datatype.backend_handle().storage());
```
nkaralis, line 43: do we need to check if the IRI is null?

src/rdf4cpp/CompensatedSum.cpp
```diff
@@ -30,6 +30,20 @@ void CompensatedSum::add(Literal const &lit, uint64_t multiplicity) {
...
 void CompensatedSum::add(DeferredLiteral const &value, uint64_t multiplicity) {
+    if (multiplicity == 1) {
+        add_once(value);
+        return;
+    }
+
+    if (is_exact(value.datatype)) {
+        // use multiplication by multiplicity instead of loop
+        // no precision loss possible because the dt is exact
+        auto const mult = make_deferred_from_multiplicity(multiplicity, value.datatype);
```
nkaralis: `multiplicity == 0` make the sum null. i thinks this happens only in this branch

tests/nodes/tests_Literal.cpp
```diff
@@ -1662,3 +1662,77 @@ TEST_CASE("is_numeric/is_timepoint/is_duration regression") {
...
+                    CHECK(huge_value_expected.null());
+                } else {
+                    CHECK_EQ(materialize_deferred(lit), huge_value_expected);
+                }
+            }
+        }
+    }
+}
+
+TEST_CASE("from_multiplicity") {
```
nkaralis, line 1720: suggestions for more tests:
- test with 0 multiplicity
- test with values near the boundaries for double and float
- test with null and unregistered IRIs

nkaralis (commented):
Thank you for the PR. I have some comments.

src/rdf4cpp/Literal.cpp
liss-h: we have one `make_deferred_from_multiplicity` it is used in the compensated sum

src/rdf4cpp/datatypes/registry/util/ConstexprString.hpp
liss-h: just the contract. `c_str()` always returns null-terminated, `data()` does not necessarily have to

src/rdf4cpp/DeferredLiteral.cpp
liss-h: I can add an assert

src/rdf4cpp/CompensatedSum.cpp
liss-h: it does not make the sum 0. It adds zero. I think this is the same for all branches

src/rdf4cpp/datatypes/xsd/Float.cpp
liss-h: idk what to comment here. The names make it pretty clear what is happening, no?

tests/nodes/tests_Literal.cpp
liss-h: I would not test for null IRIs. Nothing in rdf4cpp tests for that

src/rdf4cpp/CompensatedSum.cpp
nkaralis: right, sorry. however, the loop does not consider multiplicity 0. here, `add(1.5^^decimal, 0)` returns `0.0^^decimal` instead of `0^^integer`

`add(1/3^^rational, 0); add(1^^integer)` seems to be a problematic case. it returns `null`

i think that multiplicity 0 should always be skipped.

liss-h: ah ok, now I get it

src/rdf4cpp/Literal.hpp
liss-h: idk, do you have a suggestion?

nkaralis: something like `make_from_uint`, but it's not really helpful. i guess we can leave it as it is

src/rdf4cpp/datatypes/xsd/Float.cpp
nkaralis: okay

nkaralis (approved):

PR #503, opened 2026-09-15
Use botan instead of openssl

liss-h:
Switch to botan for hashes and crypto rng over OpenSSL to avoid regularly shipping vulnerabilities.


liss-h:
~~Tests failing because ubuntu arm cannot download cmake for some reason. Pretty sure this is unrelated to this PR.~~

src/rdf4cpp/util/Anonymizer.cpp
```diff
@@ -21,18 +20,8 @@ struct GlobalCryptoRng {
...
-        if (ret != 1) [[unlikely]] {
-            std::array<char, 120> err_buf;
-            unsigned long const code = ERR_get_error();
-            ERR_error_string_n(code, err_buf.data(), err_buf.size());
-
-            throw std::runtime_error{std::format("Unable to generate random bytes for anonymizer: {}", std::string_view{err_buf.data()})};
-        }
-
-        return std::bit_cast<result_type>(buf);
+        thread_local Botan::AutoSeeded_RNG rng;
```
mcb5637, line 23: why are we using `AutoSeeded_RNG` over `System_RNG`?

mcb5637 (changes requested):
Just one question:

src/rdf4cpp/util/Anonymizer.cpp
liss-h: tbh I haven't given it much thought. I will look up the difference

liss-h: As discussed: autoseed rng is good enough security wise, system rng would need to go into kernel

mcb5637 (approved):

